PR #62, opened 2016-10-05
Support weights in file and improve vis / weights / flags API

ludwigschwardt:
The main reason for this PR is to support the latest MeerKAT HDF5
files with proper weights. This also allows h5toms to work with these
files (currently broken).

In the process some long-standing warts have been removed in
the vis / weights / flags API. All three attributes are now properties,
where weights and flags used to be methods to allow selection.
The selection is now done via the usual select() call. Flags and
weights have also been properly extended to all file formats.

The `squeeze` parameter on v2 and v3 datasets has been renamed
to `keepdims` to match the corresponding concept in NumPy (and it
now has the opposite sense). Its default behaviour also changed,
so that singleton dimensions are now discarded where indexed away
by default. Use `keepdims=True` to obtain the old behaviour.

There is still a lot of duplication in the code (e.g. between v2 and v3)
but this needs more careful refactoring which is left for another
week.

This addresses JIRA ticket [MKAIV-345](https://skaafrica.atlassian.net/browse/MKAIV-345).

@bmerry gets it because he wrote the weights part on the ingest side...

### DIFF
--- a/katdal/h5datav2.py
+++ b/katdal/h5datav2.py
@@ -156,8 +156,8 @@ class H5DataV2(DataSet):
         True if synthesised timestamps should be used to partition data set even
         if real timestamps are irregular, thereby avoiding the slow loading of
         real timestamps at the cost of slightly inaccurate label borders
-    squeeze : {False, True}, optional
-        Don't force vis / weights / flags to be 3-dimensional
+    keepdims : {False, True}, optional
+        Force vis / weights / flags to be 3-dimensional, regardless of selection
     kwargs : dict, optional
         Extra keyword arguments, typically meant for other formats and ignored
 
@@ -168,7 +168,7 @@ class H5DataV2(DataSet):
 
     """
     def __init__(self, filename, ref_ant='', time_offset=0.0, mode='r',
-                 quicklook=False, squeeze=False, **kwargs):
+                 quicklook=False, keepdims=False, **kwargs):
         DataSet.__init__(self, filename, ref_ant, time_offset)
 
         # Load file
@@ -228,7 +228,7 @@ class H5DataV2(DataSet):
         self._time_keep = np.ones(num_dumps, dtype=np.bool)
         self.start_time = katpoint.Timestamp(data_timestamps[0] - 0.5 * self.dump_period)
         self.end_time = katpoint.Timestamp(data_timestamps[-1] + 0.5 * self.dump_period)
-        self._squeeze = squeeze
+        self._keepdims = keepdims
 
         # ------ Extract flags ------
 
@@ -566,7 +566,7 @@ class H5DataV2(DataSet):
                             for dim_keep in keep]
             return data[tuple(keep_singles)]
         force_3dim = LazyTransform('force_3dim', _force_3dim)
-        transforms = [extractor] if self._squeeze else [extractor, force_3dim]
+        transforms = [extractor, force_3dim] if self._keepdims else [extractor]
         return LazyIndexer(dataset, stage1, transforms)
 
     @property
--- a/katdal/h5datav2.py
+++ b/katdal/h5datav2.py
@@ -156,6 +156,8 @@ class H5DataV2(DataSet):
         True if synthesised timestamps should be used to partition data set even
         if real timestamps are irregular, thereby avoiding the slow loading of
         real timestamps at the cost of slightly inaccurate label borders
+    squeeze : {False, True}, optional
+        Don't force vis / weights / flags to be 3-dimensional
     kwargs : dict, optional
         Extra keyword arguments, typically meant for other formats and ignored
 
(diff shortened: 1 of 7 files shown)


mauch:
With this change will the old flags and weights API be supported? We will need to fix a number of things in katsdpscripts that will break if not.

ludwigschwardt:
We will fix those number of things :-)

mauch:
What I meant was that the scripts that are run by the workflow manager will break when this is merged so it would be easier to fix things in katsdpscripts first to avoid failed reductions in the workflow manager.

Also, you mention getting h5toms to work again now that weights are fixed. Currently h5toms does not actually insert weights from the h5 file into measurement sets. After they are read from the h5 file they are then never passed to populate_main_dict [here](https://github.com/ska-sa/katdal/blob/e6c22f750117b2ae085224c6fdc8f76729b016bc/scripts/h5toms.py#L404-L407). Inside populate_main_dict, 'WEIGHT' is just set to 1. Someone will need to create 'WEIGHT_SPECTRUM' and fill it with the weights from the h5 file. Fixing this will also put to bed the long standing (3 years old now) [KAT-1209](https://skaafrica.atlassian.net/browse/KAT-1209). Perhaps that is a separate pull request:)

ludwigschwardt:
We should just fix katsdpscripts after this PR is in and then roll out the
changes together to site. That should keep everything running.

The problem with h5toms is that it still attempts to read the weights, even
if it doesn't use it, and this crashes currently.

You are right - proper weights would be another PR.

mauch:
Sure. I keep katsdpscripts up to date on the workflow manager machine via a
cron job (katsdpscripts changes too frequently for me to want to do it
manually). I'll switch this off until changes to katsdpscripts etc. are
merged.

ludwigschwardt:
More people are starting to hit the h5toms bug (now 2, up from 1)... So the urgency for this has gone up.

katdal/h5datav2.py
```diff
@@ -443,6 +453,66 @@ def __str__(self):
...
+                logger.warning("%r is not a legitimate weight type for this file, "
+                               "supported ones are %s" % (name, known_weights))
+        if known_weights and not selection:
+            logger.warning('No valid weights were selected - setting all weights to 1.0 by default')
+        self._weights_select = selection
+
+    @property
+    def _flags_keep(self):
+        if not hasattr(self, '_flags_description'):
+            return []
```
bmerry: I'm guessing these various `hasattr` tests are because the base class touches `_flags_keep` and `_weights_keep` before the derived class constructor has completed. That has a bit of code smell about it, and at a minimum needs some comments. It might be cleaner (if a bit less efficient) to just compute the selection when accessing the `weights` property rather than when `_weights_keep` is set.

katdal/h5datav2.py
```diff
@@ -457,6 +527,48 @@ def timestamps(self):
...
+        """
+        # Create first-stage index from dataset selectors
+        time_keep = self._time_keep
+        # If there is a duplicate final dump, these lengths don't match -> ignore last dump in file
+        if len(time_keep) == len(dataset) - 1:
+            time_keep = np.zeros(len(dataset), dtype=np.bool)
+            time_keep[:len(self._time_keep)] = self._time_keep
+        stage1 = (time_keep, self._freq_keep, self._corrprod_keep)
+
+        def _force_3dim(data, keep):
```
bmerry: I can't see anything that this function is capturing from the outer scope, so it could be a class method rather than an inner function.

katdal/h5datav2.py
```diff
@@ -473,110 +585,55 @@ def vis(self):
...
+        matches the length of :meth:`corr_products`. To get the data array
+        itself from the indexer `x`, do `x[:]` or perform any other form of
+        indexing on it. Only then will data be loaded into memory.
 
         """
-        names = names.split(',') if isinstance(names, basestring) else WEIGHT_NAMES if names is None else names
+        # We currently only cater for a single weight type (i.e. either select it or fall back to 1.0)
+        def transform(weights, keep):
+            return weights.astype(np.float32) if self._weights_select else \
+                np.ones_like(weights, dtype=np.float32)
```
bmerry: For the v2 format, don't you need to index weights by the selected weight type?

katdal/h5datav3.py
```diff
@@ -309,13 +309,21 @@ def register_telstate_sensor(name, obj):
...
+        self._flags_keep = 'all'
 
         # ------ Extract weights ------
 
-        # check if weight group present, else use dummy weight data
+        # Check if weights and weights_channel datasets are present, else use dummy weight data
         self._weights = data_group['weights'] if 'weights' in data_group else \
-            dummy_dataset('dummy_weights', shape=self._vis.shape[:-1] + (1,), dtype=np.float32, value=1.0)
-        self._weights_description = np.array(zip(WEIGHT_NAMES, WEIGHT_DESCRIPTIONS))
+            dummy_dataset('dummy_weights', shape=self._vis.shape[:-1], dtype=np.float32, value=1.0)
```
bmerry: To make it behave exactly like the real dataset, it should be `np.uint8`. The difference might not be observable though.

PR #63, opened 2016-11-03
Support multiple receivers

ludwigschwardt:
This is a step towards answering the age-old question "What is the centre frequency?".

The subarray band sensor is the most reliable indicator of the selected band as it matches the actual receivers and digitisers in use. Use that instead of the flakey AP indexer positions. Once we know the band, use that to figure out the centre frequency, which is correct for wideband modes.

The current supported receivers are:

- 'L': the standard 'l' band
- 'Ku': the only 'x' band so far, frequency derived from mixer sensor
- 'UHF': the only 'u' band so far, a hacked / custom version with an L-band digitiser and flipped spectrum

When the real 'x' and 'u' finally arrive we will deal with the disambiguation (serial numbers should suffice).

The Ku mixer sensor was accidentally dropped from AR1 and hence now RTS but will be restored soon ([katconfig PR](https://github.com/ska-sa/katconfig/pull/312) waiting).

If the centre frequency is not found, fall back to 0 Hz instead. This is an indication that the user should override the value, without preventing the file from being opened. The band can also be overridden.

Printing the file object now indicates the band, e.g.:
```
Spectral Windows: 1
  ID Band Product  CentreFreq(MHz)  Bandwidth(MHz)  Channels  ChannelWidth(kHz)
   0 Ku   -         11696.000         856.000           4096       208.984
```

This addresses JIRA ticket [MRTS-369](https://skaafrica.atlassian.net/browse/MRTS-369).

### DIFF
--- a/katdal/dataset.py
+++ b/katdal/dataset.py
@@ -114,7 +114,8 @@ class SpectralWindow(object):
     channel width. The channels are assumed to be regularly spaced and to be the
     result of either lower-sideband downconversion (channel frequencies
     decreasing with channel index) or upper-sideband downconversion (frequencies
-    increasing with index).
+    increasing with index). For further information the receiver band and
+    correlator product names are also available.
 
     Parameters
     ----------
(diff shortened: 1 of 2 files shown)


spassmoor:
This looks ok 
Tested on file:
/var/kat/archive/data/RTS/telescope_products/2016/11/07/1478517958.h5	
the band and the frequency data is the correct way round and associated information is correct 

 
  ID Band Product  CentreFreq(MHz)  Bandwidth(MHz)  Channels  ChannelWidth(kHz)
   0 UHF  c856M4k     428.000         856.000           4096       208.984

PR #64, opened 2016-11-14
Order dumps by time in h5toms

sjperkins:
Previously when handling a scan, h5toms would iterate over antenna pairs, handling all time, frequency and polarisation data for the associated baseline. This meant that data would be written to the Measurement Set in baseline ordering, rather than the more natural time ordering.

This commit modifies h5toms.py to iterate over chunks of time within a scan, thereby ordering
the MS by time. The time chunk size is chosen as a multiple of the time dump average.

### DIFF
--- a/scripts/h5toms.py
+++ b/scripts/h5toms.py
@@ -303,9 +303,10 @@ for win in range(len(h5.spectral_windows)):
         Create individual correlator product index
         from antenna pair and polarisation
         """
-        return corrprod_to_index.get((
-            "%s%s" % (a1.name, pol[0].lower()),
-            "%s%s" % (a2.name, pol[1].lower())))
+        a1 = "%s%s" % (a1.name, pol[0].lower())
+        a2 = "%s%s" % (a2.name, pol[1].lower())
+
+        return corrprod_to_index.get((a1, a2))
 
     nbl = ant1_index.size
     npol = len(pols_to_use)
@@ -315,14 +316,13 @@ for win in range(len(h5.spectral_windows)):
         for a1, a2 in itertools.izip(ant1, ant2)
         for p in pols_to_use])
 
-    # Identify valid and invalid correlator products
+    # Identify missing correlator products
     # Reshape for broadcast on time and frequency dimensions
-    valid_cp = np.asarray([i is not None for i in cp_index])
-    invalid_cp = np.logical_not(valid_cp)
+    missing_cp = np.logical_not([i is not None for i in cp_index])
 
     # Zero any None indices, but use the above masks to reason
     # about there existence in later code
-    cp_index[invalid_cp] = 0
+    cp_index[missing_cp] = 0
 
     field_names, field_centers, field_times = [], [], []
     obs_modes = ['UNKNOWN']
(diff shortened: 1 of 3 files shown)


sjperkins:
@ludwigschwardt Can you recommend a method of checking that this change has not broken any existing functionality? In particular I want to check that time and channel averaging are behaving as expected. I was able to test channel averaging that ran through at least but the h5 file I found didn't have enough time slots. In short, I'm looking for a representative hdf5 file and a way to test it.

ludwigschwardt:
I would ask @mauch to check this.

mauch:
I'm happy to take a look.

scripts/h5toms.py
```diff
@@ -335,78 +368,117 @@
...
+            # cp_index could be used above when the LazyIndexer
+            # supports advanced integer indices
+            vis_data = scan_data[:,:,cp_index]
+            weight_data = scan_weight_data[:,:,cp_index]
+            flag_data = scan_flag_data[:,:,cp_index]
+
+            # Zero any invalid correlator products
+            vis_data[:,:,invalid_cp] = 0 + 0j
+            weight_data[:,:,invalid_cp] = 0
+            flag_data[:,:,invalid_cp] = False
```
sjperkins: @ludwigschwardt  If there's no correlator product available, should we not flag the data? However, the above reflects original functionality.

scripts/h5toms.py
```diff
@@ -335,78 +368,117 @@
...
+
+            # Overwrite the input visibilities with averaged visibilities,flags,weights,timestamps,channel freqs
+            if average_data:
+                vis_data, weight_data, flag_data, out_utc, out_freqs = \
+                    averager.average_visibilities(vis_data, weight_data, flag_data, out_utc, out_freqs,
+                                                  timeav=dump_av, chanav=chan_av, flagav=options.flagav)
+
+                # Time and channel dimensions change now
+                tdiff /= dump_av
+                nchan /= chan_av
```
sjperkins: Check that tdiff and nchan inherit parent scope.

sjperkins:
@mauch Note I've just pushed https://github.com/ska-sa/katdal/pull/64/commits/124145587a6ac46ea384eed591f0676c6cec6b22. Without this fix, the channel dimension would've been exponentially reduced on each iteration of the time loop.

scripts/h5toms.py
```diff
@@ -335,83 +374,123 @@
...
-                # write the data to the ms.
-                main_dict = ms_extra.populate_main_dict(uvw_coordinates, vis_data, flag_data,
-                                                        out_mjd, ant1_index, ant2_index,
-                                                        dump_time_width, field_id, state_id,
-                                                        scan_itr, model_data, corrected_data)
-                ms_extra.write_rows(main_table, main_dict, verbose=options.verbose)
+        # Iterate over time in some multiple of dump average
+        ntime = utc_seconds.size
+        tsize = dump_av
+        ntime_av = ntime/tsize
```
mauch: `ntime_av` is only there to calculate the number of averaged timestamps for logging some output as h5toms runs. Previously it was calculated from the shape of the averaged timestamps directly, but now as the averager no longer sees all of the timestamps for a scan at once I compute it from the averaging length and the scan length. No output gets printed if `ntime_av==0` so it makes no difference in that case.
It's probably best that I move the assignment of `ntime_av` to just before the output gets printed so as to avoid confusion.

mauch: Actually it just occurred to me a neater way is to increment `ntime_av` in each loop by the number of timestamps left in the data after averaging. This is what I did in the commit I just made.

sjperkins: Agreed

sjperkins:
Looks good to me!

mauch:
I'm also now happy with this. Dirty images made using measurement sets from h5toms.py are identical between this branch and master. The only difference is that in the 'time ordered' case wsclean and casa work significantly faster.

@ludwigschwardt this PR was originally assigned to you. Perhaps you might want to take a look as well?

scripts/h5toms.py
sjperkins: @ludwigschwardt My previous comment is the only outstanding question I have on this PR

PR #65, opened 2016-11-28
Prefer sdp_l0_bls_ordering to cbf_bls_ordering

bmerry:
If `sdp_l0_bls_ordering` is present in the TelescopeState, it is used instead of `bls_ordering` in the CBF component of the telescope model.

This depends on ska-sa/katsdpdata#55.


ludwigschwardt:
The `rotate_bls` option might be needed again in future---who knows? :-)

But I'm keeping it as is, thanks!

PR #70, opened 2017-01-03
Correct creation of model and corrected data

sjperkins:
https://github.com/ska-sa/katdal/pull/64 modified h5toms to output visibility data in time ordered format. The optional creation of MODEL_DATA and CORRECTED_DATA columns was not modified to handle this new ordering, leading to table conformance errors caused by incorrectly shaped arrays.

This commit moves the creation MODEL_DATA and CORRECTED_DATA forward in the script to the point where the shape of vis_data has been reshaped into the expected (ntime\*nbl, nchan, npol) ordering, instead of the previous location where the shape of vis_data was (ntime, nchan, nbl\*npol).

### DIFF
--- a/scripts/h5toms.py
+++ b/scripts/h5toms.py
@@ -421,14 +421,6 @@ for win in range(len(h5.spectral_windows)):
             #Increment the number of averaged dumps
             ntime_av += tdiff
 
-            model_data = None
-            corrected_data = None
-            if options.model_data:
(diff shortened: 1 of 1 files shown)


sjperkins:
We've tested this extensively since we need these columns for the automated imaging pipeline. Doing a cheeky merge. May the maintainer forgive me.

ludwigschwardt:
Forgiven! 😀

PR #71, opened 2017-01-09
Don't crash if there are no ants or corr products

ludwigschwardt:
This fixes places in DataSet that would crash if the list of antennas
(or correlation products) is empty. The last place affects printing
of the data set and the rest are extraction of sensor values that
should be empty arrays.

### DIFF
--- a/katdal/dataset.py
+++ b/katdal/dataset.py
@@ -455,9 +455,10 @@ class DataSet(object):
                                                             (self.size / 1e6, 'MB') if self.size > 1e6 else
                                                             (self.size / 1e3, 'KB'))]))
         autocorrs = np.array([(inpA[:-1] == inpB[:-1]) for inpA, inpB in self.corr_products])
+        crosscorrs = np.array([(inpA[:-1] != inpB[:-1]) for inpA, inpB in self.corr_products])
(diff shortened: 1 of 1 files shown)


katdal/dataset.py
```diff
@@ -1018,14 +1019,29 @@ def lst(self):
...
+        return np.column_stack([sensor_data(ant.name) for ant in self.ants]) \
+            if self.ants else np.zeros((self.shape[0], 0))
+
+    def _sensor_per_corrprod(self, base_name):
+        """Extract a single sensor per corrprod and safely stack the results."""
+        def sensor_data(antA, antB):
+            return self.sensor['Antennas/%s/%s_%s' % (antA, base_name, antB)]
+        return np.column_stack([sensor_data(inpA[:-1], inpB[:-1])
+                                for inpA, inpB in self.corr_products]) \
+            if self.corr_products else np.zeros((self.shape[0], 0))
```
spassmoor: This line is returning the error 

ValueError: The truth value of an array with more than one element is ambiguous. Use a.any() or a.all()

when there are core-prods and you are requesting    the u,v,w parameters 
e.g   h5.u

ludwigschwardt: Thank you for picking this up. Of course I only checked on the problematic corner case and not the normal files... I really need some unit tests here!

PR #80, opened 2017-03-20
Single pol

mauch:
Changes to allow h5toms to deal with single polarisation datasets by default.

I've removed the --HH and --VV options from h5toms.py. These previously would fake a Stokes I dataset using just a single pol, making CASA very unhappy as it looks for receptors that correspond to polarisations when it applies calibration models.
Instead I have now added a '--pols-to-use' option so that the user can specify which polarisations from the input h5 file that want in a (now correctly labelled) MS. The default behaviour is to look for any single pols in the input data and use these for the output (HH,VV if both are present or only one of these if that is all that is available).
I've added a bit of checking to make sure that --pols-to-use corresponds to pols that are available in the input file, and to enforce --full-pol in preference to any selected pols if it is used as an option.

@ludwigschwardt - this is for you.

### DIFF
--- a/scripts/h5toms.py
+++ b/scripts/h5toms.py
@@ -119,7 +119,8 @@ h5 = katdal.open(args, ref_ant=options.ref_ant)
 pols_in_file = list(np.unique([(cp[0][-1] + cp[1][-1]).upper() for cp in h5.corr_products]))
 
 #Which polarisation do we want to write into the MS
-#all possible pols if full-pol selected, otherwise the selected polarisations via pols_to_use.
+#all possible pols if full-pol selected, otherwise the selected polarisations via pols_to_use
+#finally select any of HH,VV present (the default).
 pols_to_use = ['HH', 'HV', 'VH', 'VV'] if (options.full_pol or options.circular) else \
               options.pols_to_use.split(',') if options.pols_to_use else \
               [pol for pol in ['HH','VV'] if pol in pols_in_file]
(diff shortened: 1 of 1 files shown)


scripts/h5toms.py
```diff
@@ -60,10 +60,8 @@
                   help="More verbose progress information")
 parser.add_option("-w", "--stop-w", action="store_true", default=False,
                   help="Use W term to stop fringes for each baseline")
-parser.add_option("-x", "--HH", action="store_true", default=False,
-                  help="Produce a Stokes I MeasurementSet using only HH")
-parser.add_option("-y", "--VV", action="store_true", default=False,
-                  help="Produce a Stokes I MeasurementSet using only VV")
+parser.add_option("-p", "--pols-to-use", default=None,
+                  help="Select polarisation products to include in MS from HH,VV,HV,VH, default is all available from HH,VV")
```
ludwigschwardt: Is this a comma-separated list of polarisation products, i.e. can you go "HV,VH" if you so wished?

ludwigschwardt: Also, does the order matter?

mauch: Yup, a comma separated list (I'll update the help text to be a bit more clear).

Indeed you can go HV,VH if you so desire- giving the user the freedom to do this was the reason for the change in the first place:)

The order shouldn't matter as its dealt with by the populate_polarisation_dict function. However I think that in all cases the order going into the MS will be HH, HV, VH, VV (which is what it was before) due to np.unique sorting the elements of its output array at line 125.

I tried making measurement sets in pathological cases like ones with only HV,VH as you mentioned above- and also ones with different ordering. CASA seemed happy with the files. BTW- XX is HH and YY is VV. This was already hard-coded in the populate_polarisation_dict function.

PR #82, opened 2017-04-07
Correct type to int for SOURCE_ID and correct dimensions of phase_cen…

mauch:
…ters.

So it looks like this change wasn't as innocuous as it first appeared. First of all CASA was unable to read the SOURCE table due to the SOURCE_ID's being floats rather than the required ints, then this opened up another much older problem in which the 'DIRECTION' field in the SOURCE table had 3 dimensions rather then the required 2 (as stated in Memo 229).

@spassmoor can you check you are happy with my corrections to yours
@ludwigschwardt you might also want to make sure you're happy with this.


ludwigschwardt:
How about fixing the rest of the entries with dubious floating-point types too (CALIBRATION_GROUP and NUM_LINES)?

PR #84, opened 2017-04-12
Fix upside-down MeerKAT images

bmerry:
There are two major changes:

1. When the spectral window uses lower sideband, conjugate the
visibilities on load, because the correlator/ingest do not correct for
it. This brings all data to the MeerKAT sign convention for
visibility phase. For v1 and v2 it's hardcoded (since they hardcode for
lower sideband), while for v3 it is dependent on the band.

2. Change the sign of UVW coordinates to match the CASA convention.

For MeerKAT the second change will make images made via h5toms come out
the right way up. For KAT-7 the two changes will cancel out when
imaging, although the sign conventions will now match MeerKAT.

Fixes SR-811.

### DIFF
--- a/katdal/dataset.py
+++ b/katdal/dataset.py
@@ -144,6 +144,7 @@ class SpectralWindow(object):
         self.channel_width = channel_width
         self.num_chans = num_chans
         self.product = product if product is not None else ''
+        self.sideband = sideband
         self.band = band
         # Don't subtract half a channel width as channel 0 is centred on 0 Hz in baseband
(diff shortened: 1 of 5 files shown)


bmerry:
@mauch maybe you can repeat your tests using this branch to make sure that everything is now right-way up?

mauch:
Will do.

mauch:
KAT-7 images are now have the same (correct) orientation pre- and
post-fixvis.

I'll check it for MeerKAT as well and let you know.

mauch:
MeerKAT images also now have the correct orientation before and after fixvis.

PR #85, opened 2017-05-23
Access sensor data in a TelescopeState object

ludwigschwardt:
This introduces the ```TelstateSensorData``` class which wraps raw (uninterpolated) sensor data stored in a ```TelescopeState``` object. The existing class with the same name is renamed to ```H5TelstateSensorData``` to highlight that it is wrapping telstate stored in HDF5 as opposed to the live database.

In the process, clarify the ```SensorData``` API, add the ```ComparableArrayWrapper``` class to support object-valued sensors like "cal_product_B" and clean up ```dtype``` support. This also required quite a bit of tinkering with ```CategoricalData``` and concatenated versions of these classes.

### DIFF
--- a/katdal/categorical.py
+++ b/katdal/categorical.py
@@ -104,7 +104,27 @@ class ComparableArrayWrapper(object):
 
 
 def infer_dtype(values):
-    """Figure out dtype of sequence of sensor values, or None if unavailable."""
+    """Figure out dtype of sequence of sensor values.
+
+    The common dtype is determined by implicit NumPy coercion. If the values
+    are array-like themselves, treat them as opaque objects to simplify
+    sensor processing. If the sequence is empty, the dtype is unknown and
+    set to None. In addition, short-circuit to an actual dtype for objects
+    with this attribute to simplify calling this on a mixed collection of
+    sensor data.
+
+    Parameters
+    ----------
+    values : sequence, or object with dtype
+        Sequence of sensor values (typically a list), or a sensor data object
+        with a dtype attribute (like ndarray or :class:`SensorData`)
+
+    Returns
+    -------
+    dtype : :class:`numpy.dtype` object or None
+        Inferred dtype, or None if `values` is an empty sequence
+
+    """
     # If values already has a dtype (because it is an ndarray, SensorData,
     # CategoricalData, etc), return that instead
     if hasattr(values, 'dtype'):
--- a/katdal/categorical.py
+++ b/katdal/categorical.py
@@ -95,38 +95,44 @@ class HashableValueWrapper(object):
         return pickle.loads(self.hashable_value)
 
 
-def unique_in_order(elements, return_inverse=False):
+def unique_in_order(elements, unwrap=False, return_inverse=False):
     """Extract unique elements from *elements* while preserving original order.
 
     Parameters
     ----------
-    elements : sequence
-        Sequence of sortable objects (i.e. objects supporting the < operator)
+    elements : sequence of hashables
+        Sequence of hashable objects (i.e. objects supporting __hash__ operator)
+    unwrap : {False, True}, optional
+        If True, unwrap any objects wrapped in a :class:`HashableValueWrapper`
     return_inverse : {False, True}, optional
         If True, also return sequence of indices that can be used to reconstruct
         original `elements` sequence via `unique_elements[inverse]`
 
     Returns
     -------
-    unique_elements : array
-        Array of unique objects, in original order they were found in `elements`
+    unique_elements : list
+        List of unique objects, in original order they were found in `elements`
     inverse : array of int, optional
         If `return_inverse` is True, sequence of indices that can be used to
         reconstruct original sequence
 
     """
-    elements = np.atleast_1d(elements)
-    # Ensure that objects support the appropriate comparison operators for sorting (i.e. <)
-    if (elements.dtype == np.object) and not hasattr(elements[0], '__lt__') and not hasattr(elements[0], '__cmp__'):
-        raise TypeError("Cannot identify unique objects of %s as it has no __lt__ or __cmp__ methods" %
-                        (elements[0].__class__,))
-    # Find unique elements based on sorting
-    unique_elements, indices = np.unique(elements, return_inverse=True)
-    # Shuffle unique values back into their original order as they were found in elements
-    indices_list = indices.tolist()
-    original_order = np.argsort([indices_list.index(n) for n in range(len(unique_elements))])
-    return (unique_elements[original_order], original_order.argsort()[indices]) \
-        if return_inverse else unique_elements[original_order]
+    # Surprisingly, a zero generator like itertools.repeat does not buy you anything
+    lookup = collections.OrderedDict(zip(elements, len(elements) * [0]))
+    for index, element in enumerate(lookup):
+        lookup[element] = index
+    if unwrap:
+        def maybe_unwrap(v):   # noqa: E301
+            return v.unwrap() if isinstance(v, HashableValueWrapper) else v
+        unique_elements = [maybe_unwrap(element) for element in lookup]
+    else:
+        unique_elements = lookup.keys()
+    if return_inverse:
+        inverse = np.array([lookup[element] for element in elements], dtype=np.int)
+        return unique_elements, inverse
+    else:
+        return unique_elements
+
 
 # -------------------------------------------------------------------------------------------------
 # -- CLASS :  CategoricalData
@@ -137,12 +143,12 @@ class CategoricalData(object):
     """Container for categorical (i.e. non-numerical) sensor data.
 
     This container allows simple manipulation and interpolation of a time series
-    of non-numerical data represented as discrete events. The data are stored in
-    three arrays:
+    of non-numerical data represented as discrete events. The data is stored as
+    a list of sensor values and two integer arrays:
 
-    * `events` stores the time indices (dumps) where each event occurs
     * `unique_values` stores one copy of each unique object in the data series
-    * `indices` stores indices linking each event to the `unique_values` array
+    * `events` stores the time indices (dumps) where each event occurs
+    * `indices` stores indices linking each event to the `unique_values` list
 
     The __getitem__ interface (i.e. `data[dump]`) returns the data associated
     with the last event before the requested dump(s), in effect doing a
@@ -162,8 +168,8 @@ class CategoricalData(object):
 
     Attributes
     ----------
-    unique_values : array, shape (*M*,)
-        Array of unique sensor values, in order they were found in `sensor_values`
+    unique_values : list, length *M*
+        List of unique sensor values in order they were found in `sensor_values`
     indices : array of int, shape (*N*,)
         Array of indices into `unique_values`, one per sensor event
 
@@ -175,10 +181,24 @@ class CategoricalData(object):
     dummy object value and any Nones entering through this initialiser will
     probably not cause any issues.
 
+    It is better to make `unique_values` a list instead of an array because an
+    array assimilates objects such as tuples, lists and other arrays. The
+    alternative is an array of :class:`HashableValueWrapper` objects but these
+    then need to be unpacked at some later stage which is also tricky.
+
     """
     def __init__(self, sensor_values, events):
-        self.unique_values, self.indices = unique_in_order(sensor_values, return_inverse=True)
+        self.unique_values, self.indices = unique_in_order(sensor_values, unwrap=True,
+                                                           return_inverse=True)
         self.events = np.asarray(events)
+
+    @property
+    def _hashable_values(self):
+        """Hashable version of unique values, wrapping any objects."""
+        if self.dtype == np.object:
+            return [HashableValueWrapper.wrap(value) for value in self.unique_values]
+        else:
+            return self.unique_values
 
     def _lookup(self, dumps):
         """Look up relevant indices occurring at specified *dumps*.
@@ -210,8 +230,8 @@ class CategoricalData(object):
 
         Returns
         -------
-        val : array
-            Sensor values at selected dumps
+        val : object or list of objects
+            Sensor values at selected dumps, either single value or list of them
 
         """
         if isinstance(key, slice):
@@ -220,12 +240,17 @@ class CategoricalData(object):
         # Convert sequence of bools (one per dump) to sequence of indices where key is True
         elif np.asarray(key).dtype == np.bool and len(np.asarray(key)) == self.events[-1]:
             key = np.nonzero(key)[0]
-        return self.unique_values[self._lookup(key)]
+        indices = self._lookup(key)
+        # Interpret indices as either a sequence of ints or a single int
+        try:
+            return [self.unique_values[index] for index in indices]
+        except TypeError:
+            return self.unique_values[indices]
 
     def __repr__(self):
         """Short human-friendly string representation of categorical data object."""
         return "<katdal.CategoricalData events=%d values=%d type='%s' at 0x%x>" % \
-               (len(self.indices), len(self.unique_values), self.unique_values.dtype, id(self))
+               (len(self.indices), len(self.unique_values), self.dtype, id(self))
 
     def __str__(self):
         """Long human-friendly string representation of categorical data object."""
@@ -237,58 +262,49 @@ class CategoricalData(object):
         """Length operator indicates number of events produced by sensor."""
         return len(self.indices)
 
+    def _bool_per_dump(self, bool_per_value):
+        """Turn list of bools per unique value into an array of bools per dump."""
+        bool_per_event = np.atleast_1d(np.array(bool_per_value)[self.indices])
+        bool_per_dump = np.empty(self.events[-1], dtype=np.bool)
+        for n, (start, end) in enumerate(zip(self.events[:-1], self.events[1:])):
+            bool_per_dump[start:end] = bool_per_event[n]
+        return bool_per_dump
+
     def __eq__(self, other):
         """Equality comparison operator."""
-        index_truths = np.atleast_1d((self.unique_values == other)[self.indices])
-        result = np.empty(self.events[-1], dtype=np.bool)
-        for n, (start, end) in enumerate(zip(self.events[:-1], self.events[1:])):
-            result[start:end] = index_truths[n]
-        return result
+        return self._bool_per_dump([value == other for value in self._hashable_values])
 
     def __ne__(self, other):
         """Inequality comparison operator."""
-        index_truths = np.atleast_1d((self.unique_values != other)[self.indices])
-        result = np.empty(self.events[-1], dtype=np.bool)
-        for n, (start, end) in enumerate(zip(self.events[:-1], self.events[1:])):
-            result[start:end] = index_truths[n]
-        return result
+        return self._bool_per_dump([value != other for value in self._hashable_values])
 
     def __lt__(self, other):
         """Less-than comparison operator."""
-        index_truths = np.atleast_1d((self.unique_values < other)[self.indices])
-        result = np.empty(self.events[-1], dtype=np.bool)
-        for n, (start, end) in enumerate(zip(self.events[:-1], self.events[1:])):
-            result[start:end] = index_truths[n]
-        return result
+        return self._bool_per_dump([value < other for value in self._hashable_values])
 
     def __gt__(self, other):
         """Greather-than comparison operator."""
-        index_truths = np.atleast_1d((self.unique_values > other)[self.indices])
-        result = np.empty(self.events[-1], dtype=np.bool)
-        for n, (start, end) in enumerate(zip(self.events[:-1], self.events[1:])):
-            result[start:end] = index_truths[n]
-        return result
+        return self._bool_per_dump([value > other for value in self._hashable_values])
 
     def __le__(self, other):
         """Less-than-or-equal comparison operator."""
-        index_truths = np.atleast_1d((self.unique_values <= other)[self.indices])
-        result = np.empty(self.events[-1], dtype=np.bool)
-        for n, (start, end) in enumerate(zip(self.events[:-1], self.events[1:])):
-            result[start:end] = index_truths[n]
-        return result
+        return self._bool_per_dump([value <= other for value in self._hashable_values])
 
     def __ge__(self, other):
         """Greater-than-or-equal comparison operator."""
-        index_truths = np.atleast_1d((self.unique_values >= other)[self.indices])
-        result = np.empty(self.events[-1], dtype=np.bool)
-        for n, (start, end) in enumerate(zip(self.events[:-1], self.events[1:])):
-            result[start:end] = index_truths[n]
-        return result
+        return self._bool_per_dump([value >= other for value in self._hashable_values])
 
     @property
     def dtype(self):
         """Sensor value type."""
-        return self.unique_values.dtype
+        if self.unique_values:
+            # Put all values into array to find maximum length of any string type
+            test_data = np.array(self.unique_values)
+            # Beware array-valued sensors; treat their values as opaque objects
+            # This forces sensor values to be 1-D at all times (an invariant)
+            return test_data.dtype if test_data.ndim == 1 else np.object
+        else:
+            return np.object
 
     def segments(self):
         """Generator that iterates through events and returns segment and value.
@@ -320,12 +336,14 @@ class CategoricalData(object):
 
         """
         # If value has not been seen before, add it to unique_values (and create new index for it)
-        value_index = np.nonzero(self.unique_values == value)[0] if value is not None else [self._lookup(event)]
-        if len(value_index) == 0:
-            value_index = len(self.unique_values)
-            self.unique_values = np.r_[self.unique_values, [value]]
+        if value is not None:
+            try:
+                value_index = self._hashable_values.index(value)
+            except ValueError:
+                value_index = len(self.unique_values)
+                self.unique_values += [value]
         else:
-            value_index = value_index[0]
+            value_index = self._lookup(event)
         # If new event coincides with existing event, simply change value of that event, else insert new event
         event_index = self.events.searchsorted(event)
         before, after = event_index, (event_index + 1 if self.events[event_index] == event else event_index)
@@ -335,21 +353,25 @@ class CategoricalData(object):
     def remove(self, value):
         """Remove sensor value, remapping indices and merging segments in process.
 
+        If the sensor value does not exist, do nothing.
+
         Parameters
         ----------
         value : object
             Sensor value to remove from container
 
         """
-        index = np.nonzero(self.unique_values == value)[0]
-        if len(index) > 0:
-            index = index[0]
+        try:
+            index = self._hashable_values.index(value)
+        except ValueError:
+            pass
+        else:
             keep = (self.indices != index)
             remap = np.arange(len(self.unique_values))
             remap[index:] -= 1
             self.indices = remap[self.indices[keep]]
             self.events = np.r_[self.events[keep], self.events[-1]]
-            self.unique_values = np.r_[self.unique_values[:index], self.unique_values[index + 1:]]
+            self.unique_values = self.unique_values[:index] + self.unique_values[index + 1:]
 
     def add_unmatched(self, segments, match_dist=1):
         """Add duplicate events for segment starts that don't match sensor events.
@@ -401,7 +423,7 @@ class CategoricalData(object):
         # When multiple sensor events are associated with the same segment, only keep the final one
         final = np.nonzero(np.diff(events) > 0)[0]
         subset, self.indices = np.unique(self.indices[final], return_inverse=True)
-        self.unique_values = self.unique_values[subset]
+        self.unique_values = [self.unique_values[index] for index in subset]
         self.events = np.r_[events[final], events[-1]]
 
     def partition(self, segments):
@@ -481,9 +503,10 @@ def concatenate_categorical(split_data, **kwargs):
     segments = np.cumsum([0] + [cat_data.events[-1] for cat_data in split_data])
     data = CategoricalData([], [])
     # Combine all unique values in the order they are found in datasets
-    split_values = [cat_data.unique_values for cat_data in split_data]
+    split_values = [cat_data._hashable_values for cat_data in split_data]
     inverse_splits = np.cumsum([0] + [len(vals) for vals in split_values])
-    data.unique_values, inverse = unique_in_order(np.concatenate(split_values), return_inverse=True)
+    data.unique_values, inverse = unique_in_order(sum(split_values, []),
+                                                  unwrap=True, return_inverse=True)
     indices, events = [], []
     for n, cat_data in enumerate(split_data):
         # Remap indices to new unique_values array
@@ -545,6 +568,9 @@ def sensor_to_categorical(sensor_timestamps, sensor_values, dump_midtimes, dump_
     sensor_timestamps = np.atleast_1d(sensor_timestamps)
     sensor_values = np.atleast_1d(sensor_values)
     dump_midtimes = np.atleast_1d(dump_midtimes)
+    # Check if sensor values are objects wrapped in HashableValueWrappers
+    wrapped_values = len(sensor_values) and isinstance(sensor_values[0],
+                                                       HashableValueWrapper)
     # Convert sensor event times to dump indices (pick the dump during which each sensor event occurred)
     # The last event is fixed at one-past-the-last-dump, to indicate the end of the last segment
     events = np.r_[dump_midtimes.searchsorted(sensor_timestamps - 0.5 * dump_period), len(dump_midtimes)]
@@ -554,16 +580,24 @@ def sensor_to_categorical(sensor_timestamps, sensor_values, dump_midtimes, dump_
     sensor_values, events = sensor_values[non_empty], np.r_[events[non_empty], len(dump_midtimes)]
     # Force first dump to have valid sensor value (use initial value or first proper value advanced to start)
     if events[0] != 0 and initial_value is not None:
+        if wrapped_values:
+            initial_value = HashableValueWrapper.wrap(initial_value)
         sensor_values, events = np.r_[[initial_value], sensor_values], np.r_[0, events]
     events[0] = 0
     # Apply optional transform to sensor values
-    sensor_values = np.array([transform(y) for y in sensor_values]) if transform is not None else sensor_values
+    if transform is not None:
+        if wrapped_values:
+            unwrapped_transform = transform
+            transform = lambda value: HashableValueWrapper.wrap(unwrapped_transform(value.unwrap()))   # noqa: E731
+        sensor_values = np.array([transform(y) for y in sensor_values])
     # Discard sensor events that do not change the (transformed) sensor value (i.e. that repeat the previous value)
     if not allow_repeats:
         changes = [n for n in range(len(sensor_values)) if (n == 0) or (sensor_values[n] != sensor_values[n - 1])]
         sensor_values, events = sensor_values[changes], np.r_[events[changes], len(dump_midtimes)]
     # Extend segments with "greedy" values to include first dump of next segment where sensor value is changing
     if greedy_values is not None:
+        if wrapped_values:
+            greedy_values = [HashableValueWrapper.wrap(value) for value in greedy_values]
         greedy_to_nongreedy = [n for n in range(1, len(sensor_values))
                                if sensor_values[n - 1] in greedy_values and sensor_values[n] not in greedy_values]
         events[greedy_to_nongreedy] += 1
--- a/katdal/categorical.py
+++ b/katdal/categorical.py
@@ -45,8 +45,8 @@ class HashableValueWrapper(object):
     Note
     ----
     The current implementation uses pickle to make generic objects hashable,
-    even though it is not recommended - see e.g.
-    http://www.aminus.org/blogs/index.php/2007/11/03/pickle_dumps_not_suitable_for_hashing
+    even though it is not recommended - see e.g. this `blog post
+    <http://www.aminus.org/blogs/index.php/2007/11/03/pickle_dumps_not_suitable_for_hashing>`_.
     The main problem is with unordered data like dicts and sets, where the item
     order in the pickle string may differ even though the objects are equal;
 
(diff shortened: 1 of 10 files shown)


katdal/categorical.py
```diff
@@ -16,124 +16,135 @@
...
-    unique_indices : ndarray, optional
-        The indices of the first occurrences of the unique values in the
-        (flattened) original array. Only provided if `return_index` is True.
-    unique_inverse : ndarray, optional
-        The indices to reconstruct the (flattened) original array from the
-        unique array. Only provided if `return_inverse` is True.
+    Note
+    ----
+    The current implementation uses pickle to make generic objects hashable,
+    even though it is not recommended (see http://bit.ly/2ruNkH2). The main
```
bmerry: Rather put in the original URL, so that this won't break if bit.ly goes down or goes out of business.

katdal/categorical.py
```diff
@@ -16,124 +16,135 @@
...
-    unique_elements, indices = unique(elements, return_inverse=True)
-    # Shuffle unique values back into their original order as they were found in elements
-    indices_list = indices.tolist()
-    original_order = np.argsort([indices_list.index(n) for n in range(len(unique_elements))])
-    return (unique_elements[original_order], original_order.argsort()[indices]) \
-        if return_inverse else unique_elements[original_order]
+    # Surprisingly, a zero generator like itertools.repeat does not buy you anything
+    lookup = collections.OrderedDict(zip(elements, len(elements) * [0]))
+    for index, element in enumerate(lookup):
+        lookup[element] = index
```
bmerry: Aren't these three lines equivalent to
```python
lookup = collections.OrderedDict(zip(elements, range(len(elements))))
```

katdal/categorical.py
```diff
@@ -544,6 +582,9 @@ def sensor_to_categorical(sensor_timestamps, sensor_values, dump_midtimes, dump_
     sensor_timestamps = np.atleast_1d(sensor_timestamps)
     sensor_values = np.atleast_1d(sensor_values)
     dump_midtimes = np.atleast_1d(dump_midtimes)
+    # Check if sensor values are objects wrapped in HashableValueWrappers
+    wrapped_values = len(sensor_values) and isinstance(sensor_values[0],
+                                                       HashableValueWrapper)
```
bmerry: This function seems to be complicated by the need to do alternative versions of all the processing which does un-wrapping and rewrapping. Wouldn't it be simpler to do:
1. Unwrap if wrapped.
2. Do the operations on the raw values (avoiding putting them into numpy).
3. Rewrap if appropriate.

katdal/concatdata.py
```diff
@@ -206,18 +206,17 @@ class ConcatenatedSensorData(SensorData):
...
-    but provides quick access to metadata such as sensor dtype, name and number
-    of data points.
+    but provides quick access to metadata such as sensor dtype and name.
 
     Parameters
     ----------
-    data : sequence of recarray-like with fields ('timestamp', 'value', 'status')
-        Uncached sensor data as a list of record arrays or equivalent (such as a
-        :class:`h5py.Dataset`)
+    data : sequence of recarray-like with fields 'timestamp', 'value' ('status')
```
bmerry: Not sure what this means now. Is 'status' optional?

katdal/concatdata.py
```diff
@@ -233,11 +232,13 @@ def __init__(self, data):
...
         """Extract timestamp, value and status of each sensor data point."""
-        return np.concatenate([sd[key] for sd in self.data])
+        return np.concatenate([sd[key] for sd in self._data])
 
-    def __len__(self):
-        """Number of sensor data points."""
-        return np.sum([len(sd) for sd in self.data])
+    def __bool__(self):
+        """True if sensor has at least one data point."""
+        return any([bool(sd) for sd in self._data])
```
bmerry: Doesn't need the square brackets: `any` can take a generator comprehension.

bmerry: Also, is this going to work given that `sd` is presumably an ndarray, on which `bool` doesn't do what one would normally assume?

katdal/sensordata.py
```diff
@@ -55,50 +55,110 @@
...
     ----------
-    data : recarray-like, with fields ('timestamp', 'value', 'status')
-        Uncached sensor data as record array or equivalent (such as a
-        :class:`h5py.Dataset`)
-    name : string or None, optional
-        Sensor name (assumed to be data.name by default, if it exists)
+    name : string
+        Sensor name
+    dtype : :class:`numpy.dtype` object
+        Sensor data type as NumPy dtype (beware that string lengths may be wrong)
```
bmerry: I don't follow the string length caveat. It's the sort of comment I'd expect on an attribute or return value, not on an input parameter.

katdal/sensordata.py
```diff
@@ -55,50 +55,110 @@
...
+
```
bmerry: Remove this blank line from the docstring (PEP 257).

katdal/sensordata.py
```diff
@@ -55,50 +55,110 @@
...
+        self.name = name
+        self.dtype = dtype
+
+    def __getitem__(self, key):
+        """Extract timestamp and value (and status) of each sensor data point.
+
+        Parameters
+        ----------
+        key : {'timestamp', 'value', 'status'}
+            Name of field to access ('status' is optional and may raise error)
```
bmerry: Which exception?

katdal/sensordata.py
```diff
@@ -55,50 +55,110 @@
...
+        -------
+        field : :class:`numpy.ndarray` object, shape (N,)
+            Requested field as 1-D array of appropriate dtype (float,
+            self.dtype or string, respectively, for timestamp / value / status)
+
+        Raises
+        ------
+        ValueError
+            If *key* is unsupported field name
+
```
bmerry: Remove blank line from docstring.

katdal/sensordata.py
```diff
@@ -126,34 +186,112 @@ class TelstateSensorData(SensorData):
     Parameters
     ----------
     data : recarray-like, with fields ('timestamp', 'value')
-        Uncached sensor data as record array or equivalent (such as a
-        :class:`h5py.Dataset`)
+        Uninterpolated sensor data as structured array or equivalent (such as
+        an :class:`h5py.Dataset`)
```
bmerry: Should they be wrapped or unwrapped at this point?

katdal/sensordata.py
```diff
@@ -126,34 +186,112 @@ class TelstateSensorData(SensorData):
...
+            raise KeyError('No sensor named %r in telstate (key not found)' %
+                           (name,))
+        if telstate.is_immutable(name):
+            raise KeyError("No sensor named %r in telstate (it's an attribute)" %
+                           (name,))
+        test_data = np.array([telstate[name]])
+        # Beware array-valued sensors; treat their values as opaque objects
+        # otherwise they will turn 'value' into a mega-array of e.g. float.
+        # This forces sensor values to be 1-D at all times (an invariant).
+        dtype = test_data.dtype if test_data.ndim == 1 else np.object
```
bmerry: This logic (and the comments) seem to be repeated in a few places. Perhaps it should be put into a helper function? It also has the potential to be optimised later in a way that doesn't require constituting the full array.

katdal/sensordata.py
```diff
@@ -126,34 +186,112 @@ class TelstateSensorData(SensorData):
...
+        super(TelstateSensorData, self).__init__(name, dtype)
+
+    def __bool__(self):
+        """True if sensor has at least one data point."""
+        return self.name in self._telstate and \
+            not self._telstate.is_immutable(self.name)
```
bmerry: Are either of these checks necessary, given that the constructor raises if not?

katdal/sensordata.py
```diff
@@ -249,37 +388,48 @@ def dummy_sensor_data(name, value=None, dtype=np.float64, timestamp=0.0):
...
 
 
-def remove_duplicates(sensor):
-    """Remove duplicate timestamp values from sensor data.
+def remove_duplicates_and_failures(sensor):
+    """Remove duplicate timestamp values and failures from sensor data.
 
     This sorts the 'timestamp' field of the sensor record array and removes any
     duplicate values, updating the corresponding 'value' and 'status' fields as
     well. If more than one timestamp have the same value, the value and status
```
bmerry: "one timestamp have" -> "one timestamp has'

katdal/sensordata.py
```diff
@@ -249,37 +388,48 @@ def dummy_sensor_data(name, value=None, dtype=np.float64, timestamp=0.0):
...
     of the last of these timestamps are selected. If the values differ for the
     same timestamp, a warning is logged (and the last one is still picked).
 
+    In addition, if there is a 'status' field, get rid of data with a 'failure'
+    status which indicates that the sensor could not be read and the
+    corresponding value will therefore be meaningless. Afterwards, remove the
+    'status' field from the data as this is the only place it plays a role.
```
bmerry: There are multiple statuses that mean the value is meaningless (see page 20 of the [katcp spec](https://pythonhosted.org/katcp/_downloads/NRF-KAT7-6.0-IFCE-002-Rev5.pdf)). It's better to assume that anything that isn't one of `nominal`, `warn`, `error` has no meaningful value.

katdal/categorical.py
ludwigschwardt: Done.

katdal/categorical.py
```diff
@@ -16,124 +16,135 @@
...
     ----------
-    elements : sequence
-        Sequence of sortable objects (i.e. objects supporting the < operator)
+    elements : sequence of hashables
+        Sequence of hashable objects (i.e. objects supporting __hash__ operator)
+    unwrap : {False, True}, optional
+        If True, unwrap any objects wrapped in a :class:`HashableValueWrapper`
     return_inverse : {False, True}, optional
         If True, also return sequence of indices that can be used to reconstruct
         original `elements` sequence via `unique_elements[inverse]`
```
ludwigschwardt: Changed wording.

katdal/categorical.py
```diff
@@ -236,58 +276,49 @@ def __len__(self):
...
         """Sensor value type."""
-        return self.unique_values.dtype
+        if self.unique_values:
+            # Put all values into array to find maximum length of any string type
+            test_data = np.array(self.unique_values)
+            # Beware array-valued sensors; treat their values as opaque objects
+            # This forces sensor values to be 1-D at all times (an invariant)
+            return test_data.dtype if test_data.ndim == 1 else np.object
+        else:
+            return np.object
```
ludwigschwardt: I was trying to avoid calling this on construction. The ```_hashable_values``` attribute will typically be accessed infrequently or not at all (only on comparisons, add and remove), while construction will necessarily happen.

Actually, a more relevant concern is that the ```concatenate_categorical``` function constructs a CategoricalData by assigning ```unique_values``` directly, bypassing the constructor for efficiency reasons.

Maybe the best place is within a setter on ```unique_values```.

katdal/categorical.py
```diff
@@ -334,21 +367,25 @@ def add(self, event, value=None):
     def remove(self, value):
         """Remove sensor value, remapping indices and merging segments in process.
```
ludwigschwardt: Indeed :-(

katdal/categorical.py
```diff
@@ -334,21 +367,25 @@ def add(self, event, value=None):
...
+        except ValueError:
+            pass
+        else:
             keep = (self.indices != index)
             remap = np.arange(len(self.unique_values))
             remap[index:] -= 1
             self.indices = remap[self.indices[keep]]
             self.events = np.r_[self.events[keep], self.events[-1]]
-            self.unique_values = np.r_[self.unique_values[:index], self.unique_values[index + 1:]]
+            self.unique_values = self.unique_values[:index] + self.unique_values[index + 1:]
```
ludwigschwardt: Agreed.

I always forget about ```del``` since it seems like the most out-of-place operator in Python...

katdal/categorical.py
```diff
@@ -480,9 +517,10 @@ def concatenate_categorical(split_data, **kwargs):
     segments = np.cumsum([0] + [cat_data.events[-1] for cat_data in split_data])
     data = CategoricalData([], [])
     # Combine all unique values in the order they are found in datasets
-    split_values = [cat_data.unique_values for cat_data in split_data]
+    split_values = [cat_data._hashable_values for cat_data in split_data]
     inverse_splits = np.cumsum([0] + [len(vals) for vals in split_values])
-    data.unique_values, inverse = unique_in_order(np.concatenate(split_values), return_inverse=True)
+    data.unique_values, inverse = unique_in_order(sum(split_values, []),
```
ludwigschwardt: Unfortunately not. The ```unique_in_order``` function potentially runs through the sequence twice: once to find all the unique elements, and then again to find their associated indices if return_inverse is True.

katdal/concatdata.py
ludwigschwardt: Yes, since telstate versions don't have it.

katdal/concatdata.py
ludwigschwardt: ```sd``` is actually a SensorData object with a recarray-like interface. It used to also take an ```h5py.Dataset``` directly but I don't see that anywhere in the code anymore. So the chances are low that ```sd``` will be an ndarray.

katdal/sensordata.py
ludwigschwardt: It *is* an attribute on the base class that is set by the derived classes, and this is its only docstring. I wanted to highlight that the dtype detection is a bit iffy when working with variable-length strings as classes typically only check the dtype of the first or last element to be efficient, and the resulting string length might be too short for other strings in the dataset.

For this very reason I am thinking of ditching ```dtype``` as a required parameter and only allowing it on ```RecordSensorData```. The sensor extractor funnels everything through this class on the way to ```CategoricalData``` or ndarrays (see ```remove_duplicates_and_failures``` function).

katdal/sensordata.py
ludwigschwardt: Oops, bad 12-year-old habit...

katdal/sensordata.py
ludwigschwardt: ValueError, as mentioned in the section below.

katdal/sensordata.py
ludwigschwardt: All SensorData contents will be wrapped. It's only once it hits CategoricalData that it is unwrapped.

The big picture is this:

telstate / HDF5: raw input format

SensorData: just a wrapper per (unextracted) sensor to give a common API to various input formats
CategoricalData / ndarray: replaces SensorData after extraction, contains actual data
SensorCache: dict-like collection of the above (both extracted and unextracted sensors mixed together)

katdal/sensordata.py
ludwigschwardt: Sounds good.

katdal/sensordata.py
ludwigschwardt: I could go ```return True```, yes. I thought of that. It felt too easy :-)

katdal/sensordata.py
ludwigschwardt: Done.

katdal/sensordata.py
ludwigschwardt: Good point, wasn't sure of the others, fixed now.

katdal/concatdata.py
bmerry: That wasn't completely clear from the comment. Maybe "'timestamp', 'value', and optionally 'status'"

katdal/categorical.py
bmerry: The fact that it's used in `add` seems worrying - it means that constructing a set of events via calls to `add` will have quadratic time.

katdal/sensordata.py
bmerry: It wasn't totally clear that "unsupported" also included "optional and absent". Might as well say `ValueError` instead of `error` here.

katdal/sensordata.py
```diff
@@ -55,50 +55,110 @@
...
+        Returns
+        -------
+        field : :class:`numpy.ndarray` object, shape (N,)
+            Requested field as 1-D array of appropriate dtype (float,
+            self.dtype or string, respectively, for timestamp / value / status)
+
+        Raises
+        ------
+        ValueError
+            If *key* is unsupported field name
```
bmerry: Variables should be enclosed in backticks, not stars (see https://github.com/numpy/numpy/blob/master/doc/HOWTO_DOCUMENT.rst.txt).

katdal/sensordata.py
bmerry: What exactly does "extraction" mean?

katdal/categorical.py
```diff
@@ -45,9 +45,10 @@ class HashableValueWrapper(object):
     Note
     ----
     The current implementation uses pickle to make generic objects hashable,
-    even though it is not recommended (see http://bit.ly/2ruNkH2). The main
-    problem is with unordered data like dicts and sets, where the item order
-    in the pickle string may differ even though the objects are equal. E.g.
+    even though it is not recommended - see e.g.
+    http://www.aminus.org/blogs/index.php/2007/11/03/pickle_dumps_not_suitable_for_hashing
```
bmerry: Missing full stop at the end of the sentence. This could also be a RST link rather than inline.

katdal/sensordata.py
```diff
@@ -454,10 +453,12 @@ def remove_duplicates_and_failures(sensor):
...
                          "keeping last one", katpoint.Timestamp(x[ind]).local(),
                          sensor.name, z[ind], z[replacement][ind])
-    # Remove entries where 'status' equals 'failure', if 'status' is present
+    # Remove entries where 'status' implies invalid values, if 'status' is present
     if z is not None:
         # Explicitly cast status to string type, as k7_augment produced sensors with integer statuses
-        unique_ind = unique_ind[z[unique_ind].astype('|S7') != 'failure']
+        status = z[unique_ind].astype('|S7')
+        has_valid_value = np.array([['nominal'], ['warn'], ['error']])
+        unique_ind = unique_ind[(status == has_valid_value).any(axis=0)]
```
bmerry: This seems a little complicated. How about `unique_ind[(status == 'nominal') | (status == 'warn') | (status == 'error')]`?

katdal/categorical.py
ludwigschwardt: I never intended ```add``` to be for bulk construction, but rather for tweaks to the sensor. Point taken though.

I think I'll actually move it to the constructor for now (as a setter on ```unique_values``` will be called there too).

katdal/concatdata.py
ludwigschwardt: Done.

katdal/sensordata.py
ludwigschwardt: Done.

katdal/categorical.py
ludwigschwardt: Fixed.

katdal/sensordata.py
ludwigschwardt: Much better, thanks!

katdal/sensordata.py
ludwigschwardt: The water is muddy. See e.g. http://blog.dolphm.com/pep257-good-python-docstrings-by-example/.

katdal/sensordata.py
ludwigschwardt: Fixed this and the rest of all the docstring issues.

bmerry:
Remind me: are you waiting for anything from me, or are you first looking at refactoring the dtype determination?

ludwigschwardt:
I'm still digesting your comments. Some avenues I'm exploring:

- I've figured out a way to sniff the basic type from the first few characters of the pickle. A pickle is a generic program, but hey, not really in practice :-) The basic types ('int', 'float', 'string', 'bool') start with specific codes, and anything else can be labelled an 'object'. The nice thing is that these string labels are directly usable as dtypes in NumPy. This would avoid unpickling the whole thing (not much difference in speed on small pickles, but orders faster on large ones like cal_product_B).

- I've investigated [joblib](http://pythonhosted.org/joblib/), from the same guy who brought us scikit-learn. This has two nifty features: a faster pickle for NumPy arrays and cryptographic hashing of NumPy arrays. Would the latter be preferable to hashing of pickle strings or does it have similar weaknesses?

- An alternative option to the latter is to do away with hashing and only check for equality when dealing with objects. That will force things back to a slower list index lookup instead of a dict lookup, but I don't expect object-valued sensors to have many events and also not many repeated values (consider gain solutions as typical).

Any thoughts welcome to save me from more rabbit holes...

bmerry:
On 29 May 2017 at 11:39, Ludwig Schwardt <notifications@github.com> wrote:

> I'm still digesting your comments. Some avenues I'm exploring:
>
>    -
>
>    I've figured out a way to sniff the basic type from the first few
>    characters of the pickle. A pickle is a generic program, but hey, not
>    really in practice :-) The basic types ('int', 'float', 'string', 'bool')
>    start with specific codes, and anything else can be labelled an 'object'.
>    The nice thing is that these string labels are directly usable as dtypes in
>    NumPy. This would avoid unpickling the whole thing (not much difference in
>    speed on small pickles, but orders faster on large ones like cal_product_B).
>
> Yuck! That sounds really horrific and easily breakable.


>
>
>    -
>
>    I've investigated joblib <http://pythonhosted.org/joblib/>, from the
>    same guy who brought us scikit-learn. This has two nifty features: a faster
>    pickle for NumPy arrays and cryptographic hashing of NumPy arrays. Would
>    the latter be preferable to hashing of pickle strings or does it have
>    similar weaknesses?
>
> I've used joblib before, for caching. I haven't looked at exactly how it
hashes numpy arrays, but it may be better, since it is probably canonical
(we'd just need to check that it incorporates the type information into the
hash, for example). I don't know if that solves the general object problem.

>
>
>    -
>
>    An alternative option to the latter is to do away with hashing and
>    only check for equality when dealing with objects. That will force things
>    back to a slower list index lookup instead of a dict lookup, but I don't
>    expect object-valued sensors to have many events and also not many repeated
>    values (consider gain solutions as typical).
>
> Is it worth trying to make a list of categorical sensors we expect and
classify them by size, frequency, and how likely they are to repeat? I'm
wondering if they can be separated into those that are (potentially large)
numpy arrays of continuous values where repetitions are unlikely and not
interesting (like gain solutions), and those that are small, not naturally
numpy arrays, and repeat in a predictable way (e.g. target descriptions,
activities).

Bruce
-- 
Bruce Merry
Senior Science Processing Developer
SKA South Africa

ludwigschwardt:
So, the latest thinking:

- Unpickle and chuck into an ndarray to determine the dtype as before. Fetching the pickle might be the bulk of the work and unpickling it not that onerous.
- Still wrap objects to prevent ndarray conflation and to provide a place to override __eq__ for ndarrays to avoid arrays of bools vs a single bool.
- Get rid of hashing and fall back to list.index lookup in ```unique_in_order``` (safer but slower, should be OK since object sensors are scarce and have few events).

katdal/categorical.py
```diff
@@ -17,107 +17,93 @@
...
+            return self.unwrapped == other
 
     def __ne__(self, other):
         """Inequality comparison operator."""
         return not self == other
 
     def __lt__(self, other):
         """Less-than comparison operator."""
-        raise TypeError("HashableValueWrapper objects are considered unorderable")
+        return self.unwrapped < other
```
bmerry: Shouldn't this be `self.unwrapped < other.unwrapped`, at least if `other` is a wrapper?

katdal/categorical.py
```diff
@@ -17,107 +17,93 @@
...
 def unique_in_order(elements, unwrap=False, return_inverse=False):
     """Extract unique elements from `elements` while preserving original order.
 
     Parameters
     ----------
-    elements : sequence of hashables
-        Sequence of hashable objects (i.e. objects supporting __hash__ operator)
+    elements : sequence of hashables or comparables
+        Sequence of hashable or comparable objects (i.e. objects supporting
+        either __hash__ or __eq__ operator)
```
bmerry: Hashable objects are required to implement equality comparison (see https://docs.python.org/2/glossary.html). I would say equality-comparable (since it's not necessary to support `<` etc). I'd also not refer to `__eq__`, since in Python 2 equality comparison can also be implemented by `__cmp__`.

katdal/categorical.py
```diff
@@ -17,107 +17,93 @@
...
     unwrap : {False, True}, optional
-        If True, unwrap any objects wrapped in a :class:`HashableValueWrapper`
+        If True, unwrap any objects wrapped in a :class:`ComparableArrayWrapper`
```
bmerry: This option doesn't seem like a natural fit for this function any more. It's currently two essentially unrelated functions (find unique elements, and unwrap wrapped values) bundled into one.

katdal/categorical.py
```diff
@@ -210,12 +208,9 @@ def __init__(self, sensor_values, events):
...
     @property
-    def _hashable_values(self):
-        """Hashable version of unique values, wrapping any objects."""
-        if self.dtype == np.object:
-            return [HashableValueWrapper.wrap(value) for value in self.unique_values]
-        else:
-            return self.unique_values
+    def _comparable_values(self):
+        """Comparable version of unique values, wrapping any objects."""
+        return [ComparableArrayWrapper.wrap(value) for value in self.unique_values]
```
bmerry: You removed the `.wrap` static method.

katdal/categorical.py
```diff
@@ -17,107 +17,93 @@
...
-    @staticmethod
-    def wrap(value):
-        """Wrap generic object to become hashable (pickle implementation)."""
-        return HashableValueWrapper(pickle.dumps(value))
-
-    def unwrap(self):
-        """Extract original object from wrapper (pickle implementation)."""
-        return pickle.loads(self.hashable_value)
+    # Force class to be unhashable to allow value comparison via __eq__ if possible
+    __hash__ = None
```
bmerry: What do you think of making `__hash__` return `hash(self.unwrapped)`? I admit I feel a bit odd about making a class that is only sometimes hashable, but it would let the fast path of `unique_in_order` work on a list of wrappers of hashables.

katdal/categorical.py
ludwigschwardt: Good point. I'll unwrap if needed. I added the rest of the comparison operators at the last minute, at first thinking of keeping the wrapper unorderable. Then I hit on the idea that I could wrap all types of sensor values for comparison purposes and that necessitated the change.

katdal/categorical.py
ludwigschwardt: I like this idea. At least ndarrays are not hashable and they will go the slow route. Equality and hashing will also be treated similarly for objects that support both.

katdal/categorical.py
ludwigschwardt: Done.

katdal/categorical.py
ludwigschwardt: I thought the same thing and it's already on the chopping block. The ```unwrap=True``` setting only occurs in two special places were it can easily be done outside the function.

katdal/categorical.py
ludwigschwardt: Fixed.

katdal/categorical.py
ludwigschwardt: The ```dtype``` property is now accessed infrequently in the latest version, which should make it more palatable. I'm still trying to get rid of it entirely.

katdal/categorical.py
ludwigschwardt: Fixed by ditching pickles for direct equality comparison.

katdal/sensordata.py
ludwigschwardt: Extraction is the combination of

- *getting* the raw sensor data from the store (telstate / HDF5),
- *interpolating* it onto the correlator data timestamps, and
- optionally *selecting* bits of it based on the currently active selection in the dataset.

The interpolated output (categorical or numerical) replaces the raw SensorData object in the SensorCache, meaning extraction typically only happens once per sensor.

katdal/sensordata.py
ludwigschwardt: Done in the latest commits.

katdal/sensordata.py
ludwigschwardt: The ```dtype``` machinery has been reworked and this comment is gone.

katdal/concatdata.py
ludwigschwardt: Confirmed that under normal circumstances ```sd``` will always be an instance of ```SensorData``` and not an ```ndarray```.

katdal/categorical.py
ludwigschwardt: The ```sensor_values``` variable is desired to be an ndarray to simplify manipulation, and that would break other things if it contains unwrapped values. Rather wrap the incoming parameters... At least the greedy values can now be compared directly with the new ```ComparableArrayWrapper```.

katdal/categorical.py
ludwigschwardt: I've not eliminated ```dtype``` but restricted its use. It should not be needed for "normal" operations that do not involve sensor creation or extraction. If this is still an issue the dtype can easily be cached in ```CategoricalData```.

katdal/categorical.py
```diff
@@ -16,124 +16,152 @@
...
+        return hash(self.unwrapped)
+
+    @staticmethod
+    def unwrap(v):
+        """Unwrap value if needed."""
+        return v.unwrapped if isinstance(v, ComparableArrayWrapper) else v
+
+
+def infer_dtype(values):
+    """Figure out dtype of sequence of sensor values, or None if unavailable."""
```
ludwigschwardt: I've improved the docstring.

There will basically always be a common dtype as determined by NumPy coercion, even if it is the fallback np.object. The dtype only remains unknown if there are no values to inspect.

katdal/concatdata.py
```diff
@@ -200,44 +201,57 @@ def _initial_dtype(self):
 # -------------------------------------------------------------------------------------------------
 
 
+def infer_common_dtype(sensor_data_sequence):
+    """Figure out common dtype of a sequence of objects with dtypes."""
```
ludwigschwardt: Renamed function and improved docstring.

katdal/categorical.py
```diff
@@ -236,58 +291,42 @@ def __len__(self):
         """Length operator indicates number of events produced by sensor."""
         return len(self.indices)
 
+    def _bool_per_dump(self, bool_per_value):
+        """Turn list of bools per unique value into an array of bools per dump."""
+        bool_per_event = np.atleast_1d(np.array(bool_per_value)[self.indices])
```
ludwigschwardt: I would say there are two typical categorical cases:

- a few small unique values with more events, e.g. mxxx_activity ('slew', 'track', 'slew', 'track')
- a few larger unique values with a few events, e.g. cal_product_B

This approach is informed by the former case, and doesn't really matter in the latter case where you have few of everything.

Or have I missed something?

katdal/sensordata.py
```diff
@@ -126,34 +190,96 @@ class TelstateSensorData(SensorData):
...
+            return values
+        else:
+            raise ValueError("Sensor %r data has no key '%s'" % (self.name, key))
+
+
+class TelstateSensorData(SensorData):
+    """Raw (uninterpolated) sensor data stored in original TelescopeState.
+
+    This wraps sensor data stored in a TelescopeState object. The data is
+    only read out on item access and then cached on the object. The caching
```
bmerry: Maybe explain the motivation for caching here i.e. to ensure that requesting timestamps and then values only loads the data once instead of twice. At first I was confused why this classes does caching while H5TelstateSensorData does not.

katdal/sensordata.py
```diff
@@ -249,37 +376,49 @@ def dummy_sensor_data(name, value=None, dtype=np.float64, timestamp=0.0):
             value = ''
         elif np.issubdtype(dtype, np.bool):
             value = False
+        elif np.issubdtype(dtype, np.object):
+            value = None
```
bmerry: This is a no-op, since value is already None to start with.

katdal/sensordata.py
```diff
@@ -249,37 +376,49 @@ def dummy_sensor_data(name, value=None, dtype=np.float64, timestamp=0.0):
...
     else:
         dtype = np.array(value).dtype
-    data = np.rec.fromarrays([[timestamp], [value], ['nominal']],
-                             dtype=[('timestamp', np.float64), ('value', dtype), ('status', '|S7')])
-    return SensorData(data, name)
+    data = np.array(zip([timestamp], [value]),
```
bmerry: `zip` seems overkill here - it's the same as `[(timestamp, value)]`.

katdal/sensordata.py
```diff
@@ -249,37 +376,49 @@ def dummy_sensor_data(name, value=None, dtype=np.float64, timestamp=0.0):
...
+                    dtype=[('timestamp', np.float64), ('value', dtype)])
+    return RecordSensorData(data, name)
```
bmerry: If the dtype is object, should the value be wrapped in `ComparableArrayWrapper`, or is this dummy used post-unwrapping?

katdal/concatdata.py
```diff
@@ -200,44 +201,78 @@ def _initial_dtype(self):
...
+
+    """
+    dtypes = [infer_dtype(sd) for sd in sensor_data_sequence]
+    # Ignore unavailable dtypes unless nothing is available
+    dtypes = [dt for dt in dtypes if dt is not None]
+    dtypes = unique_in_order(dtypes) if dtypes else [None]
+    if all(np.issubdtype(dtype, np.str) for dtype in dtypes):
+        # Strings of different lengths have different dtypes
+        # (e.g. '|S1' vs '|S10') but can be safely concatenated
+        dtypes = [np.dtype('|S%d' % (max([dt.itemsize for dt in dtypes]),))]
```
bmerry: The docstring implies that starting with a mix of int and string types, the string types should all be collapsed to the longest string type, but the code doesn't match that. Looking at the usage I suspect that isn't terribly important (you just want to know if there is a unique dtype), so it may be easier to just fix up the docstring.

This function would also be an ideal candidate for a unit test...

katdal/concatdata.py
```diff
@@ -200,44 +201,78 @@ def _initial_dtype(self):
...
     """
+
     def __init__(self, data):
-        self.data = data
+        self._data = data
         names = unique_in_order([sd.name for sd in data])
         if len(names) != 1:
-            raise ConcatenationError('Cannot concatenate sensors with different names: %s' % (names,))
+            raise ConcatenationError('Cannot concatenate sensor %r with different '
+                                     'underlying names: %s' % (self.name, names))
```
bmerry: `self.name` doesn't exist at this point.

katdal/concatdata.py
```diff
@@ -200,44 +201,78 @@ def _initial_dtype(self):
...
-        elif np.all([np.issubdtype(dtype, np.str) for dtype in dtypes]):
-            # Strings of different lengths have different dtypes (e.g. '|S1' vs '|S10') but can be safely concatenated
-            self.dtype = np.dtype('|S%d' % (max([dt.itemsize for dt in dtypes]),))
-        else:
-            raise ConcatenationError("Cannot concatenate sensor '%s' with different dtypes: %s" % (self.name, dtypes))
+        dtypes = infer_dtypes(data)
+        if len(dtypes) != 1:
+            raise ConcatenationError("Cannot concatenate sensor %r with different "
+                                     "dtypes: %s" % (self.name, dtypes))
+        self.dtype = dtypes[0]
```
bmerry: There should probably be a call to the base class constructor (via `__init__`) somewhere in here.

katdal/concatdata.py
```diff
@@ -267,18 +302,29 @@ class ConcatenatedSensorCache(SensorCache):
         be applied to sensor data (this can be disabled on data retrieval)
 
     """
+
     def __init__(self, caches, keep=None):
         self.caches = caches
         # Collect the names and types of all actual and virtual sensors in caches, as well as properties
```
bmerry: This comment needs to be updated given that actual now collects the sensor_data objects themselves.

katdal/concatdata.py
```diff
@@ -341,10 +417,11 @@ def get(self, name, select=False, extract=True, **kwargs):
...
         # Get array, categorical data or raw sensor data from each cache
-        split_data = [cache.get(name, select, extract, **kwargs) for cache in self.caches]
-        # If this sensor has already been partially extracted, we are forced to extract it in rest of caches too
-        if not extract and not np.all([isinstance(data, SensorData) for data in split_data]):
-            split_data = [cache.get(name, select, True, **kwargs) for cache in self.caches]
+        split_data = self._get(name, select, extract, **kwargs)
+        # If this sensor has already been partially extracted,
+        # we are forced to extract it in rest of caches too
+        if not extract and not all(isinstance(sd, SensorData) for sd in split_data):
+            split_data = self._get(name, select, True, **kwargs)
```
bmerry: In what cases would `isinstance(sd, SensorData)` be false? Did you mean `type(sd) is SensorData`?

katdal/sensordata.py
ludwigschwardt: Done.

katdal/sensordata.py
ludwigschwardt: Done.

katdal/sensordata.py
ludwigschwardt: Done.

katdal/sensordata.py
ludwigschwardt: Indeed... To preserve consistency, it needs wrapping too, as all SensorData-stage operations should deal with wrapped objects. If the filler value is e.g. ```[]``` it will get the wrong dtype too, so this now uses my freshly minted ```infer_dtype``` function instead.

Thanks for picking that up. Fixed.

katdal/concatdata.py
ludwigschwardt: Oops, that opened a whole kettle of fish...

katdal/concatdata.py
ludwigschwardt: True. I was confused by the high-level sensor name (the key in the SensorCache) and the low-level / underlying sensor name (SensorData.name). I wanted to refer to the former, but that is not available at this level. Fixed.

katdal/concatdata.py
ludwigschwardt: I will do that once the infer_dtypes kettle o' fish settles.

katdal/concatdata.py
ludwigschwardt: Done.

katdal/concatdata.py
ludwigschwardt: If sd is a ```CategoricalData``` or ```np.ndarray```, i.e. already extracted and not a raw ```SensorData``` object.

PR #87, opened 2017-06-13
Improve dump to event parsing (and introduce unit tests)

ludwigschwardt:
Rework the ```sensor_to_categorical``` function in response to the popularity of longer dump periods (e.g. 8 seconds) on MeerKAT. The existing function works OK if the dumps are shorter than the typical dish activities like tracks and slews. With 8-second dumps there is the  roblem that short slews (as found in e.g. interferometric pointing) may start and end within a single dump. This mini-slew is then filtered out by the dump -> event parser, resulting in a single track containing multiple pointings.

The new version of the function associates an entire dump with a greedy event, even if there is only a whiff of that event within the dump. Any non-greedy events at the end of such a dump is carried over to the next dump in this scenario. This logic is split out to its own function (```_single_event_per_dump```).

Other minor enhancements include the culling of sensor events outside the dump timestamp range (useful for big multi-observation telstates) and a reordering of steps so that sensor values are transformed first.

And ```katdal```'s very first unit test... providing 18% coverage!

This addresses JIRA ticket [SR-861](https://skaafrica.atlassian.net/browse/SR-861).

### DIFF
--- a/Jenkinsfile
+++ b/Jenkinsfile
@@ -1,0 +1,6 @@
+#!groovy
+
+@Library('katsdpjenkins') _
+katsdp.setDependencies(['ska-sa/katsdpdockerbase/master'])
+katsdp.standardBuild()
+katsdp.mail('schwardt@ska.ac.za')
--- a/Jenkinsfile
+++ b/Jenkinsfile
@@ -1,6 +1,6 @@
 #!groovy
 
 @Library('katsdpjenkins') _
-katsdp.setDependencies(['ska-sa/katsdpdockerbase/master'])
+katsdp.setDependencies(['ska-sa/katsdpdockerbase/master', 'ska-sa/katpoint/master'])
 katsdp.standardBuild()
-katsdp.mail('schwardt@ska.ac.za')
+katsdp.mail('ludwig@ska.ac.za')
--- a/katdal/categorical.py
+++ b/katdal/categorical.py
@@ -603,28 +603,27 @@ def _single_event_per_dump(events, greedy):
 
     Yields
     ------
-    event : non-negative int
+    event_index : non-negative int
         Index into `events` sequence of the next cleaned up event (does not
         yield the one-past-last-dump terminal event)
 
     """
-    # This will generate consecutive event indices with associated dump indices
-    dumps = enumerate(events)
     # The previous winning event is the dominant event in the previous dump
-    previous_dump, previous_winning_event = 0, 0
-    while True:
-        current_event, current_dump = next(dumps)
+    previous_winning_event = 0
+    previous_dump = 0
+    # This generates consecutive event indices with associated dump indices
+    for current_event, current_dump in enumerate(events):
         # At the start of a new dump, process the events of the previous dump
         if current_dump > previous_dump:
             # This previous event segment is assumed to straddle dump boundary
-            # ASSERT: current_event >= 1 since we are now at least at 2nd dump
+            assert current_event >= 1, "First sensor event not at dump 0"
             event_at_dump_start = current_event - 1
             # The normal victory condition is to be the final event in the dump
             if not greedy[previous_winning_event]:
                 previous_winning_event = event_at_dump_start
             winning_dump = events[previous_winning_event]
             # Only yield winning event in immediate past to avoid duplicates
-            if previous_dump <= winning_dump and winning_dump < current_dump:
+            if previous_dump <= winning_dump < current_dump:
                 yield previous_winning_event
             # If winning event was greedy and final event was non-greedy,
             # push final event to the start of next dump and yield if it is
@@ -674,7 +673,7 @@ def sensor_to_categorical(sensor_timestamps, sensor_values, dump_midtimes,
         Duration of each dump, in seconds
     transform : callable or None, optional
         Transform [unwrapped] sensor values before fixing initial value,
-        applying greed and discarding repeats
+        mapping dumps to events and discarding repeats
     initial_value : object or None, optional
         Sensor value [transformed, unwrapped] to use for dump = 0 up to first
         proper event (force first proper event to start at dump = 0 by default)
@@ -699,9 +698,8 @@ def sensor_to_categorical(sensor_timestamps, sensor_values, dump_midtimes,
     wrapped_values = len(sensor_values) and isinstance(sensor_values[0],
                                                        ComparableArrayWrapper)
     # Convert sensor event times to dump indices by picking the dump during
-    # which each sensor event occurred (more precisely: closest dump centroid).
-    # Last event is fixed at one-past-last-dump to indicate end of last segment.
-    events = np.r_[dump_endtimes.searchsorted(sensor_timestamps) - 1, num_dumps]
+    # which each sensor event occurred (more precisely: closest dump centroid)
+    events = dump_endtimes.searchsorted(sensor_timestamps) - 1
     # Get rid of excess events before the first dump and after the last dump
     # The dump index of prior events is -1 and of later events is `num_dumps`
     first_proper_event = events.searchsorted(-1, side='right')
@@ -712,7 +710,7 @@ def sensor_to_categorical(sensor_timestamps, sensor_values, dump_midtimes,
     one_past_last_event = events.searchsorted(num_dumps)
     within_dumps = slice(first_proper_event, one_past_last_event)
     sensor_values = sensor_values[within_dumps]
-    events = np.r_[events[within_dumps], num_dumps]
+    events = events[within_dumps]
     # Apply optional transform to sensor values
     if transform is not None:
         if wrapped_values:
@@ -732,14 +730,15 @@ def sensor_to_categorical(sensor_timestamps, sensor_values, dump_midtimes,
     # Clean up dump->event mapping, taking into account greedy values
     greedy_values = () if greedy_values is None else greedy_values
     greedy = [value in greedy_values for value in sensor_values]
-    cleaned_up = [event for event in _single_event_per_dump(events, greedy)]
+    cleaned_up = list(_single_event_per_dump(np.r_[events, num_dumps], greedy))
     sensor_values = sensor_values[cleaned_up]
-    events = np.r_[events[cleaned_up], num_dumps]
+    events = events[cleaned_up]
     # Discard sensor events that do not change the (transformed) sensor value
     # (i.e. that repeat the previous value)
     if not allow_repeats:
-        changes_value = [n for n in range(len(sensor_values)) if (n == 0) or
-                         (sensor_values[n] != sensor_values[n - 1])]
+        changes_value = [n for n in range(len(sensor_values)) if n == 0 or
+                         sensor_values[n] != sensor_values[n - 1]]
         sensor_values = sensor_values[changes_value]
-        events = np.r_[events[changes_value], num_dumps]
-    return CategoricalData(sensor_values, events)
+        events = events[changes_value]
+    # Last event is fixed at one-past-last-dump to indicate end of last segment
+    return CategoricalData(sensor_values, np.r_[events, num_dumps])
--- a/katdal/categorical.py
+++ b/katdal/categorical.py
@@ -580,8 +580,71 @@ def concatenate_categorical(split_data, **kwargs):
     return data
 
 
-def sensor_to_categorical(sensor_timestamps, sensor_values, dump_midtimes, dump_period,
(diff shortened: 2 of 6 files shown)


bmerry:
Since there's now a unit test, can I suggest adding a `Jenkinsfile` and a `test-requirements.txt`, so that Jenkins will do CI on the unit test? You can look at katsdptelstate for an example.

katdal/categorical.py
```diff
@@ -591,69 +654,92 @@ def sensor_to_categorical(sensor_timestamps, sensor_values, dump_midtimes, dump_
...
         Duration of each dump, in seconds
-    greedy_values : sequence or None, optional
-        List of (transformed) sensor values considered "greedy"
-    initial_value : object or None, optional
-        Sensor value to use for dump = 0 up to the first proper event
-        (the default is to force the first proper event to start at dump = 0)
     transform : callable or None, optional
-        Transform sensor values before discarding repeats and applying greed
+        Transform [unwrapped] sensor values before fixing initial value,
+        applying greed and discarding repeats
```
bmerry: greed -> greedy?

katdal/categorical.py
```diff
@@ -591,69 +654,92 @@ def sensor_to_categorical(sensor_timestamps, sensor_values, dump_midtimes, dump_
...
-    # Force first dump to have valid sensor value (use initial value or first proper value advanced to start)
-    if events[0] != 0 and initial_value is not None:
-        if wrapped_values:
-            initial_value = ComparableArrayWrapper(initial_value)
-        sensor_values, events = np.r_[[initial_value], sensor_values], np.r_[0, events]
-    events[0] = 0
+    # Convert sensor event times to dump indices by picking the dump during
+    # which each sensor event occurred (more precisely: closest dump centroid).
+    # Last event is fixed at one-past-last-dump to indicate end of last segment.
+    events = np.r_[dump_endtimes.searchsorted(sensor_timestamps) - 1, num_dumps]
```
bmerry: Appending num_dumps is I think both unnecessary (because it gets appended again after slicing out the interesting period) and harmful (because if there are past-the-end values, `events` is no longer sorted yet you're using `searchsorted` on it).

katdal/categorical.py
```diff
@@ -591,69 +654,92 @@ def sensor_to_categorical(sensor_timestamps, sensor_values, dump_midtimes, dump_
...
+    if events[0] != 0 and initial_value is not None:
+        if wrapped_values:
+            initial_value = ComparableArrayWrapper(initial_value)
+        sensor_values = np.r_[[initial_value], sensor_values]
+        events = np.r_[0, events]
+    events[0] = 0
+    # Clean up dump->event mapping, taking into account greedy values
+    greedy_values = () if greedy_values is None else greedy_values
+    greedy = [value in greedy_values for value in sensor_values]
+    cleaned_up = [event for event in _single_event_per_dump(events, greedy)]
```
bmerry: Could also write a `list(_single_event_per_dump(events, greedy))`. May just be a matter of taste.

katdal/categorical.py
```diff
@@ -591,69 +654,92 @@ def sensor_to_categorical(sensor_timestamps, sensor_values, dump_midtimes, dump_
...
+    # (i.e. that repeat the previous value)
     if not allow_repeats:
-        changes = [n for n in range(len(sensor_values)) if (n == 0) or (sensor_values[n] != sensor_values[n - 1])]
-        sensor_values, events = sensor_values[changes], np.r_[events[changes], len(dump_midtimes)]
-    # Extend segments with "greedy" values to include first dump of next segment where sensor value is changing
-    if greedy_values is not None:
-        greedy_to_nongreedy = [n for n in range(1, len(sensor_values))
-                               if sensor_values[n - 1] in greedy_values and sensor_values[n] not in greedy_values]
-        events[greedy_to_nongreedy] += 1
+        changes_value = [n for n in range(len(sensor_values)) if (n == 0) or
```
bmerry: Parentheses around `(n==0)` are redundant.

katdal/categorical.py
```diff
@@ -591,69 +654,92 @@ def sensor_to_categorical(sensor_timestamps, sensor_values, dump_midtimes, dump_
...
+                         (sensor_values[n] != sensor_values[n - 1])]
+        sensor_values = sensor_values[changes_value]
+        events = np.r_[events[changes_value], num_dumps]
```
bmerry: This function seems to be doing a lot of work to always re-add `num_dumps` to the end of `events`. Would the code get cleaner if that was deferred to the end of the function?

katdal/test/test_categorical.py
```diff
@@ -0,0 +1,20 @@
...
+
+
+def test_dump_to_event_parsing():
+    values = np.array([c for c in 'ABCDEFGH'])
+    events = np.array([0, 0, 1, 3, 3, 4, 4, 6, 8])
+    greedy = np.array([1, 0, 0, 1, 1, 0, 0, 0])
+    cleaned = [i for i in _single_event_per_dump(events, greedy)]
+    new_values = values[cleaned]
+    new_events = np.r_[events[cleaned], events[-1]]
+    print cleaned, new_values, new_events
```
bmerry: If you want to be Python3-ready, `from __future__ import print_function` and use `print` as a function.

katdal/test/test_categorical.py
```diff
@@ -0,0 +1,20 @@
...
+    assert_equal(cleaned, [0, 2, 4, 6, 7], 'Dump -> event parser failed')
+    assert_equal(''.join(new_values), 'ACEGH', 'Dump -> event parser failed')
+    events_equal = np.array_equal(new_events, [0, 1, 3, 5, 6, 8])
+    assert_true(events_equal, 'Dump -> event parser failed')
```
bmerry: There is `np.testing.assert_equal` for testing array equality (which will print both arrays if they differ).

ludwigschwardt:
Agreed about Jenkins - I've copied katsdptelstate's version for starters.

katdal/categorical.py
```diff
@@ -580,8 +580,71 @@ def concatenate_categorical(split_data, **kwargs):
...
+        The last event is one past the last dump (i.e. the total number of
+        dumps). Be aware that this parameter is mutated by the function.
+    greedy : sequence of bool, length *N*
+        Flags indicating whether the sensor value at a given event is "greedy"
+
+    Yields
+    ------
+    event : non-negative int
+        Index into `events` sequence of the next cleaned up event (does not
+        yield the one-past-last-dump terminal event)
```
ludwigschwardt: Done.

katdal/categorical.py
```diff
@@ -580,8 +580,71 @@ def concatenate_categorical(split_data, **kwargs):
...
+
+    """
+    # This will generate consecutive event indices with associated dump indices
+    dumps = enumerate(events)
+    # The previous winning event is the dominant event in the previous dump
+    previous_dump, previous_winning_event = 0, 0
```
ludwigschwardt: Point taken.

katdal/categorical.py
```diff
@@ -580,8 +580,71 @@ def concatenate_categorical(split_data, **kwargs):
...
+    while True:
+        current_event, current_dump = next(dumps)
```
ludwigschwardt: Indeed. This whole loop evolved from having an explicit break condition with a ```return``` to not having one in the last few minutes before the commit and I missed that.

katdal/categorical.py
```diff
@@ -580,8 +580,71 @@ def concatenate_categorical(split_data, **kwargs):
...
+        # At the start of a new dump, process the events of the previous dump
+        if current_dump > previous_dump:
+            # This previous event segment is assumed to straddle dump boundary
+            # ASSERT: current_event >= 1 since we are now at least at 2nd dump
```
ludwigschwardt: Thought I should have, too.

katdal/categorical.py
```diff
@@ -580,8 +580,71 @@ def concatenate_categorical(split_data, **kwargs):
...
+            event_at_dump_start = current_event - 1
+            # The normal victory condition is to be the final event in the dump
+            if not greedy[previous_winning_event]:
+                previous_winning_event = event_at_dump_start
+            winning_dump = events[previous_winning_event]
+            # Only yield winning event in immediate past to avoid duplicates
+            if previous_dump <= winning_dump and winning_dump < current_dump:
```
ludwigschwardt: Excellent, I was kind of hoping that but did not check the docs!

katdal/categorical.py
```diff
@@ -580,8 +580,71 @@ def concatenate_categorical(split_data, **kwargs):
...
+                yield previous_winning_event
+            # If winning event was greedy and final event was non-greedy,
+            # push final event to the start of next dump and yield if it is
+            # the only event in that dump (otherwise it has to fight it out...)
+            if event_at_dump_start != previous_winning_event:
+                # NB: This modifies `events`! It simplifies bookkeeping.
```
ludwigschwardt: Yes. The "event at dump start" will either be resolved immediately or affect at most the next iteration of `dumps`. There it will be handled specially, i.e. it will explicitly index `events` to retrieve its now modified dump index. I'm effectively modifying the past of the `dumps` iterator, not the present or future.

katdal/categorical.py
ludwigschwardt: I meant it in plain English: applying [noun] but maybe it is too cute.

katdal/categorical.py
ludwigschwardt: I accept the unnecessary bit (it was left in there as this function organically morphed into its current incarnation).

I'm not worried about the harmful bit as `dump_endtimes.searchsorted(sensor_timestamps) - 1` will always be less than or equal to `num_dumps` so `events` is guaranteed to be sorted.

katdal/categorical.py
ludwigschwardt: Nice, I did not think of that, very much preferred!

katdal/categorical.py
ludwigschwardt: Fixed. Same for the `!=` on the next line?

katdal/categorical.py
ludwigschwardt: I was actually thinking of also combining the values + events filtering into a single function as it happens thrice, only with different selections. But your suggestion seems better, done!

katdal/test/test_categorical.py
ludwigschwardt: Oops, that was a debug print...

bmerry:
Re the Jenkins build failing: I think it's because you also need a `requirements.txt` (corresponding to the `install_requires`).

Jenkinsfile
```diff
@@ -0,0 +1,6 @@
+#!groovy
+
+@Library('katsdpjenkins') _
+katsdp.setDependencies(['ska-sa/katsdpdockerbase/master'])
```
bmerry: Add 'ska-sa/katpoint/master' as well, so that this gets retested after katpoint is updated.

katdal/categorical.py
```diff
@@ -580,8 +580,70 @@ def concatenate_categorical(split_data, **kwargs):
...
+        if current_dump > previous_dump:
+            # This previous event segment is assumed to straddle dump boundary
+            assert current_event >= 1, "First sensor event not at dump 0"
+            event_at_dump_start = current_event - 1
+            # The normal victory condition is to be the final event in the dump
+            if not greedy[previous_winning_event]:
+                previous_winning_event = event_at_dump_start
+            winning_dump = events[previous_winning_event]
+            # Only yield winning event in immediate past to avoid duplicates
+            if previous_dump <= winning_dump < current_dump:
```
ludwigschwardt: Good question. The other `yield` further down has a need for the dump to be in the past, maybe I included it here too. The `event_at_dump_start` should be associated with `previous_dump` and `previous_winning_event` should not be ahead of it. So I think you are right that it is belt-and-braces.

katdal/test/test_categorical.py
```diff
@@ -0,0 +1,18 @@
+"""Tests for :py:mod:`katdal.categorical`."""
+
+import numpy as np
+from numpy.testing import assert_equal
```
ludwigschwardt: Actually, `np.testing.assert_array_equal` prints both arrays. I could not get any of the other variants of `assert_equal` to do that, so switch to that instead.

Jenkinsfile
ludwigschwardt: Done.

PR #90, opened 2017-07-19
Fix categorical sensor parser

ludwigschwardt:
The `_single_event_per_dump` filter is supposed to mutate the events array to simplify the process of pushing forward displaced events. Unfortunately the code in `sensor_to_categorical` passed it a copy of `events`, which defeats the whole purpose of the mutation.

Pass in the real `events`, slap on more warnings about the mutation and even add a unit test to verify the finicky sensor->categorical conversion.

### DIFF
--- a/katdal/test/test_categorical.py
+++ b/katdal/test/test_categorical.py
@@ -3,7 +3,7 @@
 import numpy as np
 from numpy.testing import assert_array_equal
 
-from katdal.categorical import _single_event_per_dump
+from katdal.categorical import _single_event_per_dump, sensor_to_categorical
 
 
 def test_dump_to_event_parsing():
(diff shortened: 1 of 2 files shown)


katdal/test/test_categorical.py
```diff
@@ -16,3 +16,19 @@ def test_dump_to_event_parsing():
...
+    timestamps = [-363.784, 2.467, 8.839, 8.867, 15.924, 48.925, 54.897, 88.982]
+    values = ['stop', 'slew', 'track', 'slew', 'track', 'slew', 'track', 'slew']
+    dump_period = 8.
+    dump_times = np.arange(4., 100., dump_period)
+    categ = sensor_to_categorical(timestamps, values, dump_times, dump_period,
+                                  greedy_values=('slew', 'stop'),
+                                  initial_value='slew')
+    assert_array_equal(categ.unique_values, ['slew', 'track'],
+                       'Sensor->categorical failed')
+    assert_array_equal(categ.events, [0, 2, 6, 7, 11, 12],
```
bmerry: It's probably just me, but is this correct? The slew at 48.925 starts during dump 5 (44-52), so greediness should make it take over dump 5, rather than only starting at dump 6.

ludwigschwardt: Those dump times are at the middle of the dump, so dump 5 is actually 40-48...

PR #91, opened 2017-07-27
Support standard UHF receivers

ludwigschwardt:
The standard UHF receivers have a digitiser sampling at 1088 MHz with an unflipped bandwidth of 544 MHz in the second Nyquist zone (544 - 1088 Mhz), resulting in a centre frequency of 816 MHz.

Delegate the old "fake" UHF receiver to the corner cases (only active if the bandwidth is 856 MHz, i.e. it uses an L-band digitiser). Move the calculation of bandwidth and channels earlier to facilitate this.

### DIFF
--- a/katdal/h5datav3.py
+++ b/katdal/h5datav3.py
@@ -450,13 +450,20 @@ class H5DataV3(DataSet):
         rx_table = {
             # Standard L-band receiver
             'l': dict(band='L', centre_freq=1284e6, sideband=1),
-            # Custom UHF receiver (real receiver + L-band digitiser, flipped spectrum)
-            'u': dict(band='UHF', centre_freq=428e6, sideband=-1),
+            # Standard UHF receiver
+            'u': dict(band='UHF', centre_freq=816e6, sideband=1),
             # Custom Ku receiver for RTS
             'x': dict(band='Ku', sideband=1),
         }
         spw_params = rx_table.get(band, dict(band='', sideband=1))
-        # Cater for receivers with mixers
+        # XXX Cater for future narrowband mode at some stage
+        num_chans = cbf_group.attrs['n_chans']
+        bandwidth = cbf_group.attrs['bandwidth']
+        # Work around a bc856M4k CBF bug active from 2016-04-28 to 2016-06-01 that got the bandwidth wrong
+        if bandwidth == 857152196.0:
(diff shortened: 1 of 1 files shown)


katdal/h5datav3.py
```diff
@@ -450,13 +450,20 @@ def register_telstate_sensor(name, obj):
...
+            'u': dict(band='UHF', centre_freq=816e6, sideband=1),
             # Custom Ku receiver for RTS
             'x': dict(band='Ku', sideband=1),
         }
         spw_params = rx_table.get(band, dict(band='', sideband=1))
-        # Cater for receivers with mixers
+        # XXX Cater for future narrowband mode at some stage
+        num_chans = cbf_group.attrs['n_chans']
+        bandwidth = cbf_group.attrs['bandwidth']
+        # Work around a bc856M4k CBF bug active from 2016-04-28 to 2016-06-01 that got the bandwidth wrong
```
spassmoor: We should not be adding for arounds for old bugs

katdal/h5datav3.py
```diff
@@ -468,13 +475,10 @@ def register_telstate_sensor(name, obj):
...
             if siggen_freq:
                 spw_params['centre_freq'] = 100. * siggen_freq + 1284e6
-        # XXX Cater for future narrowband mode, at some stage
-        num_chans = cbf_group.attrs['n_chans']
-        bandwidth = cbf_group.attrs['bandwidth']
-        # Work around a bc856M4k CBF bug active from 2016-04-28 to 2016-06-01 that got the bandwidth wrong
-        if bandwidth == 857152196.0:
-            logger.warning('Worked around CBF bandwidth bug (857.152 MHz -> 856.000 MHz)')
-            bandwidth = 856000000.0
+        # "Fake UHF": a real receiver + L-band digitiser and flipped spectrum
```
spassmoor: We should be reading the information from config at runtime

ludwigschwardt:
I feel your pain, but the file is in the archive and someone wants to open it with the right settings...

At least these workarounds are limited and documented.

ajoubertza:
We will be rolling out a new version of CAM next week Wednesday, which will have the new centre frequency related sensors per stream for MKAT (not RTS).  So hopefully we won't need too many more workarounds after that...

ludwigschwardt:
Awesome!

Then I'll default to those sensors (once we store them :-)) and the chances of hitting these workarounds will gently fade away...

PR #94, opened 2017-08-24
Load new-style noise diode sensors

ludwigschwardt:
The latest noise diode sensors have the band identifier as part of their name. For example, the old L-band `dig_noise_diode` is now `dig_l_band_noise_diode`. This caters for the different origins of the sensors as not all digitisers are served by the MeerKAT DMC.

Modify the main alias `nd_coupler` manually if a new-style noise diode sensor was found (else remain with the default 'dig_noise_diode'). This should have been a proper sensor alias, but that machinery can only change suffixes and we need to move from `TelescopeState` to `Antennas` as well.

Also ensure that all sensors ending in `noise_diode` are interpreted correctly. And allow aliases to be added to the `SensorCache` after construction for posterity.

This addresses JIRA ticket [MKAT-99](https://skaafrica.atlassian.net/browse/MKAT-99).

### DIFF
--- a/katdal/h5datav3.py
+++ b/katdal/h5datav3.py
@@ -45,11 +45,11 @@ SENSOR_PROPS = dict(DEFAULT_SENSOR_PROPS)
 SENSOR_PROPS.update({
     '*activity': {'greedy_values': ('slew', 'stop'), 'initial_value': 'slew',
                   'transform': lambda act: SIMPLIFY_STATE.get(act, 'stop')},
-    '*target': {'initial_value': '', 'transform': _robust_target},
     '*ap_indexer_position': {'initial_value': ''},
-    '*_serial_number': {'initial_value': 0},
-    '*dig_noise_diode': {'categorical': True, 'greedy_values': (True,),
-                         'initial_value': 0.0, 'transform': lambda x: x > 0.0},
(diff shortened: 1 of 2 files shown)


katdal/h5datav3.py
```diff
@@ -45,11 +45,11 @@
...
     '*activity': {'greedy_values': ('slew', 'stop'), 'initial_value': 'slew',
                   'transform': lambda act: SIMPLIFY_STATE.get(act, 'stop')},
-    '*target': {'initial_value': '', 'transform': _robust_target},
     '*ap_indexer_position': {'initial_value': ''},
-    '*_serial_number': {'initial_value': 0},
-    '*dig_noise_diode': {'categorical': True, 'greedy_values': (True,),
-                         'initial_value': 0.0, 'transform': lambda x: x > 0.0},
+    '*noise_diode': {'categorical': True, 'greedy_values': (True,),
+                     'initial_value': 0.0, 'transform': lambda x: x > 0.0},
+    '*serial_number': {'initial_value': 0},
```
spassmoor: You dropped an underscore here was it intended ?

ludwigschwardt: Yes. I also sorted them alphabetically as a clean-up. It just reads easier without changing the behaviour.

PR #96, opened 2017-09-05
Handle output L0 stream attributes

bmerry:
This is to handle the new attributes written by ingest
(ska-sa/katsdpingest#194) and filewriter (ska-sa/katsdpdata#71), which
allow for
- The file to contain either the continuum or spectral L0 stream
  (disambiguated by an HDF5 attribute on the /Data group);
- The file to contain only a subset of channels, and thus have separate
  n_chans, bandwidth, centre frequency to the CBF.

An area that will need some review is the precedence of the different
ways of determining centre frequency. Currently I've made it trust the
SDP-written value (e.g. sdp_l0_center_freq) over anything else, because
values determined from the band don't take account of channel subsetting
in ingest. However, that in turn means trusting the CAM centre
frequency, which is probably not correct for Ku band. I propose that the
right way to handle that is to fix CAM to be Ku band aware before trying
to update RTS with the latest ingest and filewriter.

### DIFF
--- a/katdal/h5datav3.py
+++ b/katdal/h5datav3.py
@@ -152,6 +152,9 @@ class H5DataV3(DataSet):
     ----------
     file : :class:`h5py.File` object
         Underlying HDF5 file, exposed via :mod:`h5py` interface
+    stream_name : str
+        Name of the L0 data stream, for finding corresponding telescope
+        state keys.
 
     Notes
     -----
@@ -174,6 +177,7 @@ class H5DataV3(DataSet):
 
         # Load main HDF5 groups
         data_group, tm_group = f['Data'], f['TelescopeModel']
+        self.stream_name = data_group.attrs.get('stream_name', 'sdp_l0')
         # Pick first group with appropriate class as CBF
         cbfs = [comp for comp in tm_group
                 if tm_group[comp].attrs.get('class') == 'CorrelatorBeamformer']
@@ -216,6 +220,7 @@ class H5DataV3(DataSet):
         self.dump_period = cbf_dump_period = cbf_group.attrs['int_time']
         if 'sdp' in tm_group:
             self.dump_period = tm_group['sdp'].attrs.get('l0_int_time', self.dump_period)
+        self.dump_period = self._get_telstate_stream_attr('int_time', self.dump_period)
         # Determine if timestamps are already aligned with middle of dumps
         try:
             ts_ref = data_group['timestamps'].attrs['timestamp_reference']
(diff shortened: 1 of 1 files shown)


ludwigschwardt:
When are we rolling this out to site?

katdal/h5datav3.py
```diff
@@ -482,17 +482,17 @@ def register_telstate_sensor(name, obj):
...
         stream_centre_freq = self._get_telstate_stream_attr('center_freq')
         if stream_centre_freq is not None:
             spw_params['centre_freq'] = stream_centre_freq
-        # Get channel width from original CBF parameters
+        # Get channel width from original CBF / SDP parameters
         spw_params['channel_width'] = bandwidth / num_chans
         # Continue with different channel count, but invalidate centre freq (keep channel width though)
         if num_chans != self._vis.shape[1]:
-            logger.warning('Number of channels received from ingest (%d) differs '
+            logger.warning('Number of channels received from CBF / ingest (%d) differs '
```
bmerry: Why this change? When does a v3 file ever contain data received from CBF?

katdal/h5datav3.py
```diff
@@ -579,6 +579,7 @@ def _get_telstate_attr(self, key, default=None, no_unpickle=()):
         try:
             value = self.file['TelescopeState'].attrs[key]
             if value in no_unpickle:
+                # Telstate attributes were serialised via str() until 2016-11-29
```
bmerry: Maybe merge this comment into the last sentence of the docstring.

katdal/h5datav3.py
ludwigschwardt: I was thinking of the source of the information - ingest just passed CBF spead attrs through in the past. Now it can also modify them.

bmerry:
> When are we rolling this out to site?

Do you mean this PR, or the code that produces these attributes? The former you could deploy ASAP, since it (should be) backwards compatible. I'll try to get the latter out this month, although it affects multiple images so will need some coordination.

ludwigschwardt:
The latter, yes.

katdal/h5datav3.py
ludwigschwardt: Improved.

katdal/h5datav3.py
ludwigschwardt: Done.

bmerry:
Latest updates look good. It's your package, so I'll let you merge when it's convenient for you.

PR #100, opened 2017-10-23
Put HDF5 data into S3-compatible store

ludwigschwardt:
This extends the `h5toceph.py` script to write visibility data to S3 (via botocore), intended for the Ceph RADOS gateway and also tested on Minio. In the process the script is renamed to `h5toobj.py` to be more generic.

Also use dask to chunk the dataset and parallelise the writes to the store for both raw RADOS and S3 API, in anticipation of wider use of dask within katdal.


ludwigschwardt:
Ready for review again...

scripts/h5toobj.py
```diff
@@ -0,0 +1,343 @@
...
+    vis_chunk_shape = list(vis.shape)
+    vis_chunk_shape[0] = 1
+    vis_chunk_shape[1] = freq_chunksize
+    vis_chunk_shape = tuple(vis_chunk_shape)
+    ts.add("obj_basename", args.basename)
+    ts.add("obj_vis_chunk_shape", vis_chunk_shape)
+    ts.add("obj_vis_chunk_size", obj_size)
+    ts.add("obj_count", obj_count)
+
+    use_rados = args.ceph_pool is not None
```
sratcliffe: Minor quibble - if the user specifies both ceph.pool and s3-url, they will only get ceph...

ludwigschwardt:
Here is a basic template of what I want to achieve (eventually all is wrapped up into `katdal.open`):

```Python
import katdal
from katdal.datasources import open_data_source
from katdal.visdatav4 import VisibilityDataV4

s = open_data_source('telstate+redis://localhost:6379/latest/correlator')
d = VisibilityDataV4(s)
print d
```

sratcliffe:
As the Welsh would say: 'Tidy'

ludwigschwardt:
The last important bit on the writing side is the metadata pertaining to object store endpoint, program block, stream name and whatnot. I probably need to sort that out in order to get to a usable reader, so I'll probably sneak it into this PR along the way.

scripts/h5toobj.py
ludwigschwardt: Yeah... The problem with putting both into telstate (which makes sense), is which one do you pick when you open telstate since it's all supposed to be "self-contained"... I could go for rados first and then fall back to s3, but is that always a good idea?

ludwigschwardt: (I want to avoid having to specify the explicit endpoint of the data in the DataSource, as we discussed).

ludwigschwardt: Or we could add a param / query part to the URL that picks one explicitly.

sratcliffe: I think a default fallback, with the ability to be explicit via the URL when needed should cover all cases.

ludwigschwardt: True. That's mostly on the reader side though. On the writer side it would be good to store all possible ways to access the data. But what about writing the data itself? If both s3 and rados point to the same underlying store only the latter is probably needed.

ludwigschwardt:
This should be the final version of the object writer (for now!). I've mostly rewritten it, so please have a quick scan.

scripts/h5toobj.py
```diff
@@ -0,0 +1,342 @@
...
+except ImportError:
+    botocore = None
+else:
+    import botocore.config
+    import botocore.session
+
+try:
+    from cStringIO import StringIO
+except ImportError:
+    from io import BytesIO as StringIO
```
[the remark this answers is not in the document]
ludwigschwardt: I agree. I've got no idea where I copy and pasted that from anymore. It looks like plain `io.BytesIO` will do the trick, so I'll try that instead.

scripts/h5toobj.py
```diff
@@ -0,0 +1,342 @@
...
+        use_s3 = args.s3_url is not None
+        use_rados = args.ceph_pool is not None
+        if use_rados and use_s3:
+            parser.error('Please specify either --ceph-pool or --s3-*')
+        if use_rados and not rados:
+            parser.error('Cannot use Ceph as rados Python library is not installed')
+        if use_s3 and not botocore:
+            parser.error('Cannot use S3 as botocore Python library is not installed')
+    if args.base_name is None:
+        args.base_name = args.file[0].split(".")[0]
```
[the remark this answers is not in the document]
ludwigschwardt: Yes, but why check for something no-one in their right mind will do? :-)

I was thinking more along the lines that this is a script aimed at power users, and --*-only is a power option, so power * power will prevent embarrassment.

scripts/h5toobj.py
```diff
@@ -0,0 +1,342 @@
...
+        yield key, slices, obj_name
+
+
+if __name__ == '__main__':
+    args = parse_args()
+    try:
+        f = katdal.open(args.file[0])
+        h5_file = f.file
+    except Exception as e:
+        logger.error("Failed to open specified HDF5 file. %s", e)
```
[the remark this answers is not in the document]
ludwigschwardt: I like.

PR #104, opened 2017-12-04
Fix opening of older v3 files

ludwigschwardt:
On pre-telstate v3 files, the `_get_telstate_attr()` helper function crashes when it looks for improved versions of the dump period attribute in the sensor cache. The problem is that the sensor cache
does not exist this early on in the metadata parsing... Fall back to the default obtained from the TelescopeModel on this exception.

### DIFF
--- a/katdal/h5datav3.py
+++ b/katdal/h5datav3.py
@@ -596,7 +596,7 @@ class H5DataV3(DataSet):
             # the most recent value.
             try:
                 return self.sensor['TelescopeState/' + key][-1]
-            except (KeyError, IndexError):
+            except (KeyError, IndexError, TypeError):
                 return default
 
     def _get_telstate_stream_attr(self, key, default=None, no_unpickle=()):
--- a/katdal/h5datav3.py
+++ b/katdal/h5datav3.py
@@ -596,7 +596,7 @@ class H5DataV3(DataSet):
             # the most recent value.
             try:
                 return self.sensor['TelescopeState/' + key][-1]
-            except (KeyError, IndexError, TypeError):
+            except (KeyError, IndexError):
                 return default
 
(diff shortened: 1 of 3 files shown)


katdal/h5datav3.py
```diff
@@ -596,7 +596,7 @@ def _get_telstate_attr(self, key, default=None, no_unpickle=()):
             # the most recent value.
             try:
                 return self.sensor['TelescopeState/' + key][-1]
-            except (KeyError, IndexError):
+            except (KeyError, IndexError, TypeError):
```
bmerry: Why would this give a `TypeError` rather than an `AttributeError`?

Either way, I think a comment would be useful to make the reader aware that the case of the sensor cache not existing yet needs to be handled. Otherwise someone in future might not see where the exception could come from and remove it again.

For completeness, an alternative would be initially set `self.sensor = {}` so that the sensor cache appears empty rather than missing. I'm not sure I like that though, since the type changes when the real cache is created.

ludwigschwardt: It's a `TypeError` because I set `self.sensor` to `None` at the start.

I like your `dict` suggestion. It is not that far-fetched since `SensorCache` is a derived class of `dict`... Alternatively I could cobble together an empty `SensorCache`.

bmerry:
> It is not that far-fetched since SensorCache is a derived class of dict... Alternatively I could cobble together an empty SensorCache.

I guess the user will never see that the type changes from `dict` to `SensorCache`, since it all happens during `__init__` - so it's probably not worth the effort to make an empty `SensorCache` if it takes more than a line or two of code.

ludwigschwardt:
Indeed, so I went the empty dict route.

PR #106, opened 2017-12-07
Require python-casacore 2.2.1 when importing

sjperkins:
python-casacore 2.2.1 is required for creating Measurement Sets. Raise an informative ImportError if the required version is not installed, rather than failing due to missing functionality.

### DIFF
--- a/katdal/ms_extra.py
+++ b/katdal/ms_extra.py
@@ -53,9 +53,11 @@ except:
         req_ver = parse_version("2.2.1")
 
         if not pyc_ver >= req_ver:
-            raise ImportError("python-casacore '%s' is required, "
-                              "but the current version is '%s'."
-                                              % (req_ver, pyc_ver))
+            raise ImportError("python-casacore %s is required, "
+                              "but the current version is %s. "
+                              "Note that python-casacore %s "
+                              "requires at least casacore 2.3.0."
+                                      % (req_ver, pyc_ver,req_ver))
 
 
 def std_scalar(comment, valueType='integer', option=0, **kwargs):
--- a/katdal/ms_extra.py
+++ b/katdal/ms_extra.py
@@ -46,14 +46,15 @@ except:
         casacore_binding = ''
     else:
         # Perform python-casacore version checks
-        import pkg_resources
-        pyc_dist = pkg_resources.get_distribution('python-casacore')
-        pyc_ver = pkg_resources.parse_version(pyc_dist.version)
-        req_ver = pkg_resources.parse_version("2.2.1")
+        from pkg_resources import parse_version
+        import casacore
+
+        pyc_ver = parse_version(casacore.__version__)
+        req_ver = parse_version("2.2.1")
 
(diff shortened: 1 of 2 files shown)


katdal/ms_extra.py
```diff
@@ -44,6 +44,18 @@
         casacore_binding = 'pyrap'
     except ImportError:
         casacore_binding = ''
+    else:
+        # Perform python-casacore version checks
```
bmerry: Wouldn't it be easier to check `casacore.__version__`?

katdal/ms_extra.py
```diff
@@ -44,6 +44,18 @@
...
+        import pkg_resources
+        pyc_dist = pkg_resources.get_distribution('python-casacore')
+        pyc_ver = pkg_resources.parse_version(pyc_dist.version)
+        req_ver = pkg_resources.parse_version("2.2.1")
+
+        if not pyc_ver >= req_ver:
+            raise ImportError("python-casacore '%s' is required, "
+                              "but the current version is '%s'"
+                                              % (req_ver, pyc_ver))
+
```
bmerry: Looks like too many blank lines here.

katdal/ms_extra.py
sjperkins: I poked around for it, but it doesn't exist.

katdal/ms_extra.py
sjperkins: Fixed in https://github.com/ska-sa/katdal/pull/106/commits/42ee40dad409ed0ce12f0a8e8e36e0cbc2912a9d

katdal/ms_extra.py
bmerry:
```python
>>> import casacore
>>> casacore.__version__
'2.2.1'
```
Works for me. Definitely also exists in 2.1.2, haven't checked older version. I'm always a bit distrustful of `get_distribution` because it reports what pip remembers installing, which isn't always the same as what's being picked up from the Python path.

sjperkins: Apologies, I should have read more carefully. I've updated the PR to use `casacore.__version__`.

katdal/ms_extra.py
```diff
@@ -46,15 +46,18 @@
...
 
         if not pyc_ver >= req_ver:
-            raise ImportError("python-casacore '%s' is required, "
-                              "but the current version is '%s'"
-                                              % (req_ver, pyc_ver))
+            raise ImportError("python-casacore %s is required, "
+                              "but the current version is %s. "
+                              "Note that python-casacore %s "
+                              "requires at least casacore 2.3.0."
+                                      % (req_ver, pyc_ver,req_ver))
```
bmerry: Space after the second comma.

sjperkins: Fixed

katdal/ms_extra.py
ludwigschwardt: Exactly as Bruce said. That's why we have a little package called `katversion` :-)

PR #110, opened 2018-01-12
Be more specific when comparing numpy dtypes

ludwigschwardt:
This removes a FutureWarning issued by numpy 1.14 and later. It is
no longer cool to be vague about the second argument to `np.issubdtype`.
It currently does some unexpected type conversions internally that
will go away eventually. If we are checking for a general floating-point
type, use np.floating instead of np.float (which is np.float64).
The same goes for np.integer, np.string_ and np.bool_ (which also have
sub-types). In those cases, also convert the default value to the
correct type (i.e. np.nan != np.float32(np.nan)).

### DIFF
--- a/katdal/sensordata.py
+++ b/katdal/sensordata.py
@@ -389,15 +389,15 @@ def dummy_sensor_data(name, value=None, dtype=np.float64, timestamp=0.0):
 
     """
     if value is None:
-        if np.issubdtype(dtype, np.float):
-            value = np.nan
-        elif np.issubdtype(dtype, np.int):
-            value = -1
-        elif np.issubdtype(dtype, np.str):
+        if np.issubdtype(dtype, np.floating):
+            value = np.dtype(dtype).type(np.nan)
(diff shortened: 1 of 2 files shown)


katdal/sensordata.py
```diff
@@ -389,15 +389,15 @@ def dummy_sensor_data(name, value=None, dtype=np.float64, timestamp=0.0):
...
     if value is None:
-        if np.issubdtype(dtype, np.float):
-            value = np.nan
-        elif np.issubdtype(dtype, np.int):
-            value = -1
-        elif np.issubdtype(dtype, np.str):
+        if np.issubdtype(dtype, np.floating):
+            value = np.dtype(dtype).type(np.nan)
+        elif np.issubdtype(dtype, np.integer):
+            value = np.dtype(dtype).type(-1)
```
bmerry: Are unsigned types used anywhere?

ludwigschwardt: Don't think so - the default value is pretty arbitrary in this case anyway.

PR #111, opened 2018-01-15
SR919: Prepare for dask

ludwigschwardt:
These smaller changes pave the way for dask:

- Improve RADOS error report if chunk is too big (unrelated)
- Rename `get` to `get_chunk` etc (there will soon be a `get_dask_array`)
- Switch to class-level fixtures in the tests (we need to add `test_put_and_get_dask_array`)
- Override botocore settings (unrelated, speeds up tests a lot)

### DIFF
--- a/katdal/chunkstore_rados.py
+++ b/katdal/chunkstore_rados.py
@@ -111,7 +111,8 @@ class RadosChunkStore(ChunkStore):
         if actual_bytes != expected_bytes:
             # Get the actual value via stat() to improve error reporting
             if actual_bytes > expected_bytes:
-                actual_bytes, _ = self.ioctx.stat(key)
+                with self._standard_errors(key):
+                    actual_bytes, _ = self.ioctx.stat(key)
             raise BadChunk('Chunk {!r}: dtype {} and shape {} implies an '
                            'object size of {} bytes, got {} bytes instead'
                            .format(key, dtype, shape, expected_bytes,
--- a/katdal/chunkstore_rados.py
+++ b/katdal/chunkstore_rados.py
@@ -103,14 +103,19 @@ class RadosChunkStore(ChunkStore):
         """See the docstring of :meth:`ChunkStore.get`."""
         dtype = np.dtype(dtype)
         key, shape = self.chunk_metadata(array_name, slices, dtype=dtype)
-        num_bytes = int(np.prod(shape)) * dtype.itemsize
+        expected_bytes = int(np.prod(shape)) * dtype.itemsize
         with self._standard_errors(key):
             # Try to read an extra byte to see if data is more than expected
-            data_str = self.ioctx.read(key, num_bytes + 1)
-        if len(data_str) != num_bytes:
+            data_str = self.ioctx.read(key, expected_bytes + 1)
+        actual_bytes = len(data_str)
+        if actual_bytes != expected_bytes:
+            # Get the actual value via stat() to improve error reporting
+            if actual_bytes > expected_bytes:
+                actual_bytes, _ = self.ioctx.stat(key)
             raise BadChunk('Chunk {!r}: dtype {} and shape {} implies an '
                            'object size of {} bytes, got {} bytes instead'
-                           .format(key, dtype, shape, num_bytes, len(data_str)))
+                           .format(key, dtype, shape, expected_bytes,
+                                   actual_bytes))
         return np.ndarray(shape, dtype, data_str)
 
     def put(self, array_name, slices, chunk):
--- a/katdal/chunkstore_rados.py
+++ b/katdal/chunkstore_rados.py
@@ -99,7 +99,7 @@ class RadosChunkStore(ChunkStore):
             raise StoreUnavailable(str(e))
         return cls(ioctx)
 
-    def get(self, array_name, slices, dtype):
+    def get_chunk(self, array_name, slices, dtype):
         """See the docstring of :meth:`ChunkStore.get`."""
         dtype = np.dtype(dtype)
         key, shape = self.chunk_metadata(array_name, slices, dtype=dtype)
@@ -118,12 +118,12 @@ class RadosChunkStore(ChunkStore):
                                    actual_bytes))
         return np.ndarray(shape, dtype, data_str)
 
-    def put(self, array_name, slices, chunk):
+    def put_chunk(self, array_name, slices, chunk):
         """See the docstring of :meth:`ChunkStore.put`."""
         key, shape = self.chunk_metadata(array_name, slices, chunk=chunk)
         data_str = chunk.tobytes()
         with self._standard_errors(key):
             self.ioctx.write_full(key, data_str)
 
-    get.__doc__ = ChunkStore.get.__doc__
-    put.__doc__ = ChunkStore.put.__doc__
+    get_chunk.__doc__ = ChunkStore.get_chunk.__doc__
+    put_chunk.__doc__ = ChunkStore.put_chunk.__doc__
--- a/katdal/chunkstore_s3.py
+++ b/katdal/chunkstore_s3.py
@@ -70,16 +70,22 @@ class S3ChunkStore(ChunkStore):
         self.client = client
 
     @classmethod
-    def from_url(cls, url):
+    def from_url(cls, url, **kwargs):
         """Construct S3 chunk store from endpoint URL.
 
         S3 authentication (i.e. the access + secret keys) is handled externally
-        via the botocore config file or environment variables.
+        via the botocore config file or environment variables. Extra keyword
+        arguments are interpreted as botocore config settings (see
+        :class:`botocore.config.Config`) or arguments to the client creation
+        method (see :meth:`botocore.session.Session.create_client`), in that
+        order, overriding the defaults.
 
         Parameters
         ----------
         url : string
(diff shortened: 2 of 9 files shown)


katdal/chunkstore_rados.py
```diff
@@ -99,26 +99,31 @@ def from_config(cls, config, pool, keyring=None, timeout=5.):
...
         with self._standard_errors(key):
             # Try to read an extra byte to see if data is more than expected
-            data_str = self.ioctx.read(key, num_bytes + 1)
-        if len(data_str) != num_bytes:
+            data_str = self.ioctx.read(key, expected_bytes + 1)
+        actual_bytes = len(data_str)
+        if actual_bytes != expected_bytes:
+            # Get the actual value via stat() to improve error reporting
+            if actual_bytes > expected_bytes:
+                actual_bytes, _ = self.ioctx.stat(key)
```
bmerry: Should this be done inside `_standard_errors`, so that if it fails (e.g. because the link to the store failed, or the object got deleted in between) you get a standard exception instead of a rados one?

katdal/chunkstore_rados.py
```diff
@@ -99,26 +99,31 @@ def from_config(cls, config, pool, keyring=None, timeout=5.):
...
             raise BadChunk('Chunk {!r}: dtype {} and shape {} implies an '
                            'object size of {} bytes, got {} bytes instead'
-                           .format(key, dtype, shape, num_bytes, len(data_str)))
+                           .format(key, dtype, shape, expected_bytes,
+                                   actual_bytes))
```
bmerry: Not really related to this PR, but I'm wondering if the `BadChunk` class should store extra fields such as the key, so that given an exception object you can interrogate it to programmatically determine where things went wrong.

katdal/chunkstore_s3.py
```diff
@@ -70,16 +70,22 @@ def __init__(self, client):
...
+    def from_url(cls, url, **kwargs):
         """Construct S3 chunk store from endpoint URL.
 
         S3 authentication (i.e. the access + secret keys) is handled externally
-        via the botocore config file or environment variables.
+        via the botocore config file or environment variables. Extra keyword
+        arguments are interpreted as botocore config settings (see
+        :class:`botocore.config.Config`) or arguments to the client creation
+        method (see :meth:`botocore.session.Session.create_client`), in that
+        order, overriding the defaults.
```
bmerry: I don't know what arguments are taken by each, but is there any risk that
(a) a current client argument is shadowed by a config argument and hence cannot be set through this interface?
(b) a future version of botocore would introduce a config argument that has the same name as a client argument, causing code that was expecting an argument to be passed to the client to be passed to the config instead?

katdal/test/test_chunkstore_rados.py
```diff
@@ -26,29 +26,22 @@
...
-class TestDudRadosChunkStore(object):
-    """Tests that don't need a RADOS connection, only a 'dud' store."""
-
-    def setup(self):
-        if not rados:
-            raise SkipTest('Rados not installed')
-
     def test_store_unavailable(self):
         # Pretend that rados is not installed
         katdal.chunkstore_rados.rados = None
```
bmerry: With this no longer being in a separate TestDudRadosChunkStore class, won't it cause things to break if the test runner decides to run this test before the others?

katdal/test/test_chunkstore_s3.py
```diff
@@ -73,28 +75,24 @@ def setup(self):
...
-
-    def setup(self):
-        if not botocore:
-            raise SkipTest('S3 botocore dependency not installed')
-
     def test_store_unavailable(self):
-        assert_raises(StoreUnavailable, S3ChunkStore.from_url, 'http://i.nvalid/')
+        # Drastically reduce the default botocore timeout of nearly 7 seconds
+        assert_raises(StoreUnavailable, S3ChunkStore.from_url,
+                      'http://i.nvalid/',
```
bmerry: Rather use a `.invalid` address. There is an RFC that guarantees that `.invalid` will never be a top-level domain, whereas this could theoretically be a valid address in future.

katdal/test/test_chunkstore.py
```diff
@@ -77,31 +77,31 @@ def __init__(self):
...
+        assert_raises(BadChunk, self.store.get_chunk, name, s, self.y.dtype)
         # Check basic put + get on 3-D float
         name = self.array_name('y')
         s = (slice(3, 7), slice(2, 5), slice(1, 2))
-        self.put_and_get('y', s)
+        self.put_and_get_chunk('y', s)
         # Stored object has more bytes than expected (and wrong dtype)
-        assert_raises(BadChunk, self.store.get, name, s, self.x.dtype)
+        assert_raises(BadChunk, self.store.get_chunk, name, s, self.x.dtype)
         # Try a chunk with zero size
```
bmerry: Now that you have class-level fixtures, it would be nice to split this test up into smaller pieces.

katdal/test/test_chunkstore_rados.py
```diff
@@ -26,29 +26,22 @@
...
 
-
-class TestDudRadosChunkStore(object):
-    """Tests that don't need a RADOS connection, only a 'dud' store."""
-
-    def setup(self):
-        if not rados:
-            raise SkipTest('Rados not installed')
-
     def test_store_unavailable(self):
```
bmerry: Not sure this is testing what it says on the tin. Shouldn't you test with a config file that points at a non-existent rados server?

katdal/chunkstore_rados.py
ludwigschwardt: Good idea, done.

katdal/chunkstore_rados.py
ludwigschwardt: I like that. It's a little more complicated since `BadChunk` can also happen while constructing the key, but still a useful idea. I've only recently started to store things on the exception, but it is already turning out useful. I think I need a grand error rework in the not-so-near future (once multiple chunk errors are also combined).

katdal/chunkstore_s3.py
ludwigschwardt: (a) The only common one is `region_name` which has the same meaning.
(b) The risk is low as the two is meant to complement each other. I'll worry about it if it happens. In that case one is still free to build a whole new custom botocore `Client` from scratch.

katdal/test/test_chunkstore_rados.py
ludwigschwardt: It restores the module three lines down, so should be idempotent.

katdal/test/test_chunkstore_rados.py
ludwigschwardt: "Store unavailable" is quite a broad concept which includes "no auth" and "bad config file". I'm testing the latter, which is sufficient for me to get my ~100% code coverage... They all go through the same top-level RADOS call, so no need to test each of them separately.

katdal/test/test_chunkstore.py
ludwigschwardt: Good idea.

katdal/test/test_chunkstore_s3.py
ludwigschwardt: I did not know that... Here's the [RFC](https://tools.ietf.org/html/rfc6761). Done.

katdal/test/test_chunkstore_rados.py
ludwigschwardt: Although I guess they could raise different exceptions, so maybe still work a check if simple enough.

katdal/test/test_chunkstore_rados.py
bmerry: Thanks, that hadn't shown up the diff.

katdal/test/test_chunkstore_rados.py
bmerry: Github wasn't showing me the whole function, so I'd missed the second part of the test. You can ignore my comment.

katdal/test/test_chunkstore_rados.py
bmerry: It should probably be done in a try/finally block though, so that if the test fails it still restores the previous state.

katdal/test/test_chunkstore_rados.py
ludwigschwardt: I'm now checking your suggestion too - it bags me two more lines of coverage and an extra exception too :-)

katdal/test/test_chunkstore_rados.py
ludwigschwardt: Done.

katdal/test/test_chunkstore_rados.py
```diff
@@ -43,10 +43,19 @@ def array_name(self, path):
...
+            assert_raises(ImportError, RadosChunkStore, None)
+        finally:
+            katdal.chunkstore_rados.rados = rados
+            katdal.chunkstore_rados._rados_import_error = None
+        # Missing config file
         assert_raises(StoreUnavailable, RadosChunkStore.from_config, 'x', 'y')
+        # Bad config (unknown config variable)
+        assert_raises(StoreUnavailable, RadosChunkStore.from_config,
+                      {'mon_hos': ''}, 'y')
+        # Config OK but host not found (use Test-Net IP from RFC5737)
```
bmerry: Nice, wasn't aware of these reserved addresses.

PR #112, opened 2018-01-17
Add dask support to ChunkStore

ludwigschwardt:
Add `get_dask_array` method that combines all the chunks associated with an array into a single dask Array and loads the associated data only on compute(). The corresponding `put_dask_array` method stores the chunks of a dask array in the store and returns a results array that also suppresses any exceptions (for inspection by the caller).

The supporting `generate_chunks` function provides a standard way to split an array into roughly equal-sized chunks. Dask is only a test requirement for now - wait until we have a fully integrated v4 loader to make it a main requirement.

### DIFF
--- a/katdal/chunkstore.py
+++ b/katdal/chunkstore.py
@@ -19,6 +19,7 @@
 from __future__ import division
 
 import contextlib
+import functools
 
 import numpy as np
 import dask
@@ -320,9 +321,8 @@ class ChunkStore(object):
             If requested chunk was not found in store
         """
         # Use dask utility function that forms the core of da.from_array
-        dask_graph = da.core.getem(array_name, chunks, self.get_chunk)
-        # Tweak parameter list of get_chunk() to include dtype
-        dask_graph = {k: v + (dtype,) for k, v in dask_graph.items()}
+        getter = functools.partial(self.get_chunk, dtype=dtype)
+        dask_graph = da.core.getem(array_name, chunks, getter)
         return da.Array(dask_graph, array_name, chunks, dtype)
 
     def put_dask_array(self, array_name, array):
@@ -335,20 +335,26 @@ class ChunkStore(object):
         array : :class:`dask.array.Array` object
             Dask input array
 
-        Raises
-        ------
-        :exc:`chunkstore.StoreUnavailable`
-            If interaction with chunk store failed (offline, bad auth, bad config)
-        :exc:`chunkstore.ChunkNotFound`
-            If `array_name` is incompatible with store
+        Returns
+        -------
+        success : :class:`dask.array.Array` object
+            Dask array indicating success of the transfer of each chunk
+            (None indicates success, otherwise there is an exception object)
         """
         dask_graph = dask.sharedict.ShareDict()
         dask_graph.update(array.dask)
+        # Give better names to these two very similar variables
+        in_name = array.name
+        out_name = array_name
+        # Disambiguate out_name to avoid clashes in the dask graph
+        if out_name == in_name:
+            out_name = 'store-' + out_name
         # Construct output graph on same chunks as input, but with new name
-        graph = da.core.getem(array_name, array.chunks, self.put_chunk_noraise)
+        graph = da.core.getem(array_name, array.chunks, self.put_chunk_noraise,
+                              out_name=out_name)
         # Set chunk parameter of put_chunk() to corresponding key in input array
-        graph = {k: v + ((array.name,) + k[1:],) for k, v in graph.items()}
+        graph = {k: v + ((in_name,) + k[1:],) for k, v in graph.items()}
         dask_graph.update(graph)
         # The success array has one element per chunk in the input array
         out_chunks = tuple(len(c) * (1,) for c in array.chunks)
-        return da.Array(dask_graph, array_name, out_chunks, np.object)
+        return da.Array(dask_graph, out_name, out_chunks, np.object)
--- a/katdal/chunkstore.py
+++ b/katdal/chunkstore.py
@@ -24,6 +24,7 @@ import functools
 import numpy as np
 import dask
 import dask.array as da
+import toolz
 
 
 class ChunkStoreError(Exception):
@@ -42,9 +43,9 @@ class BadChunk(ValueError, ChunkStoreError):
     """The chunk is malformed, e.g. bad dtype or slices, wrong buffer size."""
 
 
-def _ceil_power_of_two(x):
-    """The smallest power of two greater than or equal to `x`."""
-    return 2 ** int(np.ceil(np.log2(x)))
+def _floor_power_of_two(x):
+    """The largest power of two smaller than or equal to `x`."""
+    return 2 ** int(np.floor(np.log2(x)))
 
 
 def generate_chunks(shape, dtype, max_chunk_size, dims_to_split=None,
@@ -62,7 +63,8 @@ def generate_chunks(shape, dtype, max_chunk_size, dims_to_split=None,
     dims_to_split : sequence of int, optional
         Indices of dimensions that may be split into chunks (default all dims)
     power_of_two : bool, optional
-        True if chunk size should be a power of two on dimensions that allow it
+        True if chunk size should be rounded down to a power of two
+        (the last chunk size along each dimension will potentially be smaller)
 
     Returns
     -------
@@ -72,8 +74,11 @@ def generate_chunks(shape, dtype, max_chunk_size, dims_to_split=None,
     if dims_to_split is None:
         dims_to_split = range(len(shape))
     dataset_size = np.prod(shape) * np.dtype(dtype).itemsize
-    num_chunks = np.ceil(dataset_size / max_chunk_size)
+    # The ideal number of chunks to achieve requested chunk size (can be float)
+    num_chunks = dataset_size / max_chunk_size
+    # Start with the whole array as a single big chunk
     chunks = [(s,) for s in shape]
+    # Split the array greedily along each dimension, in order of `dims_to_split`
     for dim in dims_to_split:
         if dim >= len(shape):
             continue
@@ -81,14 +86,17 @@ def generate_chunks(shape, dtype, max_chunk_size, dims_to_split=None,
             # Split the dimension into the maximum number of chunks
             chunk_sizes = (1,) * shape[dim]
         else:
-            if power_of_two and _ceil_power_of_two(shape[dim]) == shape[dim]:
-                chunks_per_dim = _ceil_power_of_two(num_chunks)
-            else:
-                chunks_per_dim = np.ceil(num_chunks)
             items = np.arange(shape[dim])
-            chunk_indices = np.array_split(items, chunks_per_dim)
+            if power_of_two:
+                # Chunk sizes will be (2^P, 2^P, 2^P, ..., 1 <= M <= 2^P)
+                chunksize_per_dim = _floor_power_of_two(shape[dim] / num_chunks)
+                chunk_indices = toolz.partition_all(chunksize_per_dim, items)
+            else:
+                # Chunk sizes generally will be (N, N, N, ..., N-1, N-1)
+                chunk_indices = np.array_split(items, np.ceil(num_chunks))
             chunk_sizes = tuple([len(chunk) for chunk in chunk_indices])
         chunks[dim] = chunk_sizes
+        # Update number of remaining chunks to realise
         num_chunks /= len(chunk_sizes)
     return tuple(chunks)
 
@@ -346,9 +354,9 @@ class ChunkStore(object):
         # Give better names to these two very similar variables
         in_name = array.name
         out_name = array_name
-        # Disambiguate out_name to avoid clashes in the dask graph
-        if out_name == in_name:
-            out_name = 'store-' + out_name
+        # Make out_name reasonably unique to avoid clashes and caches
+        token = da.core.tokenize(object())[:8]
+        out_name = 'store-{}-{}'.format(out_name, token)
         # Construct output graph on same chunks as input, but with new name
         graph = da.core.getem(array_name, array.chunks, self.put_chunk_noraise,
                               out_name=out_name)
--- a/katdal/chunkstore.py
+++ b/katdal/chunkstore.py
@@ -16,9 +16,13 @@
 
 """Base class for accessing a store of chunks (i.e. N-dimensional arrays)."""
 
+from __future__ import division
+
 import contextlib
 
 import numpy as np
+import dask
+import dask.array as da
 
 
 class StoreUnavailable(OSError):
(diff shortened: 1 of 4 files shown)


katdal/chunkstore.py
```diff
@@ -33,6 +37,43 @@ class BadChunk(ValueError):
     """The chunk is malformed, e.g. bad dtype or slices, wrong buffer size."""
 
 
+def generate_chunks(shape, dtype, max_chunk_size, dims_to_split=(0, 1)):
```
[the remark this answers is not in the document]
ludwigschwardt: Thought about that, yes.

katdal/chunkstore.py
```diff
@@ -126,6 +167,16 @@ def put_chunk(self, array_name, slices, chunk):
         """
         raise NotImplementedError
 
+    def put_chunk_noraise(self, array_name, slices, chunk):
+        """Put chunk into store but return any exceptions instead of raising."""
+        try:
+            self.put_chunk(array_name, slices, chunk)
+            # Return result as ndarray with same number of (singleton)
+            # dimensions as chunk to enable assembly into a dask array
+            return np.array(None).reshape(len(slices) * (1,))
```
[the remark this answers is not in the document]
ludwigschwardt: It's debatable but I'll do an internal debate.

katdal/chunkstore.py
```diff
@@ -126,6 +167,16 @@ def put_chunk(self, array_name, slices, chunk):
...
+        except Exception as err:
```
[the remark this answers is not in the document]
ludwigschwardt: I like that.

katdal/chunkstore.py
```diff
@@ -224,3 +275,61 @@ def _standard_errors(self, chunk_name=None):
...
+        Raises
+        ------
+        :exc:`chunkstore.StoreUnavailable`
+            If interaction with chunk store failed (offline, bad auth, bad config)
+        :exc:`chunkstore.ChunkNotFound`
+            If requested chunk was not found in store
+        """
+        # Use dask utility function that forms the core of da.from_array
+        dask_graph = da.core.getem(array_name, chunks, self.get_chunk)
+        # Tweak parameter list of get_chunk() to include dtype
```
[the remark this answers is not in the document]
ludwigschwardt: Ehm... Yes :-) Forgot about that along the way. I might use that trick even more in a future version of this.

katdal/chunkstore.py
```diff
@@ -224,3 +275,61 @@ def _standard_errors(self, chunk_name=None):
...
+        dask_graph = {k: v + (dtype,) for k, v in dask_graph.items()}
+        return da.Array(dask_graph, array_name, chunks, dtype)
+
+    def put_dask_array(self, array_name, array):
+        """Put dask array into the store.
```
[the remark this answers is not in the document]
ludwigschwardt: Yes, that's intended. I would actually like to add an offset the other way (an *output* offset) to simplify vis_writer.

katdal/chunkstore.py
```diff
@@ -224,3 +275,61 @@ def _standard_errors(self, chunk_name=None):
...
+            Identifier of array in chunk store
+        array : :class:`dask.array.Array` object
+            Dask input array
+
+        Raises
+        ------
+        :exc:`chunkstore.StoreUnavailable`
+            If interaction with chunk store failed (offline, bad auth, bad config)
+        :exc:`chunkstore.ChunkNotFound`
+            If `array_name` is incompatible with store
```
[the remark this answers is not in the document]
ludwigschwardt: Copy-n-paste - ignore.

katdal/test/test_chunkstore.py
```diff
@@ -18,10 +18,28 @@
...
+    dtype = np.complex64
+    chunks = generate_chunks(shape, dtype, 3e6)
+    assert_equal(chunks, (10 * (1,), 4 * (2048,), (144,)))
+    chunks = generate_chunks(shape, dtype, 3e6, (0, 10))
+    assert_equal(chunks, (10 * (1,), (8192,), (144,)))
+    chunks = generate_chunks(shape, dtype, 1e6)
+    assert_equal(chunks, (10 * (1,), 2 * (820,) + 8 * (819,), (144,)))
+    chunks = generate_chunks(shape, dtype, 1e6, ())
+    assert_equal(chunks, ((10,), (8192,), (144,)))
+    chunks = generate_chunks(shape, dtype, 94371840)
```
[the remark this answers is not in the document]
ludwigschwardt: I should expand it, yes (it's the expected size in bytes).

katdal/test/test_chunkstore.py
```diff
@@ -73,6 +94,7 @@ def __init__(self):
         self.x = np.ones(10, dtype=np.bool)
         self.y = np.arange(96.).reshape(8, 6, 2)
         self.z = np.array(2.)
+        self.dask_x = np.arange(960.).reshape(8, 60, 2)
```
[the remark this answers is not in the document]
ludwigschwardt: Point taken.

katdal/test/test_chunkstore.py
```diff
@@ -85,6 +107,20 @@ def put_and_get_chunk(self, var_name, slices):
...
+        divisions_per_dim = [len(c) for c in chunks]
+        dask_array = da.from_array(array, chunks)
+        push = self.store.put_dask_array(array_name, dask_array)
+        results = push.compute()
+        assert_array_equal(results, np.tile(None, divisions_per_dim))
+        pull = self.store.get_dask_array(array_name, chunks, array.dtype)
+        array_retrieved = pull.compute()
+        assert_array_equal(array_retrieved, array,
+                           "Error storing {} / {}".format(var_name, chunks))
+
```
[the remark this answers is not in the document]
ludwigschwardt: Will do that separately.

requirements.txt
```diff
@@ -1,5 +1,6 @@
 botocore==1.8.9
 Cython
+dask==0.16.1
```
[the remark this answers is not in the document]
ludwigschwardt: Agreed.

katdal/chunkstore.py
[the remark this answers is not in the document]
ludwigschwardt: Done. Also added the power-of-two feature we discussed. I know it's somewhat non-generic again, but hey... At least it is off by default. I would like the chunk generation to be stable between various users of the chunk store, so I prefer that they call the same function in the core module.

katdal/chunkstore.py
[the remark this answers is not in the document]
ludwigschwardt: Gone your route.

katdal/chunkstore.py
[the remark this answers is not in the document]
ludwigschwardt: Done. Introduced a `ChunkStoreError` base class for this.

katdal/test/test_chunkstore.py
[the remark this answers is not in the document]
ludwigschwardt: Done.

katdal/chunkstore.py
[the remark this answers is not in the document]
ludwigschwardt: Done. The original version is a little bit more readable if one ever inspects the graph though (and one fewer calling layer)...

katdal/chunkstore.py
```diff
@@ -19,25 +19,36 @@
...
 
-class BadChunk(ValueError):
+class BadChunk(ValueError, ChunkStoreError):
     """The chunk is malformed, e.g. bad dtype or slices, wrong buffer size."""
 
 
-def generate_chunks(shape, dtype, max_chunk_size, dims_to_split=(0, 1)):
+def _ceil_power_of_two(x):
+    """The smallest power of two greater than or equal to `x`."""
+    return 2 ** int(np.ceil(np.log2(x)))
```
bmerry: This breaks down above about 2<sup>49</sup> due to rounding errors and for `x=0` since its log is -inf, but I assume neither should be an issue in practice.

katdal/chunkstore.py
```diff
@@ -66,8 +81,12 @@ def generate_chunks(shape, dtype, max_chunk_size, dims_to_split=(0, 1)):
             # Split the dimension into the maximum number of chunks
             chunk_sizes = (1,) * shape[dim]
         else:
+            if power_of_two and _ceil_power_of_two(shape[dim]) == shape[dim]:
+                chunks_per_dim = _ceil_power_of_two(num_chunks)
+            else:
+                chunks_per_dim = np.ceil(num_chunks)
```
bmerry: Any particular reason to turn off the power-of-two feature if the shape isn't a power-of-two? For example, one might discard the first and last 1024 of 32768 channels, but still prefer to have a chunk size of 128 (which will divide in evenly) than 125.

katdal/chunkstore.py
```diff
@@ -316,20 +335,26 @@ def put_dask_array(self, array_name, array):
...
-        ------
-        :exc:`chunkstore.StoreUnavailable`
-            If interaction with chunk store failed (offline, bad auth, bad config)
-        :exc:`chunkstore.ChunkNotFound`
-            If `array_name` is incompatible with store
+        Returns
+        -------
+        success : :class:`dask.array.Array` object
+            Dask array indicating success of the transfer of each chunk
+            (None indicates success, otherwise there is an exception object)
```
bmerry: Okay for now, but I think once you've written the vis-writer code that goes on top of this, see if the logic it uses to aggregate the results into a single pass/error can be pushed down into the chunk store.

katdal/chunkstore.py
```diff
@@ -316,20 +335,26 @@ def put_dask_array(self, array_name, array):
...
         """
         dask_graph = dask.sharedict.ShareDict()
         dask_graph.update(array.dask)
+        # Give better names to these two very similar variables
+        in_name = array.name
+        out_name = array_name
+        # Disambiguate out_name to avoid clashes in the dask graph
+        if out_name == in_name:
+            out_name = 'store-' + out_name
```
bmerry: There is no guarantee that this is globally unique. For example, if you write an array into two different chunk stores with the same array name, these will clash. Writing to a chunk store isn't really a "pure" function (just as one example, it may return an exception the first time, and succeed the second time), so I'd suggest using 'store-' + array_name + random string (I forget how to get one, but I know dask has the capability).

katdal/chunkstore.py
ludwigschwardt: Indeed.

katdal/chunkstore.py
ludwigschwardt: That's the plan, yes. Ideally I would like to handle get_dask_array's errors in a similar manner, which might imply some rework anyway.

katdal/chunkstore.py
ludwigschwardt: Of course. I just followed the lead of the dask gurus in their equivalent `da.core.insert_to_ooc` function.

I'm also considering a unique chunk store ID for this purpose, to make debugging easier. The `da.core.tokenize` function can give you a deterministic hash of the unique parts of the store.

katdal/chunkstore.py
bmerry: You've already written the code. Just remove the `if` statement and the `else` branch. There may be cases where you have only mostly power-of-two chunks plus a small piece at the end, but is that the end of the world?

ludwigschwardt: This is just a bit trickier than anticipated... For example, if we have 32768 channels for starters and we lob off 1024 at each end we get 30720 = 15 * 2048. So it is possible to get a nice power-of-2 compliant split if chunk_size = 2048. But what if the desired chunksize is 8192? I've now implemented this as 30720 = 8192 + 8192 + 8192 + 6144, giving more importance to the target object size than the power-of-two requirement (better to get most chunks large enough, I think).

I've thought of moving the power-to-two business to vis_writer but I prefer centralised logic to get the same splits everywhere.

katdal/chunkstore.py
ludwigschwardt: Made this mostly unique now with a random token (1 in 2 ** 32 is good enough for me).

bmerry: Dask seems to use `str(uuid.uuid1())`, which should give better than 1 in 2<sup>32</sup>.

ludwigschwardt: Yes, but those long strings hurt my eyes :-)

katdal/chunkstore.py
```diff
@@ -346,9 +354,9 @@ def put_dask_array(self, array_name, array):
         # Give better names to these two very similar variables
         in_name = array.name
         out_name = array_name
-        # Disambiguate out_name to avoid clashes in the dask graph
-        if out_name == in_name:
-            out_name = 'store-' + out_name
+        # Make out_name reasonably unique to avoid clashes and caches
+        token = da.core.tokenize(object())[:8]
```
bmerry: Looking at what `tokenize` does on `object`, it tries a bunch of things before falling back `uuid.uuid4().hex`; which then gets run through MD5. You might as well just use `uuid.uuid4().hex` and save the trip through MD5 (and the construction of a temporary object).

Also don't truncate it.

katdal/test/test_chunkstore.py
```diff
@@ -18,10 +18,50 @@
...
-from katdal.chunkstore import (ChunkStore, StoreUnavailable,
-                               ChunkNotFound, BadChunk)
+from nose.tools import assert_raises, assert_equal, assert_is_instance
+import dask.array as da
+
+from katdal.chunkstore import (ChunkStore, generate_chunks,
+                               StoreUnavailable, ChunkNotFound, BadChunk)
+
+
+def test_generate_chunks():
```
bmerry: This seems to be testing quite a lot of different things - maybe put it in a test class with multiple tests (at the very least, power-of-two could be a separate test).

katdal/chunkstore.py
ludwigschwardt: I was trying to avoid the extra line of `import uuid` :-)

The truncation is on purpose to make the key more readable while still random enough. On the other hand, inspecting the graphs by eye is hopefully not a regular action, so............ OK.

katdal/test/test_chunkstore.py
ludwigschwardt: I would say it is testing one thing - `generate_chunks` - but I hear ya. I had proto-thoughts about splitting it up. My threshold is just higher :-)

ludwigschwardt: Done.

katdal/chunkstore.py
ludwigschwardt: Done.

ludwigschwardt:
Added offset support to both `put_dask_array` and `get_dask_array` (the symmetry helped the tests).

I won't be plumbing this into vis_writer just yet - I first want to focus on the reader.

katdal/test/test_chunkstore.py
```diff
@@ -128,26 +128,37 @@ def array_name(self, name):
...
+            array = array[offset_slice]
         divisions_per_dim = [len(c) for c in chunks]
         dask_array = da.from_array(array, chunks)
-        push = self.store.put_dask_array(array_name, dask_array)
+        push = self.store.put_dask_array(array_name, dask_array, offset)
         results = push.compute()
         assert_array_equal(results, np.tile(None, divisions_per_dim))
-        pull = self.store.get_dask_array(array_name, chunks, array.dtype)
+        pull = self.store.get_dask_array(array_name, chunks, array.dtype,
+                                         offset)
```
bmerry: This test is a bit weak, because it only tests passing the same offset to put and then fetch the data. For example, if `_add_offset_to_slices` was replaced with an identity function, I think the test would still pass. It may be best to test e.g. writing an array in several pieces, then pulling it all back in one go.

ludwigschwardt: Done, thanks, it's much better now.

PR #113, opened 2018-01-20
fix typo in sensor name for the revciever serial number

spassmoor:
. The sensor name now includes the band

The h5 file has the sensors :
In [13]: print h5.sensor['TelescopeState/m005_ap_indexer_position']
['l' 'l' 'l' ..., 'l' 'l' 'l']
In [14]: print h5.sensor['TelescopeState/m005_rsc_rxl_serial_number']
[4029 4029 4029 ..., 4029 4029 4029]

But the code wants : 
            # Try sanitised version of RX serial number first
            rx_sensor = 'TelescopeState/%s_rx_serial_number' % (ant,)
            rx_serial = self.sensor[rx_sensor][0] if rx_sensor in self.sensor else 0

Should it rather be :
rx_sensor = 'TelescopeState/%s_rsc_rx%s_serial_number' % (ant, band)

### DIFF
--- a/katdal/h5datav3.py
+++ b/katdal/h5datav3.py
@@ -451,7 +451,7 @@ class H5DataV3(DataSet):
         # Populate antenna -> receiver mapping and figure out noise diode
         for ant in cam_ants:
             # Try sanitised version of RX serial number first
-            rx_sensor = 'TelescopeState/%s_rx_serial_number' % (ant,)
+            rx_sensor = 'TelescopeState/%s_rsc_rx%s_serial_number' % (ant, band)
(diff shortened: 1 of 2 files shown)


spassmoor:
I have made the same change to visdatav4.py   but I don't have a v4 file to test on , but the logic of the sensor name is the same. Although the edited code may be obsolete .

ludwigschwardt:
No, no, no :-)

The latest cam2telstate renames the serial number to be without the band label. If this is not found in the file, it falls back to the original one. Read the comments around this bit of code.

If it doesn't work, let's look at why instead.

Issue #114, opened 2018-01-22
katdal not finding TelescopeState receiver serial number sensor due to ignoring the band 


spassmoor:
For example in the file produced on the AR1 correlator :
```
/var/kat/archive3/data/RTS/telescope_products/2018/01/18/1516315038.h5
```
The h5 file has the sensors :
```Python
In [13]: print h5.sensor['TelescopeState/m005_ap_indexer_position']
Out[13]: ['l' 'l' 'l' ..., 'l' 'l' 'l']
In [14]: print h5.sensor['TelescopeState/m005_rsc_rxl_serial_number']
Out[14]: [4029 4029 4029 ..., 4029 4029 4029]
```
But the code does not examine the band :
```Python
# Try sanitised version of RX serial number first
rx_sensor = 'TelescopeState/%s_rx_serial_number' % (ant,)
rx_serial = self.sensor[rx_sensor][0] if rx_sensor in self.sensor else 0
```

ludwigschwardt:
Argh, what a mess - blame it on RTS :-P

The real digitiser sensor changed its name to `dig_l_band_noise_diode` and there was a PR ([katsdpingest #189](https://github.com/ska-sa/katsdpingest/pull/189)) to deal with this in August 2017. As part of this, I also took the liberty to rename `dig_l_band_noise_diode` to `dig_noise_diode` in cam2telstate since only one band will be active at a time anyway. This change suggested the same treatment for `rsc_rxl_serial_number` to `rx_serial_number`.

Bruce then convinced me not to rename sensors unnecessarily, and the rename was dropped from the PR before it was merged (to AR1's master) on 8 August 2017. Unfortunately the news never got to RTS, which received the sensor renaming version on 2 August 2017 already. In fact, it was used to test the code as it had to receive the changes before AR1.

So the end result is that RTS renames its noise diode and serial number sensors since the site update on 2017-11-15. Since PR #102 (approved by one Passmoor :-)) katdal supports the RTS version, with a fallback to the full sensor name in the old `TelescopeModel`, which inadvertently supported AR1 too. However, the recent site update on 2018-01-16 stripped those sensors out of `TelescopeModel`, leaving ARx in the lurch.

The solution would be to look for the sensor names in 3 places in v3, and 2 places in v4 (bye-bye RTS, good riddance!). I'll whip something up.

PR #115, opened 2018-01-22
Look for unrenamed RX serial number sensor first

ludwigschwardt:
Since 2017-11-15 RTS renamed the `rsc_rxl_serial_number` sensor to
`rx_serial_number` (i.e. dropping the band label). AR1 / MKAT would
have done this too, but stopped short at the last minute (see
katsdpingest PR #189). Since 2018-01-16 the MKAT files no longer have
the corresponding TelescopeModel sensor to fall back on. To fix this,
first look for the original sensor name in TelescopeState.

The `dig_l_band_noise_diode` sensor (renamed as `dig_noise_diode`)
luckily don't suffer from the same issue, as there the renamed sensor
matches the old name and the full name is already looked up first in
TelescopeState.

This closes issue #114.

### DIFF
--- a/katdal/h5datav3.py
+++ b/katdal/h5datav3.py
@@ -450,12 +450,18 @@ class H5DataV3(DataSet):
                            'please provide it via band parameter')
         # Populate antenna -> receiver mapping and figure out noise diode
         for ant in cam_ants:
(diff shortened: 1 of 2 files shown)


ludwigschwardt:
I feel the same way...

PR #117, opened 2018-02-01
Check presence of chunks in store

ludwigschwardt:
This does a basic check to see if a given chunk (or array) is present in the store, with the appropriate size (and dtype if supported), using the new `has_chunk` and `has_dask_array` methods. The idea is that it should be (much) faster than the default `get_chunk`, otherwise there is very little point...

I'm really adding this as a debugging tool, so it is not very optimised. Especially multi-chunk checking on e.g. S3 can still be sped up a lot.

### DIFF
--- a/katdal/chunkstore.py
+++ b/katdal/chunkstore.py
@@ -218,6 +218,37 @@ class ChunkStore(object):
         except ChunkStoreError as err:
             return np.full(singleton_shape, err, object)
 
+    def has_chunk(self, array_name, slices, dtype):
+        """Check if chunk is in the store.
+
+        Parameters
+        ----------
+        array_name : string
+            Identifier of parent array `x` of chunk
+        slices : sequence of unit-stride slice objects
+            Identifier of individual chunk, to be extracted as `x[slices]`
+        dtype : :class:`numpy.dtype` object or equivalent
+            Data type of array `x`
+
+        Returns
+        -------
+        success : bool
+            True if chunk was found in the store, with appropriate size / dtype
+
+        Raises
+        ------
+        :exc:`chunkstore.BadChunk`
+            If `slices` has wrong specification
+        :exc:`chunkstore.StoreUnavailable`
+            If interaction with chunk store failed (offline, bad auth, bad config)
+        """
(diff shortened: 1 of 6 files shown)


katdal/chunkstore.py
```diff
@@ -386,3 +432,38 @@ def put_dask_array(self, array_name, array, offset=()):
...
+
+        Returns
+        -------
+        success : :class:`dask.array.Array` object
+            Dask array of bools indicating presence of each chunk
+
+        Raises
+        ------
+        :exc:`chunkstore.StoreUnavailable`
+            If interaction with chunk store failed (offline, bad auth, bad config)
```
bmerry: Does it actually raise it, or do you get back a graph that raises it when it tries to compute?

ludwigschwardt: The latter. Not sure how to indicate this...

bmerry: I would just write about it in the general description, instead of having a section.

katdal/chunkstore_s3.py
```diff
@@ -63,6 +63,11 @@ class S3ChunkStore(ChunkStore):
     def __init__(self, client):
         if not botocore:
             raise _botocore_import_error
+        # XXX Hack to route head_object 404 error to ChunkNotFound via
+        # NoSuchKey. This is better than routing all ClientErrors to
+        # ChunkNotFound, but the best solution would be a fix to botocore.
+        boto_error_map = client.exceptions._code_to_exception
+        boto_error_map['404'] = boto_error_map['NoSuchKey']
```
bmerry: In case you wanted something that doesn't reach into the class internals, here's an alternative solution: catch the ClientError and check the error code: https://stackoverflow.com/questions/33842944/check-if-a-key-exists-in-a-bucket-in-s3-using-boto3

katdal/chunkstore.py
ludwigschwardt: Done. I'm leaving this out until I figure out good error handling for this part anyway.

katdal/chunkstore_s3.py

PR #118, opened 2018-02-07
SR919: Add ChunkStore-based VisFlagsWeights to TelstateDataSource

ludwigschwardt:
This is quite a hodge-podge of stuff to get the first v4 reader going:

- Open RDB files with `TelstateDataSource`
- Use the chunking metadata in telstate to build a `VisFlagsWeights` that speak to the chunk store (only RADOS for now)
- Make the dask arrays work like the old LazyIndexers

The next step would be to gussy up the V4 formatter based on this data source (and add S3 support).

### DIFF
--- a/katdal/chunkstore_rados.py
+++ b/katdal/chunkstore_rados.py
@@ -16,6 +16,9 @@
 
 """A store of chunks (i.e. N-dimensional arrays) based on the Ceph RADOS API."""
 
+import os
+import tempfile
+
 import numpy as np
 try:
     import rados
@@ -66,7 +69,7 @@ class RadosChunkStore(ChunkStore):
         Parameters
         ----------
         config : string or dict
-            Path to Ceph configuration file or config dict version of that file
+            Path to Ceph config file or contents of that file or config dict
         pool : string
             Name of the Ceph pool
         keyring : string, optional
@@ -86,8 +89,13 @@ class RadosChunkStore(ChunkStore):
         try:
             if isinstance(config, dict):
                 cluster = rados.Rados(conf=config)
-            else:
+            elif os.path.isfile(config):
                 cluster = rados.Rados(conffile=config)
+            else:
+                with tempfile.NamedTemporaryFile() as f:
+                    f.write(config)
+                    f.seek(0)
+                    cluster = rados.Rados(conffile=f.name)
             if keyring:
                 cluster.conf_set('keyring', keyring)
             if timeout is not None:
--- a/katdal/datasources.py
+++ b/katdal/datasources.py
@@ -95,20 +95,32 @@ class VisFlagsWeights(object):
     def shape(self):
         return self.vis.shape
 
-    @classmethod
-    def from_chunk_store(cls, store, base_name, chunk_info):
-        """"""
+
+class ChunkStoreVisFlagsWeights(VisFlagsWeights):
+    """Correlator data stored in a chunk store.
+
+    Parameters
+    ----------
+    store : :class:`ChunkStore` object
+        Chunk store
+    base_name : string
+        Name of dataset in store, as array name prefix (akin to a filename)
+    chunk_info : dict mapping array name to info dict
+        Dict specifying dtype, shape and chunks per array
+    """
+    def __init__(self, store, base_name, chunk_info):
+        self.store = store
         da = {}
         for array, info in chunk_info.iteritems():
             array_name = store.join(base_name, array)
             da[array] = store.get_dask_array(array_name, info['chunks'],
                                              info['dtype'])
-        return cls(DaskLazyIndexer(da['correlator_data']),
-                   DaskLazyIndexer(da['flags']),
-                   # Combine low-resolution weights and high-res weights_channel
-                   DaskLazyIndexer(da['weights'] *
-                                   da['weights_channel'][..., np.newaxis]),
-                   base_name)
+        vis = DaskLazyIndexer(da['correlator_data'])
+        flags = DaskLazyIndexer(da['flags'])
+        # Combine low-resolution weights and high-resolution weights_channel
+        weights = DaskLazyIndexer(da['weights'] *
+                                  da['weights_channel'][..., np.newaxis])
+        VisFlagsWeights.__init__(self, vis, flags, weights, base_name)
 
 
 class DataSource(object):
@@ -171,8 +183,7 @@ class TelstateDataSource(DataSource):
             ts_dtype = chunk_info['timestamps']['dtype']
             timestamps = store.get_dask_array(ts_name, ts_chunks, ts_dtype)
             timestamps = DaskLazyIndexer(timestamps)
-            data = VisFlagsWeights.from_chunk_store(store, base_name,
-                                                    chunk_info)
+            data = ChunkStoreVisFlagsWeights(store, base_name, chunk_info)
             DataSource.__init__(self, metadata, timestamps, data)
 
     @classmethod
--- a/katdal/datasources.py
+++ b/katdal/datasources.py
@@ -17,12 +17,17 @@
 """Various sources of correlator data and metadata."""
 
 import urlparse
+import os
 
 import katsdptelstate
-import numpy as np
+import redis
 
 from .dataset import AttrsSensors
 from .sensordata import TelstateSensorData
+
+
+class DataSourceNotFound(Exception):
+    """File associated with DataSource not found or server not responding."""
 
 
 class DataSource(object):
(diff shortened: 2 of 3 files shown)


katdal/chunkstore_rados.py
```diff
@@ -86,8 +89,13 @@ def from_config(cls, config, pool, keyring=None, timeout=5.):
         try:
             if isinstance(config, dict):
                 cluster = rados.Rados(conf=config)
-            else:
+            elif os.path.isfile(config):
                 cluster = rados.Rados(conffile=config)
```
bmerry: This is a little nasty. If some typo or screwup leads to the file not existing, then instead of a helpful exception that tells you the file wasn't found, you'll presumably get some sort of syntax error as it tries to parse the filename as a config file. I would prefer one of the following:
- add a separate parameter to take raw config text, and allow exactly one of them to be specified
- make writing a temporary file the caller's problem.

katdal/chunkstore_rados.py
```diff
@@ -86,8 +89,13 @@ def from_config(cls, config, pool, keyring=None, timeout=5.):
...
+            else:
+                with tempfile.NamedTemporaryFile() as f:
+                    f.write(config)
+                    f.seek(0)
```
bmerry: What is this for?

katdal/datasources.py
```diff
@@ -17,12 +17,110 @@
...
+
+
+class DaskLazyIndexer(object):
+    """Turn a dask Array into a LazyIndexer by computing it upon indexing."""
+    def __init__(self, dask_array):
+        self.da = dask_array
+
+    def __getitem__(self, keep):
+        return self.da[keep].compute()
+
```
bmerry: Is this intended to be fully compatible with LazyIndexer? The latter has `__len__`, `__iter__` and a `.name` attribute.

katdal/datasources.py
```diff
@@ -53,24 +151,73 @@ def name(self):
...
+            DataSource.__init__(self, metadata, None)
+        else:
+            # Extract VisFlagsWeights and timestamps from telstate
+            store = RadosChunkStore.from_config(telstate['ceph_conf'],
+                                                telstate['ceph_pool'])
+            ts_name = store.join(base_name, 'timestamps')
+            ts_chunks = chunk_info['timestamps']['chunks']
+            ts_dtype = chunk_info['timestamps']['dtype']
+            timestamps = store.get_dask_array(ts_name, ts_chunks, ts_dtype)
+            timestamps = DaskLazyIndexer(timestamps)
```
bmerry: Will having lazy timestamps be an issue for performance i.e. is katdal going to do a lot of operations on the timestamps array, each of which requires a fresh request to the store? Or is katdal already designed around timestamps being lazy-indexed?

katdal/datasources.py
```diff
@@ -53,24 +151,73 @@ def name(self):
...
+        url_parts = urlparse.urlparse(url, scheme='file')
+        kwargs = dict(urlparse.parse_qsl(url_parts.query))
+        kwargs['source_name'] = url_parts.geturl()
+        if url_parts.scheme == 'file':
+            # RDB dump file
+            telstate = katsdptelstate.TelescopeState()
+            try:
+                telstate.load_from_file(url_parts.path)
+            except OSError as err:
+                raise DataSourceNotFound(str(err))
```
bmerry: What do you think of using `six.raise_from`? That will make it easier to get back chained exceptions when we move to Python 3. I realise adding six as a dependency will make more work though.

katdal/datasources.py
```diff
@@ -53,24 +151,73 @@ def name(self):
...
+            return cls(telstate, **kwargs)
+        elif url_parts.scheme == 'redis':
+            # Redis server
+            try:
+                telstate = katsdptelstate.TelescopeState(url_parts.netloc)
```
bmerry: I know redis encodes the db number in the URL, I think in the query string. We should be able to pass that through.

katdal/datasources.py
```diff
@@ -53,24 +151,73 @@ def name(self):
...
+            except redis.exceptions.TimeoutError as err:
```
bmerry: What about redis.ConnectionError?

katdal/datasources.py
```diff
@@ -53,24 +151,73 @@ def name(self):
...
-        raise ValueError("Unsupported data source '%s'" % (url,))
+    try:
+        return TelstateDataSource.from_url(url)
+    except DataSourceNotFound as err:
+        # Amend the error message for the case of an IP address without scheme
+        url_parts = urlparse.urlparse(url, scheme='file')
+        if url_parts.scheme == 'file' and not os.path.isfile(url_parts.path):
+            raise DataSourceNotFound(
+                '{} (add a URL scheme if it is not meant to be a file)'
+                .format(str(err), url_parts.path))
```
bmerry: You've got one `{}` and two arguments to format. Have you tested it?

katdal/datasources.py
```diff
@@ -53,24 +151,73 @@ def name(self):
...
+    # raise ValueError("Unsupported data source {!r}".format(url))
```
bmerry: Did you mean to delete this?

katdal/datasources.py
```diff
@@ -53,24 +151,73 @@ def name(self):
...
-    else:
-        raise ValueError("Unsupported data source '%s'" % (url,))
+    try:
+        return TelstateDataSource.from_url(url)
+    except DataSourceNotFound as err:
+        # Amend the error message for the case of an IP address without scheme
+        url_parts = urlparse.urlparse(url, scheme='file')
+        if url_parts.scheme == 'file' and not os.path.isfile(url_parts.path):
+            raise DataSourceNotFound(
+                '{} (add a URL scheme if it is not meant to be a file)'
```
bmerry: "a URL" -> "an URL"? (depends whether you pronounce it "earl" or "you are ell" I guess)

katdal/chunkstore_rados.py
ludwigschwardt: Since RADOS is on its way out again I vote for the latter...

katdal/chunkstore_rados.py
ludwigschwardt: I assumed the file pointer would be at the end of the file... Enlighten me :-)

katdal/datasources.py
ludwigschwardt: Not really yet, although I could add those easily.

katdal/datasources.py
ludwigschwardt: This has already changed in the following V4 PR #119, plus this will all go away anyway once the timestamps are synthesised by me.

katdal/datasources.py
ludwigschwardt: Yes, I've left that kettle of fish for later, since `raise_from` buys you nothing on Python 2.

katdal/datasources.py
ludwigschwardt: Have we ever used a db other than 0?

katdal/datasources.py
ludwigschwardt: Not sure... A non-existent IP gave a TimeoutError. Ah, I see a bad port gives a ConnectionError. Will fix that then. You see why I like a `katsdptelstate.NotFound` error?

katdal/datasources.py
ludwigschwardt: I say the latter :-)

katdal/datasources.py
ludwigschwardt: Oops, no. katdal has no unit tests, remember? (at least at this level - ChunkStore puts the rest to shame!)

katdal/datasources.py
ludwigschwardt: It's a reminder for when I add other data sources, but yes I can probably toss it.

ludwigschwardt:
I'm going to add fixes to this PR to the next one (PR #119) to simplify my life...

katdal/chunkstore_rados.py
bmerry: It doesn't matter where the pointer associated with `f` is, because you're not passing `f` anywhere, you're only passing the filename, and whatever opens it will have its own file pointer.

However, I think in some implementations, seeking may also flush the file to the OS. You should probably have an explicit `f.flush()`.

katdal/datasources.py
bmerry: Not yet, but since telstate supports it we may as well carry the support through rather than silently doing the wrong thing.

bmerry:
> I'm going to add fixes to this PR to the next one (PR #119) to simplify my life...

Fair enough. Assuming none of the issues are going to break opening v3 files, go ahead and merge.

katdal/chunkstore_rados.py
ludwigschwardt: Good pointer... Forgot about that! Will add the flush.

ludwigschwardt: The `f.flush()` was necessary, BTW. Good reminder.

katdal/chunkstore_rados.py
ludwigschwardt: Done.

katdal/datasources.py
ludwigschwardt: I now extract db from the query and default to 0 if not there - is that OK? (One extra line of code vs a few more if I need to defer to TelescopeState's default db value).

katdal/datasources.py
ludwigschwardt: Done the simplest solution for now.

katdal/datasources.py
ludwigschwardt: Deleted.

katdal/datasources.py
ludwigschwardt: Fixed.

katdal/datasources.py
bmerry: Sounds fine to me.

PR #119, opened 2018-02-08
SR919: V4 formatter update

ludwigschwardt:
This brings the V4 formatter up to speed (note that this PR piggy-backs on #118). The main improvements are:

- Use telstate views
- Add vis / flags / weights indexers

An RDB file can currently be opened like this:

```Python
from katdal.datasources import open_data_source
from katdal.visdatav4 import VisibilityDataV4
ds = open_data_source('dump.20180207-0002.rdb?capture_block_id=20180207-0002-1517994195&stream_name=sdp_l0')
f = VisibilityDataV4(ds)
f.select(dumps=slice(0,10), ants='m003,m000,m017', channels=slice(0,1000))
f.vis[:].shape
(10, 1000, 24)
```

Output:

```
===============================================================================
Name: file:///dump.20180207-0002.rdb?capture_block_id=20180207-0002-1517994195&stream_name=sdp_l0 | test_vis_writer/20180207-0002-1517994195/sdp_l0 (version 4.0)
===============================================================================
Observer: sarah  Experiment ID: 20180207-0007
Description: 'MKAIV-387: CBF J1939+2134'
Observed from 2018-02-07 11:03:41.430 SAST to 2018-02-07 11:07:01.346 SAST
Dump rate / period: 0.12505 Hz / 7.997 s
Subarrays: 1
  ID  Antennas                            Inputs  Corrprods
   0  m000,m003,m015,m016,m017,m018,m019,m020,m021,m024,m027,m029,m031,m034,m035,m040  32      544
Spectral Windows: 1
  ID Band Product  CentreFreq(MHz)  Bandwidth(MHz)  Channels  ChannelWidth(kHz)
   0 L    bc856M4k   1284.000         856.000           4096       208.984
-------------------------------------------------------------------------------
Data selected according to the following criteria:
  subarray=0
  ants=['m018', 'm019', 'm040', 'm003', 'm000', 'm015', 'm016', 'm017', 'm021', 'm020', 'm034', 'm035', 'm024', 'm027', 'm031', 'm029']
  spw=0
-------------------------------------------------------------------------------
Shape: (25 dumps, 4096 channels, 544 correlation products) => Size: 445.645 MB
Antennas: m000,m003,m015,m016,m017,*m018,m019,m020,m021,m024,m027,m029,m031,m034,m035,m040  Inputs: 32  Autocorr: yes  Crosscorr: yes
Channels: 4096 (index 0 - 4095,  856.000 MHz - 1711.791 MHz), each 208.984 kHz wide
Targets: 1 selected out of 1 in catalogue
  ID  Name        Type      RA(J2000)     DEC(J2000)  Tags     Dumps  ModelFlux(Jy)
   0  J1939+2134  radec     19:39:38.56   21:34:59.1  Pulsars     25  
Scans: 1 selected out of 1 total       Compscans: 1 selected out of 1 total
  Date        Timerange(UTC)       ScanState  CompScanLabel  Dumps  Target
  07-Feb-2018/09:03:45 - 09:06:57    0:track    0:track         25    0:J1939+2134
```

Next steps are S3 support and better capture block / stream detection.

### DIFF
--- a/katdal/datasources.py
+++ b/katdal/datasources.py
@@ -31,23 +31,6 @@ class DataSourceNotFound(Exception):
     """File associated with DataSource not found or server not responding."""
 
 
-class DaskLazyIndexer(object):
-    """Turn a dask Array into a LazyIndexer by computing it upon indexing."""
-    def __init__(self, dask_array):
-        self.da = dask_array
-
-    def __getitem__(self, keep):
-        return self.da[keep].compute()
-
-    @property
-    def shape(self):
-        return self.da.shape
-
-    @property
-    def dtype(self):
-        return self.da.dtype
-
-
 class AttrsSensors(object):
     """Metadata in the form of attributes and sensors.
 
@@ -115,11 +98,10 @@ class ChunkStoreVisFlagsWeights(VisFlagsWeights):
             array_name = store.join(base_name, array)
             da[array] = store.get_dask_array(array_name, info['chunks'],
                                              info['dtype'])
-        vis = DaskLazyIndexer(da['correlator_data'])
-        flags = DaskLazyIndexer(da['flags'])
+        vis = da['correlator_data']
+        flags = da['flags']
         # Combine low-resolution weights and high-resolution weights_channel
-        weights = DaskLazyIndexer(da['weights'] *
-                                  da['weights_channel'][..., np.newaxis])
+        weights = da['weights'] * da['weights_channel'][..., np.newaxis]
         VisFlagsWeights.__init__(self, vis, flags, weights, base_name)
 
 
@@ -182,7 +164,8 @@ class TelstateDataSource(DataSource):
             ts_chunks = chunk_info['timestamps']['chunks']
             ts_dtype = chunk_info['timestamps']['dtype']
             timestamps = store.get_dask_array(ts_name, ts_chunks, ts_dtype)
-            timestamps = DaskLazyIndexer(timestamps)
+            # Make timestamps explicit, mutable (to be removed from store soon)
+            timestamps = timestamps.compute().copy()
             data = ChunkStoreVisFlagsWeights(store, base_name, chunk_info)
             DataSource.__init__(self, metadata, timestamps, data)
 
--- a/katdal/datasources.py
+++ b/katdal/datasources.py
@@ -95,20 +95,32 @@ class VisFlagsWeights(object):
     def shape(self):
         return self.vis.shape
 
-    @classmethod
-    def from_chunk_store(cls, store, base_name, chunk_info):
-        """"""
+
+class ChunkStoreVisFlagsWeights(VisFlagsWeights):
+    """Correlator data stored in a chunk store.
+
+    Parameters
+    ----------
+    store : :class:`ChunkStore` object
+        Chunk store
+    base_name : string
+        Name of dataset in store, as array name prefix (akin to a filename)
+    chunk_info : dict mapping array name to info dict
+        Dict specifying dtype, shape and chunks per array
+    """
+    def __init__(self, store, base_name, chunk_info):
+        self.store = store
         da = {}
         for array, info in chunk_info.iteritems():
             array_name = store.join(base_name, array)
             da[array] = store.get_dask_array(array_name, info['chunks'],
                                              info['dtype'])
-        return cls(DaskLazyIndexer(da['correlator_data']),
-                   DaskLazyIndexer(da['flags']),
-                   # Combine low-resolution weights and high-res weights_channel
-                   DaskLazyIndexer(da['weights'] *
-                                   da['weights_channel'][..., np.newaxis]),
-                   base_name)
+        vis = DaskLazyIndexer(da['correlator_data'])
+        flags = DaskLazyIndexer(da['flags'])
+        # Combine low-resolution weights and high-resolution weights_channel
+        weights = DaskLazyIndexer(da['weights'] *
+                                  da['weights_channel'][..., np.newaxis])
+        VisFlagsWeights.__init__(self, vis, flags, weights, base_name)
 
 
 class DataSource(object):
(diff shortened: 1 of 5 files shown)


bmerry:
It looks like this is a stacked PR. I'll wait until the previous one is finalised before reviewing, unless it's urgent.

katdal/datasources.py
```diff
@@ -186,9 +192,9 @@ def from_url(cls, url):
         elif url_parts.scheme == 'redis':
             # Redis server
             try:
-                telstate = katsdptelstate.TelescopeState(url_parts.netloc)
-            except redis.exceptions.TimeoutError as err:
-                raise DataSourceNotFound(str(err))
+                telstate = katsdptelstate.TelescopeState(url_parts.netloc, db)
+            except (redis.ConnectionError, redis.exceptions.TimeoutError) as e:
```
bmerry: I think both are defined in `redis.exceptions` but imported into `redis`. So you may as well be consistent and get them both from the same module.

katdal/visdatav4.py
```diff
@@ -103,43 +107,28 @@ def __init__(self, source, ref_ant='', time_offset=0.0, centre_freq=None,
...
 
         # Assemble sensor cache
-        self.sensor = SensorCache(source.metadata.sensors, self._timestamps,
+        self.sensor = SensorCache(source.metadata.sensors, source.timestamps,
                                   self.dump_period, self._time_keep,
                                   SENSOR_PROPS, VIRTUAL_SENSORS, SENSOR_ALIASES)
 
+        # ------ Extract flags ------
+
+        self._flags_select = np.array([0], dtype=np.uint8)
```
bmerry: Isn't this going to be overwritten by the setter invoked on the next line?

katdal/lazy_indexer.py
```diff
@@ -330,3 +330,39 @@ def dtype(self):
...
+    def __init__(self, dataset, keep=(), transforms=None):
+        self.name = getattr(dataset, 'name', '')
+        try:
+            dataset = dataset[keep]
+        except NotImplementedError:
+            # Dask does not like multiple boolean indices: go one dim at a time
+            for dim, keep_per_dim in enumerate(keep):
+                index = dataset.ndim * [slice(None)]
+                index[dim] = keep_per_dim
+                dataset = dataset[tuple(index)]
```
bmerry: I think this is reimplementing `da.take(dataset, index, axis=dim)`.

katdal/lazy_indexer.py
```diff
@@ -330,3 +330,39 @@ def dtype(self):
...
+        self.dataset = dataset
+        self.transforms = [] if transforms is None else transforms
```
bmerry: Where do these get applied?

katdal/visdatav4.py
```diff
@@ -343,6 +295,37 @@ def __init__(self, source, ref_ant='', time_offset=0.0, centre_freq=None,
...
+                logger.warning("%r is not a legitimate flag type, "
+                               "supported ones are %s", name, FLAG_NAMES)
+        # Pack index list into bit mask
+        # Reverse flag indices as np.packbits has bit 0 as the MSB (we want LSB)
+        flagmask = np.packbits(np.flipud(selection))
+        if not flagmask:
+            logger.warning('No valid flags were selected - setting all flags '
+                           'to False by default')
+        self._flags_select = flagmask
+
```
bmerry: Is this common code that could be pulled out into its own class and shared with v3? Maybe it will make more sense to do so once v3 has been dask-ified though.

katdal/datasources.py

katdal/lazy_indexer.py
ludwigschwardt: They don't - I left them for compatibility with the old indexer. I'll just force it to be empty.

katdal/visdatav4.py
ludwigschwardt: Sure - it's a "default default" that reminds me that such a variable exists, otherwise it magically appears later on.

katdal/visdatav4.py
ludwigschwardt: Yes, agreed.

katdal/lazy_indexer.py
ludwigschwardt: Good point - I've never used `take` before, hence the ignorance :-)

katdal/lazy_indexer.py
bmerry: If they're not applied then it's not really compatible. Maybe remove the constructor argument so that if someone tries to actually pass a transform they get an error early. Leaving the attribute (as an empty list) would help with compatibility.

katdal/lazy_indexer.py
ludwigschwardt: BTW, hopefully this workaround will soon be superfluous if dask arrays get around to to outer indexing aka oindex.

katdal/visdatav4.py
bmerry: Fair enough. Reading through the constructor it looks like this is the value it has at the end of construction though - a comment might help.

katdal/lazy_indexer.py
ludwigschwardt: Hmm, I might do this later as it would make dask an install requirement for katdal (it's only a test requirement and needed for the optional v4 stuff for now).

katdal/lazy_indexer.py
```diff
@@ -341,11 +341,12 @@ def __init__(self, dataset, keep=(), transforms=None):
         except NotImplementedError:
             # Dask does not like multiple boolean indices: go one dim at a time
             for dim, keep_per_dim in enumerate(keep):
+                # Use da.take(dataset, index, axis=dim) once dask is required
```
bmerry: Presumably dask is required for DaskLazyIndexer - so you could `import dask` inside `__init__`.

katdal/visdatav4.py
ludwigschwardt: Commented and changed to 255 instead (the intended initial value).

katdal/lazy_indexer.py
ludwigschwardt: Not required but implied... I've gone this route with a note to change it eventually.

katdal/lazy_indexer.py
ludwigschwardt: Fixed.

PR #120, opened 2018-02-09
Add support for an S3 chunk store to TelstateDataSource

ludwigschwardt:
A quick one...

If the `s3_endpoint_url` key is found in telstate, prefer that to the RADOS option. The chunk_name is gone for now (since it should be simply the capture stream name), but it might make a comeback once the data source builds from a viewed-up telstate only.

### DIFF
--- a/katdal/datasources.py
+++ b/katdal/datasources.py
@@ -18,16 +18,13 @@
 
 import urlparse
 import os
-import tempfile
 
 import katsdptelstate
 import redis
 import numpy as np
 
 from .sensordata import TelstateSensorData
-from .chunkstore_rados import RadosChunkStore
 from .chunkstore_s3 import S3ChunkStore
-from .chunkstore import StoreUnavailable
 
 
 class DataSourceNotFound(Exception):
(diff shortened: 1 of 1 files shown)


katdal/datasources.py
```diff
@@ -152,25 +154,27 @@ def __init__(self, telstate, capture_block_id=None, stream_name=None,
...
         else:
-            # Extract VisFlagsWeights and timestamps from telstate
-            with tempfile.NamedTemporaryFile() as f:
-                f.write(telstate['ceph_conf'])
-                f.flush()
-                pool = telstate['ceph_pool']
-                store = RadosChunkStore.from_config(f.name, pool)
-            ts_name = store.join(base_name, 'timestamps')
+            try:
+                store = S3ChunkStore.from_url(telstate['s3_endpoint_url'])
```
bmerry: You'll get a KeyError if `s3_endpoint_url` isn't there.

katdal/datasources.py
```diff
@@ -152,25 +154,27 @@ def __init__(self, telstate, capture_block_id=None, stream_name=None,
...
+            except StoreUnavailable:
+                # Extract VisFlagsWeights and timestamps from telstate
```
bmerry: Should this comment be above the try? (since it applies to S3 as well)?

katdal/datasources.py
ludwigschwardt: Fixed. S3 is also now the only option.

katdal/datasources.py

PR #121, opened 2018-02-12
Autodetect program block ID and stream name

ludwigschwardt:
The default TelstateDataSource constructor now takes a telstate that
already has views associated with the capture block and stream. This
requires the reintroduction of `chunk_name` to figure out the chunk
store name of the dataset (as `capture_stream` is now implicit again).
This is the intended way to open pipeline data.

When constructing telstate from an URL with no prior views, detect
the capture block ID and stream name. Crash if there are none and warn
if there are more than one. Then prepare the telstate with the relevant
views before passing on to the usual constructor.

### DIFF
--- a/katdal/datasources.py
+++ b/katdal/datasources.py
@@ -18,6 +18,7 @@
 
 import urlparse
 import os
+import logging
 
 import katsdptelstate
 import redis
@@ -25,6 +26,9 @@ import numpy as np
 
 from .sensordata import TelstateSensorData
 from .chunkstore_s3 import S3ChunkStore
+
+
+logger = logging.getLogger(__name__)
 
 
 class DataSourceNotFound(Exception):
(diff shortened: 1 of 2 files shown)


katdal/datasources.py
```diff
@@ -131,18 +135,50 @@ def name(self):
...
+                if 'obs_params' in telstate.view(cbid, exclusive=True):
+                    capture_blocks.append(cbid)
+        if not capture_blocks:
+            raise ValueError('No capture block IDs found in telstate - '
+                             'please specify it manually')
+        # Pick the latest capture block
+        capture_block_id = capture_blocks[-1]
+        if len(capture_blocks) > 1:
+            logger.warn('Telstate has more than one capture block - {} - '
+                        'picking the latest'.format(capture_blocks))
```
bmerry: `warn` is [deprecated](https://docs.python.org/3/library/logging.html?highlight=warning#logging.warning) in favour of `warning`. Also do logger formatting with %d and *args.

katdal/datasources.py
```diff
@@ -131,18 +135,50 @@ def name(self):
...
+    # Detect the captured stream
+    if not stream_name:
+        streams = []
+        for stream in telstate['sdp_config']['outputs']:
+            capture_stream = telstate.SEPARATOR.join((capture_block_id, stream))
```
bmerry: I would still check the type of the stream as recorded in sdp_config, to make sure that it's an sdp.l0 stream.

I think we also need to have a face-to-face chat (probably with @sratcliffe) about how this will all work with the flags stream. At the moment it's not explicitly configured as a separate output in sdp_config - it's just on the side of the cal output.

katdal/datasources.py
```diff
@@ -131,18 +135,50 @@ def name(self):
...
+                                                              stream_name))
+    capture_stream = telstate.SEPARATOR.join((capture_block_id, stream_name))
+    telstate = telstate.view(stream_name)
+    telstate = telstate.view(capture_block_id)
+    telstate = telstate.view(capture_stream)
+    return telstate
+
+
 class TelstateDataSource(DataSource):
     """A data source based on :class:`katsdptelstate.TelescopeState`."""
```
bmerry: Now definitely needs documentation on what `telstate` should be.

katdal/datasources.py
ludwigschwardt: Dang, I always used that because it is shorter (gotta cram those log messages into 79 lines!).

katdal/datasources.py
ludwigschwardt: I did the check and then decided it's probably superfluous and removed it, as anything with "chunk_info" will be katdal-related. But I could add it back in again.

katdal/datasources.py
ludwigschwardt: Sure.

katdal/datasources.py
bmerry: For now. But who knows, maybe someone will be inspired and start writing beamformer data in chunks :-)

katdal/datasources.py

katdal/datasources.py

katdal/datasources.py

PR #122, opened 2018-02-16
SR-1052: katdal.open() a V4 file

ludwigschwardt:
Synchronise v4 format with new `CaptureSession` as we now have `obs_params` as an attribute and `sdp_capture_block_id` needs to lose its unicodeness.

Then plumb all the goodness straight into `katdal.open()` and declare success!

### DIFF
--- a/katdal/__init__.py
+++ b/katdal/__init__.py
@@ -212,6 +212,7 @@ a back door installed at `d.file` in the case of a single-file data set.
 """
 
 import logging as _logging
+import urlparse
 
 from .dataset import DataSet, WrongVersion
 from .lazy_indexer import LazyTransform
@@ -330,7 +331,7 @@ def open(filename, ref_ant='', time_offset=0.0, **kwargs):
     datasets = []
     for f in filenames:
         # V4 RDB file with optional URL-style query string
-        if f.split('?')[0].endswith('.rdb'):
+        if urlparse.urlsplit(f).path.endswith('.rdb'):
             from .datasources import open_data_source
             from .visdatav4 import VisibilityDataV4
             dataset = VisibilityDataV4(open_data_source(f),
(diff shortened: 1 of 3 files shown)


ludwigschwardt:
Sorry about that - a bait and switch :-)

katdal/__init__.py
```diff
@@ -329,7 +329,14 @@ def open(filename, ref_ant='', time_offset=0.0, **kwargs):
     filenames = [filename] if isinstance(filename, basestring) else filename
     datasets = []
     for f in filenames:
-        dataset = _file_action('__call__', f, ref_ant, time_offset, **kwargs)
+        # V4 RDB file with optional URL-style query string
+        if f.split('?')[0].endswith('.rdb'):
```
bmerry: This does look pretty hacky. For one thing, if someone chooses to put a `?` in their filename, it's going to mysteriously fail. Maybe rather check for `://` or even `.startswith('file:///')`?

ludwigschwardt: Of course it's hacky. I called it with `'dump.20180215-0017.rdb?capture_block_id=00000000-00000-1518711320'` and I don't expect anyone else to call it with a ?. This will all be cleared up soon (hopefully after March :-)).

ludwigschwardt:
To answer your previous questions:

- This why signal displays and cal should consider katdal once the dust settles. I'll check into this.
- For v3 I intend to send a PR on katdspfilewriter that uses a viewed telstate when the TelescopeModel is derived, and then in katdal v3 I'll check for an attribute as well as a sensor as fallback.

katdal/__init__.py
bmerry: How about using
```python
if urlparse.urlsplit(f).path.endswith('.rdb'):
```
At least then it will handle the vagaries of URL parsing. It looks like urlparse is happy to be given a relative URL aka a filename.

ludwigschwardt: Done. Although it still throws out a filename with a `?`.

PR #125, opened 2018-02-27
Various v4 mop-up operations

ludwigschwardt:
- Allow chunk store to be overridden (opens NPY files during testing instead of S3)
- Fix mxxx_observer which is now an attribute
- Shorten sensor names by discarding the most relevant prefix from the telstate key

### DIFF
--- a/katdal/datasources.py
+++ b/katdal/datasources.py
@@ -219,18 +219,16 @@ def _shorten_key(telstate, key):
 
     Returns
     -------
-    short_key : string or None
-        Suffix of `key` after subtracting first matching prefix, or None if
-        `key` does not start with any of the prefixes
+    short_key : string
+        Suffix of `key` after subtracting first matching prefix, or empty
+        string if `key` does not start with any of the prefixes (or exactly
+        matches a prefix, which is also considered pathological)
 
     """
     for prefix in telstate.prefixes:
-        if not prefix:
-            return key
-        head, sep, tail = key.partition(prefix)
-        if not head and sep == prefix:
-            return tail
-    return None
+        if key.startswith(prefix):
+            return key[len(prefix):]
+    return ''
 
 
 class TelstateDataSource(DataSource):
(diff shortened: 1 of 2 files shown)


katdal/datasources.py
```diff
@@ -207,6 +207,32 @@ def view_capture_stream(telstate, capture_block_id=None, stream_name=None):
...
+    -------
+    short_key : string or None
+        Suffix of `key` after subtracting first matching prefix, or None if
+        `key` does not start with any of the prefixes
+
+    """
+    for prefix in telstate.prefixes:
+        if not prefix:
+            return key
+        head, sep, tail = key.partition(prefix)
```
bmerry: I had not come across this function before. But it seems slightly roundabout - I would have written the whole loop content as:
```python
if key.startswith(prefix):
    return key[len(prefix):]
```

katdal/datasources.py
```diff
@@ -220,38 +246,44 @@ class TelstateDataSource(DataSource):
...
-    def __init__(self, telstate, source_name='telstate'):
+    def __init__(self, telstate, chunk_store=None, source_name='telstate'):
         self.telstate = telstate
         # Collect sensors
         sensors = {}
         for key in telstate.keys():
             if not telstate.is_immutable(key):
-                sensors[key] = TelstateSensorData(telstate, key)
+                sensor_name = _shorten_key(telstate, key)
+                if sensor_name:
```
bmerry: I don't expect it to happen, but what if the sensor name is itself a prefix? In that case `sensor_name` will be the empty string here, and hence falsey.

katdal/datasources.py
ludwigschwardt: Good point.

katdal/datasources.py
ludwigschwardt: I'll happily ignore those pathological sensors with the current code.

PR #126, opened 2018-03-12
Authenticated katdal

ctgschollar:
My authenticated katdal version. It is not really necessary to include the import as part of katdal itself. Technically simply importing katadpauth.auth_botocore in any script will patch anything in it using botocore to use my authentication. I think it may be a good idea to create a small script (auth_katdal or something) which simply imports katdal and katsdpauth.auth_botocore which will allow us to not perform the pull in your code.

This current version will not ask you to log in unless you are trying to access a protected resource. If you are on a whitelisted machine, katdal will simply run as it currently does, otherwise you will be prompted to log in.

### DIFF
--- a/katdal/chunkstore_s3.py
+++ b/katdal/chunkstore_s3.py
@@ -20,7 +20,10 @@ import contextlib
 
 import numpy as np
 try:
-    import katsdpauth.auth_botocore
+    try:
+        import katsdpauth.auth_botocore
+    except:
+        import botocore
     _botocore_import_error = None
 except ImportError as e:
     botocore = None
--- a/katdal/chunkstore_s3.py
+++ b/katdal/chunkstore_s3.py
@@ -20,7 +20,7 @@ import contextlib
 
 import numpy as np
 try:
-    import botocore
+    import katsdpauth.auth_botocore
     _botocore_import_error = None
 except ImportError as e:
     botocore = None
--- a/setup.py
+++ b/setup.py
@@ -62,6 +62,6 @@ setup(name="katdal",
         # rados is not in PyPI but available as Debian package python-rados
         'rados': ['rados'],
         # Only available throught github currently
-        'auth' : ['katsdpauth'], 
+        'auth': ['katsdpauth'], 
       },
       tests_require=['nose', 'dask[array]'])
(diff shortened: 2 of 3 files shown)


katdal/chunkstore_s3.py
```diff
@@ -20,7 +20,7 @@
 
 import numpy as np
 try:
-    import botocore
+    import katsdpauth.auth_botocore
```
ludwigschwardt: Did you mean `import katsdpauth.auth_botocore as botocore`? Because `botocore` still needs to come from somewhere.

setup.py
```diff
@@ -60,6 +60,8 @@
         'ms': ['python-casacore >= 2.2.1'],
         's3': ['botocore'],
         # rados is not in PyPI but available as Debian package python-rados
-        'rados': ['rados']
+        'rados': ['rados'],
+        #Only available throught github currently
+        'authed':['katsdpauth'],
```
ludwigschwardt: I would just call it 'auth' here. Also needs a space after colon and comment hash.

katdal/chunkstore_s3.py
ctgschollar: Actually seeing as you import botocore.session and botocore,config separately and you only use them in your code it works with just the katsdpauth.auth_botocore. Although if you were to try using some other function from botocore we would need to import botocore and auth_botocore. auth_botocore does import botocore, so it should still fail if the machine doesn't have botocore installed. At least with my testing so far it works.

setup.py
```diff
@@ -60,6 +60,8 @@
         'ms': ['python-casacore >= 2.2.1'],
         's3': ['botocore'],
         # rados is not in PyPI but available as Debian package python-rados
-        'rados': ['rados']
+        'rados': ['rados'],
+        # Only available throught github currently
+        'auth' : ['katsdpauth'],
```
ludwigschwardt: Sorry to be pedantic, but now you have one space too many :-)

katdal/chunkstore_s3.py
ludwigschwardt: Good point, that initial import is just a tester - I missed the `else:` code in the review.

You still need to make `katsdpauth` optional... Something like

```Python
try:
    try:
        import katsdpauth.auth_botocore
    except ImportError:
        import botocore
    _botocore_import_error = None
except ImportError as e:
    ....
```

setup.py
ctgschollar: Haha, whoops.

ctgschollar:
Can I merge this ludwig?

PR #127, opened 2018-03-13
 Remove CBID / stream autodetection

ludwigschwardt:
Rather rely on the top-level attributes introduced in RDB files, viz `capture_block_id`, `stream_name` and `stream_type`. If any of them are missing, insist on a manual override. Also bail if the stream type is not "sdp.vis".

There is also no more need for `sdp_config` and `sdp_capture_block_id` in the "lite" RDB files.

### DIFF
--- a/katdal/datasources.py
+++ b/katdal/datasources.py
@@ -138,10 +138,9 @@ class DataSource(object):
 def view_capture_stream(telstate, capture_block_id=None, stream_name=None):
     """Create telstate view based on capture block ID and stream name.
 
-    This figures out the appropriate capture block ID (the latest one with
-    obs_params) and L0 stream name (the first one with chunk_info), or use
-    the provided ones. It then constructs a view on `telstate` with at least
-    the prefixes
+    This figures out the appropriate capture block ID and L0 stream name from
+    a capture-stream specific telstate, or use the provided ones. It then
+    constructs a view on `telstate` with at least the prefixes
 
       - <capture_block_id>_<stream_name>
       - <capture_block_id>
@@ -168,37 +167,25 @@ def view_capture_stream(telstate, capture_block_id=None, stream_name=None):
     """
     # Detect the capture block
     if not capture_block_id:
-        capture_blocks = []
-        if 'sdp_capture_block_id' in telstate:
-            for value_time in telstate.get_range('sdp_capture_block_id', st=0):
-                cbid = str(value_time[0])
-                if 'obs_params' in telstate.view(cbid, exclusive=True):
-                    capture_blocks.append(cbid)
-        if not capture_blocks:
+        try:
+            capture_block_id = telstate['capture_block_id']
+        except KeyError:
             raise ValueError('No capture block IDs found in telstate - '
                              'please specify it manually')
-        # Pick the latest capture block
-        capture_block_id = capture_blocks[-1]
-        if len(capture_blocks) > 1:
-            logger.warning('Telstate has more than one capture block - %s - '
-                           'picking the latest', capture_blocks)
     # Detect the captured stream
     if not stream_name:
-        streams = []
-        for stream, config in telstate['sdp_config']['outputs'].items():
-            if config['type'] != 'sdp.l0':
-                continue
-            capture_stream = telstate.SEPARATOR.join((capture_block_id, stream))
-            if 'chunk_info' in telstate.view(capture_stream, exclusive=True):
-                streams.append(stream)
-        if not streams:
+        try:
+            stream_name = telstate['stream_name']
+            stream_type = telstate['stream_type']
+        except KeyError:
             raise ValueError('No captured streams found in telstate - '
                              'please specify the stream manually')
-        stream_name = streams[0]
-        if len(streams) > 1:
-            logger.warning('Telstate has more than one captured stream - %s - '
-                           'picking the first', streams)
-    logger.info('Found capture block %s and stream %s',
+        expected_type = 'sdp.vis'
+        if telstate['stream_type'] != expected_type:
+            raise ValueError("Found stream {!r} but it has the wrong type {!r},"
+                             " expected {!r}".format(stream_name, stream_type,
+                                                     expected_type))
+    logger.info('Using capture block %r and stream %r',
                 capture_block_id, stream_name)
     capture_stream = telstate.SEPARATOR.join((capture_block_id, stream_name))
     telstate = telstate.view(stream_name)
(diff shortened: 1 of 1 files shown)


katdal/datasources.py
```diff
@@ -168,37 +167,25 @@ def view_capture_stream(telstate, capture_block_id=None, stream_name=None):
...
-        if 'sdp_capture_block_id' in telstate:
-            for value_time in telstate.get_range('sdp_capture_block_id', st=0):
-                cbid = str(value_time[0])
-                if 'obs_params' in telstate.view(cbid, exclusive=True):
-                    capture_blocks.append(cbid)
-        if not capture_blocks:
+        try:
+            capture_block_id = str(telstate['capture_block_id'])
+        except KeyError:
             raise ValueError('No capture block IDs found in telstate - '
```
sratcliffe: Off review, but I guess this is now ...block ID found in telstate... (and the same below for ...captured stream found in...) since we only ever have one.

katdal/datasources.py
```diff
@@ -168,37 +167,25 @@ def view_capture_stream(telstate, capture_block_id=None, stream_name=None):
...
     if not capture_block_id:
-        capture_blocks = []
-        if 'sdp_capture_block_id' in telstate:
-            for value_time in telstate.get_range('sdp_capture_block_id', st=0):
-                cbid = str(value_time[0])
-                if 'obs_params' in telstate.view(cbid, exclusive=True):
-                    capture_blocks.append(cbid)
-        if not capture_blocks:
+        try:
+            capture_block_id = str(telstate['capture_block_id'])
```
bmerry: For my education, why do we have both sdp_capture_block_id and capture_block_id? Can we unify them?

sratcliffe: capture_block_id and the stream_name and stream_type are inserted by the meta writer just before writing, since at this point we are totally sure about the CBID and stream for this dump. I need to go and check where sdp_capture_block_id comes from and how it choses which CBID to use.

bmerry: sdp_capture_block_id is a mutable key written by master controller as it inits each capture block - it's used by the session to find the current CBID for example.

Would it be sensible to use that same name for the key injected by the metadata writer?

sratcliffe: My reasoning for choosing a non sdp key is exactly to distinguish from these kinds of internal things - the idea being this is definitive statement (sdp or not) about the data in this RDB represents - so basically a system property.

bmerry: I'm still far from convinced, but let's bikeshed it face-to-face.

katdal/datasources.py
```diff
@@ -168,37 +167,25 @@ def view_capture_stream(telstate, capture_block_id=None, stream_name=None):
...
-        for stream, config in telstate['sdp_config']['outputs'].items():
-            if config['type'] != 'sdp.l0':
-                continue
-            capture_stream = telstate.SEPARATOR.join((capture_block_id, stream))
-            if 'chunk_info' in telstate.view(capture_stream, exclusive=True):
-                streams.append(stream)
-        if not streams:
+        try:
+            stream_name = str(telstate['stream_name'])
+            stream_type = str(telstate['stream_type'])
```
bmerry: Can we look up the stream type from within the `<stream>_` namespace? That's where it is now written by ingest, and it'll remove the need for the metadata writer to copy it to a top-level key.

sratcliffe: No - we are just rehashing the bikeshed now. There is no prior knowledge of the stream or cbid when katdal opens a file - we needed a simple, unambiguous way to tell what cbid/stream this file represents. The *only* way to ensure this is to have meta writer insert these values just before creating the file, and I don't want meta writer to actually mess with the live telstate.

bmerry: Sure, it needs to insert the name, but does it need to insert the type as well? Once you have the name you can get to the namespace to get the type (which is already in telstate).

katdal/datasources.py
ludwigschwardt: Point taken.

katdal/datasources.py
ludwigschwardt: Remember to invite me too :-)

bmerry: So far Simon has beaten me into leaving it as is, with the possibility to later rename sdp_capture_block_id to capture_block_id. The reason for not doing it now is that the metadata writer has no way to replace keys, only to add new ones, and it would require some refactoring to support that.

katdal/datasources.py
ludwigschwardt: That is a good point. I'll just move the `.view(stream)` earlier, so similar effort for fewer telstate keys.

katdal/datasources.py
ludwigschwardt: I think if you simply add the self-referential `<cbid>_capture_block_id` (yes, another one!) all will be well and whole. The `sdp_capture_block_id` is very useful to find all CBs in a multi-CB telstate (how else?). The `<cbid>_capture_block_id` is meant to be "the" capture block ID in a viewed single-CB telstate. Then @sratcliffe's improper `capture_block_id` is a similar construct for an unviewed single-CB telstate, which is styled to match `<cbid>_capture_block_id` after applying its view.

I'm not in favour of proper telstate keys at the top level that do not start with an identifiable subsystem / component, e.g. `sdp_`, so I would not like to see `sdp_capture_block_id` turn into `capture_block_id`.

ludwigschwardt: Also, another avenue I mentioned to @sratcliffe is that we could "preview" all keys in the relevant <stream>, <cbid> and <cbid_stream> views by stripping off their prefixes and promoting them to the top level in the RDB files. Then you could open the file and and you would automatically be in the correct view. This also resolves the discrepancy between `capture_block_id` and `sdp_capture_block_id` (the former is a flattened version of the self-referential `<cbid>_capture_block_id` and therefore not a standalone key).

ludwigschwardt:
Are you guys happy with this?

katdal/datasources.py
```diff
@@ -138,10 +138,9 @@ def name(self):
 def view_capture_stream(telstate, capture_block_id=None, stream_name=None):
     """Create telstate view based on capture block ID and stream name.
 
-    This figures out the appropriate capture block ID (the latest one with
-    obs_params) and L0 stream name (the first one with chunk_info), or use
-    the provided ones. It then constructs a view on `telstate` with at least
-    the prefixes
+    This figures out the appropriate capture block ID and L0 stream name from
+    a capture-stream specific telstate, or use the provided ones. It then
```
bmerry: use -> uses

PR #128, opened 2018-03-13
Make writing an NPY file atomic

ludwigschwardt:
This writes to a temporary filename first (ending in '.writing.npy')
before renaming it to the final '.npy' file after the file has been
closed. This ensures that other processes waiting for the file (like
the vis trawler) knows when it is safe to pick it up.

Also fix a grammar bug.


ludwigschwardt:
Of course now the unit tests pass fine on Jenkins, and fakes3 always hangs on my laptop, even with older katdal code. So my test environment has some rot...

PR #129, opened 2018-03-14
Store NPY files directly in S3

ludwigschwardt:
Instead of storing the chunk array buffer as a binary string in S3, rather store the string representation of the corresponding NPY file, which is the buffer along with a (small) additional header.

This has several advantages:

- The vis trawler and other similar tools can upload NPY files straight into S3 (or download them the same way). This is a very efficient way to convert an `NpyFileChunkStore` to an `S3ChunkStore`.

- The header contains the shape and, more important, the dtype of the chunk. This allows us to remove the `dtype` parameter being passed around all over `ChunkStore` and brings it closer to the dask.array idea of a store (as used in the `dask.array.store` function), or even [zarr](https://github.com/zarr-developers/zarr).

- Benchmarks indicate that the new format somehow *speeds up* `get_chunk` by some 15% (maybe because the NPY routines do a buffered read from the socket, but more likely the original `get_chunk` was somehow inefficient). The `put_chunk` is slightly slower, but it's not expected to be the main data conduit for MeerKAT.

The change is also indicated by adding ".npy" to the S3 keys.

This whole thing is obviously endlessly refactorable and extendable to `RadosChunkStore` and even `NpyFileChunkStore` itself. But let's not go there... yet.

### DIFF
--- a/katdal/chunkstore_s3.py
+++ b/katdal/chunkstore_s3.py
@@ -161,10 +161,13 @@ class S3ChunkStore(ChunkStore):
             header = {'shape': shape, 'fortran_order': False,
                       'descr': np.lib.format.dtype_to_descr(dtype)}
             fp = io.BytesIO()
-            np.lib.format._write_array_header(fp, header)
-            header_size = fp.tell()
-            expected_bytes = int(np.prod(shape)) * dtype.itemsize + header_size
-            return actual_bytes == expected_bytes
+            np.lib.format.write_array_header_1_0(fp, header)
+            header_size_v1 = fp.tell()
+            fp.seek(0)
+            np.lib.format.write_array_header_2_0(fp, header)
+            header_size_v2 = fp.tell()
+            data_size = int(np.prod(shape)) * dtype.itemsize
+            return actual_bytes - data_size in (header_size_v1, header_size_v2)
 
     get_chunk.__doc__ = ChunkStore.get_chunk.__doc__
     put_chunk.__doc__ = ChunkStore.put_chunk.__doc__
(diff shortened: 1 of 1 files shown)


katdal/chunkstore_s3.py
```diff
@@ -156,7 +158,12 @@ def has_chunk(self, array_name, slices, dtype):
             return False
         else:
             actual_bytes = response['ContentLength']
-            expected_bytes = int(np.prod(shape)) * dtype.itemsize
+            header = {'shape': shape, 'fortran_order': False,
+                      'descr': np.lib.format.dtype_to_descr(dtype)}
+            fp = io.BytesIO()
+            np.lib.format._write_array_header(fp, header)
```
bmerry: I was going to suggest using write_array_header_2_0 since it's public, but that won't work because it won't necessarily match the actual header. In general, you can't be sure what the header version is used on the object without reading it (e.g. at some point numpy might change its logic that autoselects a version, or create a version 3). Does S3 have a way you can actually read back the header without reading the entire object? Otherwise your best bet might be to try version 1 and version 2 and accept it if either size matches.

ludwigschwardt: Well, the header should match if we always use v2 when writing and reading NPY. But then again, it might confuse some future (external) users who are not aware of the distinction and tries to simply `np.save` some data for reading by `katdal`. Or maybe that is improbable.

The point of `has_chunk` is to avoid reading the object, which is what `head_object` does. Most of the time the object won't be there at all - the size check is just a secondary thing *because we can*. In fact, for `S3ChunkStore.has_dask_array` I am considering checking the name only using a much faster method.

Or we could just keep using the internal numpy function and update it if numpy ever changes it (they'll surely always have a "write the header" function).

bmerry: > the size check is just a secondary thing because we can

Indeed - so whatever you choose to do, it should probably accept any valid numpy header, even if it is a different version than the one put_chunk would use in the current environment.

PR #130, opened 2018-03-14
Improve autodetection of chunk store

ludwigschwardt:
For RDB files this first attempts to locate an `NpyFileChunkStore` in an adjacent directory named after the capture stream, as is typically produced by `vis_writer` and other tools. Failing this, it falls back to the S3 gateway suggested by telstate. Another fail here is treated as an error (as this is typically due to authentication or network connection errors that the user wants to know about). The redis server option only attempts the S3 option, since there is no file access.

If the user really wants no chunk store it can be specified as `None`, which is useful for metadata extraction.

This allows the dream of `katdal.open('whatever')`, at least for RDB files :-)

### DIFF
--- a/katdal/datasources.py
+++ b/katdal/datasources.py
@@ -26,6 +26,8 @@ import numpy as np
 
 from .sensordata import TelstateSensorData
 from .chunkstore_s3 import S3ChunkStore
+from .chunkstore_npy import NpyFileChunkStore
+from .chunkstore import StoreUnavailable
 
 
 logger = logging.getLogger(__name__)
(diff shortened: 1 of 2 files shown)


katdal/datasources.py
```diff
@@ -283,16 +283,30 @@ def from_url(cls, url, chunk_store=None):
...
+            # Look for adjacent data directory (presumably containing NPY files)
+            if chunk_store == 'auto':
+                capture_block_path = os.path.dirname(url_parts.path)
+                store_path = os.path.join(capture_block_path, '..')
+                try:
+                    data_path = os.path.join(store_path, telstate['chunk_name'])
+                except KeyError:
+                    chunk_store = None
+            if chunk_store == 'auto' and os.path.isdir(data_path):
+                chunk_store = NpyFileChunkStore(os.path.normpath(store_path))
```
bmerry: If you're going to use normpath, do it before checking `isdir`: normpath can alter the meaning of a path when symbolic links are involved. I think it may be a bad idea to use it though, precisely because it may end up being a different path to the one the user asked for.

ludwigschwardt: Good point - I'm not really familiar with `normpath` and just thought it was cuter to do go

`bucket/cb` -> `bucket/cb/../cb_stream` -> `bucket/cb_stream`

instead of

`bucket/cb` -> `bucket/` -> `bucket/cb_stream`

If the `..` doesn't hurt I'll leave it in.

PR #132, opened 2018-03-20
Fix explicit override of a chunk store with file URL

bmerry:
TelstateDataStore.from_url('file:///...', store) would ignore the
provided store, and use the s3_endpoint_url from the RDB file
regardless.

The logic is changed now so that chunk_store is only overridden if it is
given as 'auto'.


bmerry:
@lmagnus FYI

PR #134, opened 2018-04-09
Decode binary pickles from HDF5 v3 files

bmerry:
This allows pickles containing binary (rather than ASCII) data to be
loaded. See SR-675.

### DIFF
--- a/katdal/h5datav3.py
+++ b/katdal/h5datav3.py
@@ -30,7 +30,7 @@ except ImportError:
 from .dataset import (DataSet, WrongVersion, BrokenFile, Subarray, SpectralWindow,
                       DEFAULT_SENSOR_PROPS, DEFAULT_VIRTUAL_SENSORS, _robust_target)
 from .sensordata import (SensorCache, RecordSensorData,
-                         H5TelstateSensorData)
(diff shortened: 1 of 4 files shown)


bmerry:
@mserylak once this lands it will also affect the beamformer HDF5 files and you may need to make an equivalent change to your scripts.

mserylak:
@bmerry thank you for the heads up! I'll have a look and adjust accordingly.

bmerry:
> There is no need to import pickle in h5toms and objtoms anymore

Fixed.

> while h5datav3 only needs it for UnpicklingError. I can't think of a quick way to get rid of the latter though.

I'll leave it in then. I was tempted to remove the cPickle import and just import pickle to get the exception, but in fact `cPickle.UnpicklingError` and `pickle.UnpicklingError` are separate classes.

By the way, you may be interested in this (although possibly it doesn't apply to cPickle)
```
In [4]: pickle.UnpicklingError?
Docstring:
This exception is raised when there is a problem unpickling an object,
such as a security violation.

Note that other exceptions may also be raised during unpickling, including
(but not necessarily limited to) AttributeError, EOFError, ImportError,
and IndexError.
```

PR #137, opened 2018-04-20
Avoid using dask advanced indexing in common cases

bmerry:
Fixes SR-1166. This makes loading all the data from a file twice as fast
and much less CPU-intensive.

### DIFF
--- a/katdal/lazy_indexer.py
+++ b/katdal/lazy_indexer.py
@@ -20,6 +20,52 @@ import numpy as np
 
 # TODO support advanced integer indexing with non-strictly increasing indices (i.e. out-of-order and duplicates)
 
+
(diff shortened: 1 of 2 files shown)


katdal/lazy_indexer.py
```diff
@@ -20,6 +20,52 @@
...
+    """
+    if not isinstance(indices, tuple):
+        indices = (indices,)
+    out = []
+    axis = 0
+    for index in indices:
+        if index is Ellipsis:
+            return indices     # Not supported
+        elif index is np.newaxis:
+            out.append(index)
```
ludwigschwardt: Do you also need to increment `axis` here? If so, I would redesign the loop to have a single `out.append(index)` at the end.

PR #138, opened 2018-04-23
Merge h5toms objtoms

bmerry:
The Github PR view may not be that useful. There are two commits that are worth reviewing separately. The first adds the merged script, and you may want to review it as a diff against either objtoms.py or h5toms.py. The second deals with removing the old code; I'm open to alternative suggestions on what we should or shouldn't carry forward.


scripts/mvftoms.py
```diff
@@ -0,0 +1,706 @@
...
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+################################################################################
+
+# Produce a CASA-compatible Measurement Set from a KAT-7 HDF5 file (versions
+# 1 and 2) or a MeerKAT Visibility Format dataset (versions 3 and 4) using
```
ludwigschwardt: I would just call all of them MVF (versions 1 to 4). No need to distinguish KAT-7.

scripts/mvftoms.py
```diff
@@ -0,0 +1,706 @@
...
+from katdal.sensordata import pickle_loads
+
+
+def main():
+    tag_to_intent = {'gaincal': 'CALIBRATE_PHASE,CALIBRATE_AMPLI',
+                     'bpcal': 'CALIBRATE_BANDPASS,CALIBRATE_FLUX',
+                     'target': 'TARGET'}
+
+    parser = optparse.OptionParser(usage="%prog [options] <filename> [<filename2>]*",
+                                   description='Convert KAT dataset(s) to MeasurementSet')
```
ludwigschwardt: KAT -> MVF

scripts/mvftoms.py
```diff
@@ -0,0 +1,706 @@
...
+    if len(args) < 1:
+        parser.print_help()
+        raise RuntimeError("Please provide one or more filenames as arguments")
+
+    if options.elevation_range and len(options.elevation_range.split(',')) < 2:
+        raise RuntimeError("You have selected elevation flagging. Please provide elevation "
+                           "limits in the form 'lowest_elevation,highest_elevation'.")
+
+    if len(args) > 1:
+        print "Concatenating multiple datasets files into single MS."
```
ludwigschwardt: `files` can be removed.

scripts/mvftoms.py
```diff
@@ -0,0 +1,706 @@
...
+
+    for win in range(len(dataset.spectral_windows)):
+        dataset.select(reset='T')
+
+        # Extract MS file per spectral window in observation file
+        print 'Extract MS for spw %d: central frequency %.2f MHz' % (win, (dataset.spectral_windows[win]).centre_freq / 1e6)
+        cen_freq = '%d' % int(dataset.spectral_windows[win].centre_freq / 1e6)
+
+        # If no output MS filename supplied, infer the output filename
+        # from the first file.
```
ludwigschwardt: file -> dataset

bmerry: Done

scripts/mvftoms.py
bmerry: Done

scripts/mvftoms.py
bmerry: Done

PR #139, opened 2018-04-24
Clear up v4 niggles

ludwigschwardt:
This is a first pass to prepare for release:

  - Improve timeouts and errors for S3
  - Restore support for NumPy 1.11 / Ubuntu 16.04
  - Speed up flag access
  - Fix relative NPY store paths
  - Update docs to refer to datasets and not files (MVF still has to happen)
  - Quack by default in mvftoms
  - Remove katsdpauth from requirements.txt for now

### DIFF
--- a/README.rst
+++ b/README.rst
@@ -21,9 +21,9 @@ Open any data set through a single function to obtain a data set object:
 This automatically determines the version and storage location of the data set.
 The versions roughly map to the various instruments::
 
-  - v1 : Fringe Finder / FF (HDF5 file)
+  - v1 : Fringe Finder (HDF5 file)
   - v2 : KAT-7 (HDF5 file)
-  - v3 : RTS / AR1 (HDF5 file)
+  - v3 : MeerKAT (HDF5 file)
   - v4 : MeerKAT (chunk store based on objects in Ceph)
 
 Multiple data sets (even of different versions) may also be concatenated
--- a/README.rst
+++ b/README.rst
@@ -1,33 +1,39 @@
 katdal
 ======
 
-This package serves as a data access library to the HDF5 files produced by
-the Fringe Finder, KAT-7 and MeerKAT data capturing systems. It uses memory
-carefully, allowing files to be inspected and partially loaded into memory.
-Data sets may be concatenated and split via a flexible selection mechanism.
-In addition, it provides a script to convert these HDF5 files to CASA
-MeasurementSets.
+This package serves as a data access library to interact with the chunk stores
+and HDF5 files produced by the MeerKAT radio telescope and its predecessors
+(KAT-7 and Fringe Finder). It uses memory carefully, allowing data sets to be
+inspected and partially loaded into memory. Data sets may be concatenated and
+split via a flexible selection mechanism. In addition, it provides a script to
+convert these data sets to CASA MeasurementSets.
 
 Quick Tutorial
 --------------
 
-Open any HDF5 file through a single function to obtain a data set object:
+Open any data set through a single function to obtain a data set object:
 
 .. code:: python
 
   import katdal
   d = katdal.open('1234567890.h5')
 
-This automatically determines whether it is a version 1 (FF), version 2
-(KAT-7) or version 3 (MeerKAT) file. Multiple files (even of different
-versions) may also be concatenated together (as long as they have the
-same dump rate):
+This automatically determines the version and storage location of the data set.
+The versions roughly map to the various instruments::
+
+  - v1 : Fringe Finder / FF (HDF5 file)
+  - v2 : KAT-7 (HDF5 file)
+  - v3 : RTS / AR1 (HDF5 file)
+  - v4 : MeerKAT (chunk store based on objects in Ceph)
+
+Multiple data sets (even of different versions) may also be concatenated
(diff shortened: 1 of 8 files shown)


README.rst
```diff
@@ -1,33 +1,39 @@
...
-This automatically determines whether it is a version 1 (FF), version 2
-(KAT-7) or version 3 (MeerKAT) file. Multiple files (even of different
-versions) may also be concatenated together (as long as they have the
-same dump rate):
+This automatically determines the version and storage location of the data set.
+The versions roughly map to the various instruments::
+
+  - v1 : Fringe Finder / FF (HDF5 file)
+  - v2 : KAT-7 (HDF5 file)
+  - v3 : RTS / AR1 (HDF5 file)
```
bmerry: Maybe just call these MeerKAT - since we're still producing v3 files, and AR1 was never a separate instrument to MeerKAT.

katdal/__init__.py
```diff
@@ -14,32 +14,40 @@
...
 
-This automatically determines whether it is a version 1 (FF) or version 2
-(KAT-7) file. Multiple files (even of different versions) may also be
-concatenated together (as long as they have the same dump rate)::
+This automatically determines the version and storage location of the data set.
+The versions roughly map to the various instruments::
+
+  - v1 : Fringe Finder / FF (HDF5 file)
+  - v2 : KAT-7 (HDF5 file)
+  - v3 : RTS / AR1 (HDF5 file)
```
bmerry: Same comment as above. Do we really need two copies of it, or can we just direct people to read the README for the overview?

katdal/chunkstore_s3.py
```diff
@@ -70,13 +70,14 @@ def __init__(self, client):
...
         error_map = {EndpointConnectionError: StoreUnavailable,
+                     ConnectionError: StoreUnavailable,
                      client.exceptions.NoSuchKey: ChunkNotFound,
                      client.exceptions.NoSuchBucket: ChunkNotFound}
         super(S3ChunkStore, self).__init__(error_map)
         self.client = client
 
     @classmethod
-    def from_url(cls, url, **kwargs):
+    def from_url(cls, url, timeout=5, **kwargs):
```
bmerry: 5 is possibly a little low. I don't know about S3, but I seem to recall other places where I've had to increase 5 second connection timeouts to make things reliable.

katdal/datasources.py
```diff
@@ -286,6 +286,8 @@ def from_url(cls, url, chunk_store='auto'):
             # Look for adjacent data directory (presumably containing NPY files)
             if chunk_store == 'auto':
                 store_path = os.path.dirname(os.path.dirname(url_parts.path))
+                if not store_path:
+                    store_path = os.path.curdir
```
bmerry: I think this still won't fix it if I'm actually in the directory with the .rdb file and just specify its base filename. I would suggest either
```python
store_path = os.path.join(os.path.dirname(url_parts.path), '..', '..')
```
or
```python
store_path = os.path.dirname(os.path.dirname(os.path.abspath(url_parts.path)))
```
(both untested)

katdal/visdatav4.py
```diff
@@ -116,7 +117,7 @@ def __init__(self, source, ref_ant='', time_offset=0.0, **kwargs):
         half_dump = 0.5 * self.dump_period
         self.start_time = katpoint.Timestamp(source.timestamps[0] - half_dump)
         self.end_time = katpoint.Timestamp(source.timestamps[-1] + half_dump)
-        self._time_keep = np.full(num_dumps, True)
+        self._time_keep = np.full(num_dumps, True, dtype=bool)
```
bmerry: Maybe `np.bool_` to be a bit more precise?

katdal/visdatav4.py
```diff
@@ -20,6 +20,7 @@
 
 import numpy as np
 import katpoint
+import dask.array as da
```
bmerry: Looks like you're making dask a requirement now. Presumably it should be added to setup.py, and there is at least one place (in lazy_indexer.py) where you have a TODO to lift out a function-scoped import to the module level once katdal depends on dask.

README.rst
ludwigschwardt: Sure. These edits also predated MVF, so I'll update accordingly.

katdal/__init__.py
ludwigschwardt: I want to be able to do `katdal?` in IPython. The ideal would be to load README into the docstring on demand.

katdal/chunkstore_s3.py
ludwigschwardt: The default has an exponential backoff that takes *ages* to time out. I could be persuaded to make it 10 seconds, but limited testing never needed more than a second to connect.

katdal/datasources.py
ludwigschwardt: I like the latter.

katdal/visdatav4.py
ludwigschwardt: Sure.

katdal/visdatav4.py
ludwigschwardt: It's still optional since the whole v4 class is optional. But I should probably take my head out of the sand at some stage :-)

PR #140, opened 2018-04-24
Optimisations to mvftoms.py

bmerry:
There are a number of changes - refer to individual commits for details. Note that UVW coordinates will change very slightly.

This does introduce a hard dependency on dask, which I'm not sure how you want to handle, and which should be addressed before merge. I'm not sure if you're already planning to make it a hard dependency as part of the v4 release cleanup.

### DIFF
--- a/scripts/mvftoms.py
+++ b/scripts/mvftoms.py
@@ -28,12 +28,40 @@ import optparse
 import time
 
 import numpy as np
+import dask.array as da
 
 import katpoint
 import katdal
 from katdal import averager
 from katdal import ms_extra
 from katdal.sensordata import pickle_loads
+
+
+def load(dataset, indices, vis, weights, flags):
+    """Load data from one a lazy indexer into existing storage.
+
+    This is optimised for the MVF v4 case where we can use dask directly
+    to eliminate one copy, and so load vis, flags and weights in parallel.
+    In older formats it causes an extra copy.
+
+    Parameters
+    ----------
+    dataset : :class:`katdal.DataSet`
+        Input dataset, possibly with an existing selection
+    indices : tuple
+        Index expression for subsetting the dataset
+    vis, weights, flags : array-like
+        Outputs, which must have the correct shape and type
+    """
+    if hasattr(dataset.vis, 'dataset'):
+        da.store([dataset.vis.dataset[indices],
+                  dataset.weights.dataset[indices],
+                  dataset.flags.dataset[indices]],
+                 [vis, weights, flags], lock=False)
+    else:
+        vis[:] = dataset.vis[indices]
+        weights[:] = dataset.weights[indices]
+        flags[:] = dataset.flags[indices]
 
 
 def main():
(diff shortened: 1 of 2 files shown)


scripts/mvftoms.py
```diff
@@ -433,11 +433,14 @@ def _cp_index(a1, a2, pol):
 
                 # Select correlator products
                 # cp_index could be used above when the LazyIndexer
-                # supports advanced integer indices
-                vis_data = scan_data[:, :, cp_info.cp_index]
-
-                weight_data = scan_weight_data[:, :, cp_info.cp_index]
-                flag_data = scan_flag_data[:, :, cp_info.cp_index]
+                # supports advanced integer indices. Note that np.take is
+                # equivalent to but much faster than fancy indexing.
```
ludwigschwardt: Maybe it's because we are only selecting on one axis and `np.take` can somehow "take" advantage of that.

scripts/mvftoms.py
```diff
@@ -36,6 +37,33 @@
 from katdal.sensordata import pickle_loads
 
 
+def load(dataset, indices, vis, weights, flags):
+    """Load data from one a lazy indexer into existing storage.
```
ludwigschwardt: 'a'? The only thing to fix...

scripts/mvftoms.py
```diff
@@ -429,20 +472,19 @@ def _cp_index(a1, a2, pol):
...
-                scan_flag_data = dataset.flags[ltime:utime, :, :]
+                load(dataset, np.s_[ltime:utime, :, :],
+                     scan_vis_data, scan_weight_data, scan_flag_data)
 
                 # Select correlator products
                 # cp_index could be used above when the LazyIndexer
                 # supports advanced integer indices. Note that np.take is
                 # equivalent to but much faster than fancy indexing.
-                vis_data = np.take(scan_data, cp_info.cp_index, axis=2)
-                del scan_data
```
ludwigschwardt: I was wondering about these `del`s - glad you removed them.

scripts/mvftoms.py
```diff
@@ -36,6 +37,33 @@
...
+    ----------
+    dataset : :class:`katdal.DataSet`
+        Input dataset, possibly with an existing selection
+    indices : tuple
+        Index expression for subsetting the dataset
+    vis, weights, flags : array-like
+        Outputs, which must have the correct shape and type
+    """
+    if hasattr(dataset.vis, 'dataset'):
+        da.store([dataset.vis.dataset[indices],
```
ludwigschwardt: Does this do a `compute()` as well?

scripts/mvftoms.py
bmerry: Quite likely.

scripts/mvftoms.py
bmerry: Fixed

scripts/mvftoms.py
bmerry: No, because `compute()` generates new memory. This evaluates the graph but copies the chunks directly into the existing storage without assembling them into contiguous arrays.

PR #142, opened 2018-04-26
Further mvftoms.py optimisations

bmerry:
The big difference in user experience is that it uses numba.

The FLAG_CATEGORY column is also changed to contain the flags instead of zeros.

If reviewing a commit at a time, you may want to look at the first two commits as a unit, since the second rewrites a lot of the first.

### DIFF
--- a/scripts/mvftoms.py
+++ b/scripts/mvftoms.py
@@ -65,7 +65,7 @@ def load(dataset, indices, vis, weights, flags):
         flags[:] = dataset.flags[indices]
 
 
-@numba.jit(nopython=True, cache=True)
+@numba.jit(nopython=True, cache=True, parallel=True)
 def permute_baselines(in_vis, in_weights, in_flags, cp_index, out_vis, out_weights, out_flags):
     """Reorganise baselines and axis order.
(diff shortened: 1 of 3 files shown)


scripts/mvftoms.py
```diff
@@ -64,6 +65,48 @@ def load(dataset, indices, vis, weights, flags):
...
+    n_time, n_bls, n_chans, n_pols = out_vis.shape
+    bstep = 256
+    bblocks = (n_bls + bstep - 1) // bstep
+    for t in range(n_time):
+        for bblock in numba.prange(bblocks):
+            bstart = bblock * bstep
+            bstop = min(n_bls, bstart + bstep)
+            for c in range(n_chans):
+                for b in range(bstart, bstop):
+                    for p in range(out_vis.shape[3]):
```
ludwigschwardt: Would it help to swop the `for b` and `for p`? The baselines are separated into groups with the same polarisation.

bmerry: As discussed verbally: worth a try, but it looks like it's a bit slower.

PR #145, opened 2018-05-04
Fix mvftoms for v3 files

bmerry:
My detection of the dataset being dask-like was flawed and was
incorrectly trying to use da.store on HDF5 data.


bmerry:
@ludwigschwardt I'm going to merge this directly because it's a pretty serious bug and I don't want to leave it until you're back from leave, and the fix is simple. If you have any changes I'm happy to address them when you're back.

PR #148, opened 2018-05-09
Require telstate and dask (and bye-bye blank.ms)

ludwigschwardt:
`katdal` now requires `katsdptelstate[rdb]` and `dask[array]`. The former is about to be released to PyPI and will be needed by external users of katdal master after this PR.

Also remove good old `blank.ms`.


setup.py
```diff
@@ -53,16 +53,16 @@
...
+      install_requires=['numpy', 'katpoint', 'h5py>=2.3',
+                        'katsdptelstate[rdb]', 'dask[array]'],
       extras_require={
         'ms': ['python-casacore >= 2.2.1', 'numba'],
         's3': ['botocore'],
         # rados is not in PyPI but available as Debian package python-rados
         'rados': ['rados'],
-        # Only available throught github currently
+        # katsdpauth is currently only available via GitHub
         'auth': ['katsdpauth'],
```
bmerry: Not directly related to this PR, but this section is only 2-char indented, should be 4.

requirements.txt
```diff
@@ -1,15 +1,21 @@
...
 python-dateutil
+python-lzf==0.2.4
+rdbtools==0.1.12
+redis
 six
 toolz==0.9.0
-git+ssh://git@github.com/ska-sa/katpoint#egg=katpoint
+
+git+ssh://git@github.com/ska-sa/katpoint
+git+ssh://git@github.com/ska-sa/katsdptelstate
```
bmerry: I'm never quite clear on when pip needs the egg= - do you know?

ludwigschwardt: I'm trying to figure this out now...

This feature seems to have been more useful in the `easy_install` days. Its only relevant purpose seems to be the following (quoting @carljm from [here](https://groups.google.com/forum/#!topic/python-virtualenv/-2CZMRFixJg)):

```
In pip AFAIK the main purpose of the #egg= fragment is to help pip sort out dependencies; so if you have another package that depends on a package you're installing editable (or from a direct URL), using the fragment allows pip to know earlier in the process that you've got that dependency covered and it doesn't need to go finding it. 
```

So I guess I'll put it back.

setup.py
ludwigschwardt: Fixed.

PR #149, opened 2018-05-10
Write MS in parallel with loading+permuting

bmerry:
In mvftoms, the measurement set writing is split off into a separate multiprocessing subprocess, which communicates with the main process via shared memory. Since the MS writing is single-threaded, it was a bottleneck, but with this change it should theoretically be possible to use all the cores all the time (in practice I think it gets I/O latency bound).

This does increase memory usage: 4 dumps-worth of data are needed for the circular buffer, which in worst case could be around 14GB.


katdal/ms_async.py
```diff
@@ -0,0 +1,194 @@
+################################################################################
+# Copyright (c) 2011-2018, National Research Foundation (Square Kilometre Array)
```
ludwigschwardt: As something very minor, I usually pick the first year as the birth year of the specific module (i.e. `(c) 2018`).

bmerry: When I pressed copy that was my intention, but somehow by the time I pasted I'd forgotten. Fixed now.

PR #152, opened 2018-05-14
Fix S3 unit tests again

ludwigschwardt:
The new Docker version on sdp-jenkins seems to map 'localhost' to an IPv6 address, even when IPv6 is disabled. This prevents connection to the FakeS3 server and fails the `S3ChunkStore` tests.

Rather use the explicit '127.0.0.1'. Also suppress the FakeS3 stdout for cleaner test reports when done on the command line.


ludwigschwardt:
Seems that Jenkins is happy again, thanks for that!

PR #153, opened 2018-05-16
Flag and zero missing chunks

ludwigschwardt:
All missing chunks (i.e. chunks raising `ChunkNotFound` in the chunk store) are now replaced by zeros of the appropriate shape and dtype. In addition, missing chunks in the `correlator_data`, `weights` and `weights_channel` (but not `flags`...) arrays will get flagged as `data_lost`.

In the process, add unit tests for `ChunkStoreVisFlagsWeights`, including the ability to make fake datasets in the chunk store.

The last niggle is to replace the explicit flag value of 8 with `data_lost` but that will come in a separate PR that will rearrange all the flag info.

### DIFF
--- a/katdal/datasources.py
+++ b/katdal/datasources.py
@@ -84,17 +84,17 @@ class VisFlagsWeights(object):
         return self.vis.shape
 
 
-def _has_chunk_to_flags(has_chunk, block_id=None, full_chunks=None):
+def _has_chunk_to_flags(has_chunk, block_id, full_chunks):
     """Turn a has_chunk bool into chunk of flags with correct data_lost bit."""
     shape = tuple(chk[idx] for chk, idx in zip(full_chunks, block_id))
     return np.full(shape, 0 if has_chunk else 8, dtype=np.uint8)
 
 
 def _multi_or_3d(*args):
-    """Do bitwise 'or' of more than two 3-D arrays."""
+    """Do bitwise 'or' of two or more 3-D arrays (without modifying them)."""
     args = np.atleast_3d(*args)
-    out = args[0]
-    for arg in args[1:]:
+    out = args[0] | args[1]
+    for arg in args[2:]:
         out |= arg
     return out
 
--- a/katdal/datasources.py
+++ b/katdal/datasources.py
@@ -90,8 +90,9 @@ def _has_chunk_to_flags(has_chunk, block_id=None, full_chunks=None):
     return np.full(shape, 0 if has_chunk else 8, dtype=np.uint8)
 
 
-def _multi_or(*args):
-    """Do bitwise 'or' of more than two arrays."""
+def _multi_or_3d(*args):
+    """Do bitwise 'or' of more than two 3-D arrays."""
+    args = np.atleast_3d(*args)
     out = args[0]
     for arg in args[1:]:
         out |= arg
@@ -124,17 +125,14 @@ class ChunkStoreVisFlagsWeights(VisFlagsWeights):
                                         token='missing-chunks-' + array_name,
                                         chunks=info['chunks'], dtype=np.uint8,
                                         full_chunks=info['chunks'])
-            # The weights_channel array is only time x freq, so extend to 3 dims
-            if chunks_lost.ndim == 2:
-                chunks_lost = chunks_lost[..., np.newaxis]
             # If a flag chunk is missing but not the corresponding vis/weights,
             # the data_lost bit will *not* be set (all flags cleared, actually)
             if array != 'flags':
                 extra_flags.append(chunks_lost)
-                extra_flags.append('ijk')
+                extra_flags.append('ijk'[:chunks_lost.ndim])
         vis = darray['correlator_data']
         # Combine original L0 flags with extras (missing chunks per array)
-        flags = da.atop(_multi_or, 'ijk', darray['flags'], 'ijk',
+        flags = da.atop(_multi_or_3d, 'ijk', darray['flags'], 'ijk',
                         *extra_flags, token=store.join(base_name, 'flags_raw'),
                         dtype=np.uint8)
         # Combine low-resolution weights and high-resolution weights_channel
--- a/katdal/datasources.py
+++ b/katdal/datasources.py
@@ -97,15 +97,15 @@ class ChunkStoreVisFlagsWeights(VisFlagsWeights):
     """
     def __init__(self, store, base_name, chunk_info):
         self.store = store
-        da = {}
+        darray = {}
         for array, info in chunk_info.iteritems():
             array_name = store.join(base_name, array)
-            da[array] = store.get_dask_array(array_name, info['chunks'],
-                                             info['dtype'])
-        vis = da['correlator_data']
-        flags = da['flags']
+            darray[array] = store.get_dask_array(array_name, info['chunks'],
+                                                 info['dtype'])
+        vis = darray['correlator_data']
+        flags = darray['flags']
         # Combine low-resolution weights and high-resolution weights_channel
-        weights = da['weights'] * da['weights_channel'][..., np.newaxis]
+        weights = darray['weights'] * darray['weights_channel'][..., np.newaxis]
         VisFlagsWeights.__init__(self, vis, flags, weights, base_name)
 
 
--- a/katdal/datasources.py
+++ b/katdal/datasources.py
@@ -22,6 +22,7 @@ import logging
 
 import katsdptelstate
 import numpy as np
+import dask.array as da
 
 from .sensordata import TelstateSensorData
 from .chunkstore_s3 import S3ChunkStore
@@ -83,6 +84,20 @@ class VisFlagsWeights(object):
         return self.vis.shape
 
 
+def _has_chunk_to_flags(has_chunk, block_id=None, full_chunks=None):
+    """Turn a has_chunk bool into chunk of flags with correct data_lost bit."""
+    shape = tuple(chk[idx] for chk, idx in zip(full_chunks, block_id))
+    return np.full(shape, 0 if has_chunk else 8, dtype=np.uint8)
+
+
+def _multi_or(*args):
+    """Do bitwise 'or' of more than two arrays."""
+    out = args[0]
+    for arg in args[1:]:
+        out |= arg
+    return out
+
+
 class ChunkStoreVisFlagsWeights(VisFlagsWeights):
     """Correlator data stored in a chunk store.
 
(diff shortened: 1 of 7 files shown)


katdal/test/test_datasources.py
```diff
@@ -0,0 +1,113 @@
...
+
+def ramp(shape, offset=0.0, slope=1.0, dtype=np.float_):
+    """Generate a multidimensional ramp of values of the given dtype."""
+    x = offset + slope * np.arange(np.prod(shape), dtype=np.float_)
+    return x.astype(dtype).reshape(shape)
+
+
+def to_dask_array(x):
+    """Turn ndarray `x` into dask array with the standard vis-like chunking."""
+    n_corrprods = x.shape[2] if x.ndim >= 3 else x.shape[1] / 8
```
bmerry: I'm an advocate of always using `//` for integer division so that you don't get surprised when moving to Python 3.

katdal/test/test_datasources.py
```diff
@@ -0,0 +1,113 @@
...
+    chunk_size = 4 * n_corrprods * 8
```
bmerry: Just to check I understand this: 8 is sizeof(complex64), and the factor of 4 is the nominal number of channels per chunk?

katdal/test/test_datasources.py
```diff
@@ -0,0 +1,113 @@
...
+    data = {'correlator_data': ramp(shape, dtype=np.float32) * (1 - 1j),
+            'flags': np.zeros(shape, dtype=np.uint8),
+            'weights': ramp(shape, slope=256. / np.prod(shape), dtype=np.uint8),
+            'weights_channel': ramp(shape[:-1], dtype=np.float32)}
+    ddata = {k: to_dask_array(array) for k, array in data.items()}
+    chunk_info = {k: {'chunks': darray.chunks, 'dtype': darray.dtype,
+                      'shape': darray.shape} for k, darray in ddata.items()}
+    push = [store.put_dask_array(store.join(base_name, k), darray)
+            for k, darray in ddata.items()]
+    da.compute(push)
```
bmerry: Should this be `*push`?

katdal/test/test_datasources.py
```diff
@@ -0,0 +1,113 @@
...
+        store = NpyFileChunkStore(self.tempdir)
+        base_name = 'cb2'
+        shape = (10, 64, 30)
+        data, chunk_info = put_fake_dataset(store, base_name, shape)
+        # Delete a random chunk in each array of the dataset
+        missing_chunks = {}
+        for array, info in chunk_info.items():
+            array_name = store.join(base_name, array)
+            slices = da.core.slices_from_chunks(info['chunks'])
+            index = np.random.randint(len(slices))
```
bmerry: Use an np.RandomState with a fixed seed so that the test behaves the same every time. Or alternatively, hand-code the missing pieces. You probably want to avoid two of the arrays having the same piece go missing, since that would weaken coverage of the test on data_lost.

bmerry: By the way, I think you can just to `random.choice(slices)` to directly pick a random slice rather than first computing a random slice index and then indexing.

katdal/test/test_datasources.py
```diff
@@ -0,0 +1,113 @@
...
+            missing_chunks[array] = slices[index]
+            chunk_name, shape = store.chunk_metadata(array_name, slices[index])
+            os.remove(os.path.join(store.path, chunk_name) + '.npy')
+        vfw = ChunkStoreVisFlagsWeights(store, base_name, chunk_info)
+        # Check that missing chunks have been replaced by zeroes
```
bmerry: Should maybe check that nothing else has been replaced by zeros; and perhaps offset the ramps so that none of the data is itself zero.

katdal/test/test_datasources.py
```diff
@@ -0,0 +1,113 @@
...
+    chunk_size = 4 * n_corrprods * 8
+    chunks = generate_chunks(x.shape, x.dtype, chunk_size,
+                             dims_to_split=(0, 1), power_of_two=True)
+    return da.from_array(x, chunks)
+
+
+def put_fake_dataset(store, base_name, shape):
+    """Write a fake dataset into the chunk store."""
+    data = {'correlator_data': ramp(shape, dtype=np.float32) * (1 - 1j),
+            'flags': np.zeros(shape, dtype=np.uint8),
```
bmerry: Maybe put some more interesting values in here to check that they carry through to the output.

katdal/test/test_datasources.py
```diff
@@ -0,0 +1,113 @@
...
+        # Check that missing chunks have been replaced by zeroes
+        assert_array_equal(vfw.vis[missing_chunks['correlator_data']], 0.)
+        assert_array_equal(vfw.weights[missing_chunks['weights']], 0.)
+        assert_array_equal(vfw.weights[missing_chunks['weights_channel']], 0.)
+        # Check that (only) missing chunks have been flagged as 'data lost'
+        expected = np.zeros_like(vfw.flags)
+        expected[missing_chunks['correlator_data']] |= 8
+        expected[missing_chunks['weights']] |= 8
+        expected[missing_chunks['weights_channel']] |= 8
+        assert_array_equal(vfw.flags & 8, expected)
```
bmerry: Should check all the bits, not just the data_lost bit, to ensure that existing flags are preserved where no data is lost.

katdal/datasources.py
```diff
@@ -83,6 +84,21 @@ def shape(self):
         return self.vis.shape
 
 
+def _has_chunk_to_flags(has_chunk, block_id=None, full_chunks=None):
```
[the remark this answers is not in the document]
ludwigschwardt: These defaults should never be used and I hope things crash if they are. The function is only meant to be used together with `da.map_blocks`, which passes in the expected kwargs from its own kwargs.

I'm following the example straight from the `da.map_blocks` docstring which explains the special behaviour of the `block_id` keyword, if given. That default of `None` is also invalid but will never occur.

...

So I've checked: I can safely delete the defaults, so that's what I'll do.

katdal/datasources.py
```diff
@@ -83,6 +84,21 @@ def shape(self):
...
+    """Turn a has_chunk bool into chunk of flags with correct data_lost bit."""
+    shape = tuple(chk[idx] for chk, idx in zip(full_chunks, block_id))
+    return np.full(shape, 0 if has_chunk else 8, dtype=np.uint8)
+
+
+def _multi_or_3d(*args):
+    """Do bitwise 'or' of more than two 3-D arrays."""
```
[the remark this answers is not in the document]
ludwigschwardt: Fixed.

katdal/datasources.py
```diff
@@ -83,6 +84,21 @@ def shape(self):
...
+    args = np.atleast_3d(*args)
+    out = args[0]
+    for arg in args[1:]:
+        out |= arg
```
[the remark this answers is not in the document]
ludwigschwardt: Fixed.

katdal/datasources.py
```diff
@@ -97,15 +113,30 @@ class ChunkStoreVisFlagsWeights(VisFlagsWeights):
...
+            chunk_args = (array_name, info['chunks'], info['dtype'])
+            darray[array] = store.get_dask_array(*chunk_args)
+            # Find all missing chunks in array and convert to 'data_lost' flags
+            has_array = store.has_dask_array(*chunk_args)
+            chunks_lost = da.map_blocks(_has_chunk_to_flags, has_array,
+                                        token='missing-chunks-' + array_name,
+                                        chunks=info['chunks'], dtype=np.uint8,
+                                        full_chunks=info['chunks'])
+            # If a flag chunk is missing but not the corresponding vis/weights,
+            # the data_lost bit will *not* be set (all flags cleared, actually)
```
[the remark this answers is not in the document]
ludwigschwardt: This was @sratcliffe's suggestion, the thinking being that egregiously bad data will eventually get flagged again somewhere down the line. There is also the L1 flags, essentially enhanced copies of the L0 flags, that will be folded in soon, thereby mitigating missing flag chunks.

The conclusion of our discussion is that I'll eventually fold in L1 and if both are missing, we'll flag it `data_lost` (since this will hopefully happen very infrequently).

katdal/datasources.py
```diff
@@ -97,15 +113,30 @@ class ChunkStoreVisFlagsWeights(VisFlagsWeights):
...
+            if array != 'flags':
+                extra_flags.append(chunks_lost)
+                extra_flags.append('ijk'[:chunks_lost.ndim])
+        vis = darray['correlator_data']
+        # Combine original L0 flags with extras (missing chunks per array)
+        flags = da.atop(_multi_or_3d, 'ijk', darray['flags'], 'ijk',
+                        *extra_flags, token=store.join(base_name, 'flags_raw'),
```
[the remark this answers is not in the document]
ludwigschwardt: For me the cooked flags are the bools in `katdal.DataSet` (after the selected flag bits are ORed together). The uint8 version is 'raw'.

My thinking on the token was that this was a data product like `correlator_data` as opposed to an intermediate step. I can see that it could confuse someone.

katdal/test/test_datasources.py
ludwigschwardt: Fixed. Forgot about that.

katdal/test/test_datasources.py
ludwigschwardt: Yes, this is meant to elicit an 'interesting' and realistic chunking, where the chunks differ for most of the basic arrays.

katdal/test/test_datasources.py
ludwigschwardt: It doesn't have to be: `da.compute` traverses Python builtin collections by default looking for dask objects. I implemented your suggestion to help dask along ever so slightly :-)

katdal/test/test_datasources.py
ludwigschwardt: Done.

I knew there was a better function but could not remember the name - thanks!

ludwigschwardt: Unfortunately `random.choice` doesn't work as the list of tuples of slices are interpreted as an N-D array with N > 1. Back to `randint` (but kept the `RandomState` idea).

katdal/test/test_datasources.py
ludwigschwardt: Done.

katdal/test/test_datasources.py
ludwigschwardt: Done (same for vis and weights).

katdal/test/test_datasources.py
ludwigschwardt: Changed zeros() to ones(). At least it caught a bug :-)

katdal/datasources.py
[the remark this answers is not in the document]
bmerry: flags_raw makes sense to me now. Carry on.

katdal/datasources.py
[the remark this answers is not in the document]
ludwigschwardt: The latest thinking is: use L1 flags in preference to L0 flags if available, instead of trying to combine them somehow.

katdal/test/test_datasources.py
ludwigschwardt: Switched back to `random.choice`, not `np.random.choice` :-)

mauch:
This now seems to flag all of the data as 'data_lost'. I've tried on two different datasets with similar results:
In [18]: t=katdal.open('1524483838_sdp_l0.rdb')
In [20]: np.all(t.flags)
Out[20]: True

ludwigschwardt:
This issue is addressed in #154.

PR #155, opened 2018-05-21
Optimise has_dask_array

ludwigschwardt:
Since the flags now call `has_dask_array` on the rest of the arrays
in the dataset, it is time to optimise it. This feature can also be
used to check which chunks are present upon opening a dataset.

Some chunk store implementations allow an efficient enumeration of
all chunks associated with an array. We are only interested in the
presence or absence of a chunk and nothing more. For example, the
`NpyFileChunkStore` can `os.listdir` the array subdirectory to get
a list of all the relevant NPY files, while the `S3ChunkStore` can
send a `list_objects_v2` request with the appropriate bucket and prefix
to get a list of associated keys (via pagination).

If this efficient `list_chunk_ids` method is available, use it within
`has_dask_array`, otherwise fall back to a series of `has_chunk` calls.
Introduce the `chunk_id_str` method to simplify generation of chunk
identifier strings.

On a 2-hour dataset (1523731706) of size 250 GB served via S3 the vis
`has_dask_array` call duration drops from 90 seconds to about 13 seconds
(or 6 seconds if `has_array` is used).

### DIFF
--- a/katdal/chunkstore.py
+++ b/katdal/chunkstore.py
@@ -104,13 +104,11 @@ def generate_chunks(shape, dtype, max_chunk_size, dims_to_split=None,
 
 def _add_offset_to_slices(func, offset):
     """Modify chunk get/put/has to add an offset to its `slices` parameter."""
-    offset_slices = tuple(slice(start, None) for start in offset)
-
     def func_with_offset(array_name, slices, *args, **kwargs):
         """Shift `slices` to start at `offset`."""
-        slices = da.core.fuse_slice(offset_slices, slices)
-        return func(array_name, slices, *args, **kwargs)
-
+        offset_slices = tuple(slice(s.start + i, s.stop + i)
+                              for (s, i) in zip(slices, offset))
+        return func(array_name, offset_slices, *args, **kwargs)
     return func_with_offset
 
 
@@ -490,8 +488,9 @@ class ChunkStore(object):
         # Turn chunks + offset into list of expected chunk ID strings
         slices = da.core.slices_from_chunks(chunks)
         if offset:
-            offset_slices = tuple(slice(start, None) for start in offset)
-            slices = [da.core.fuse_slice(offset_slices, s) for s in slices]
+            slices = [tuple(slice(ss.start + i, ss.stop + i)
+                            for (ss, i) in zip(s, offset))
+                      for s in slices]
         chunk_ids = [self.chunk_id_str(s) for s in slices]
         # Look up expected IDs in set of actual IDs in store
         success = np.array([cid in store_ids for cid in chunk_ids])
--- a/katdal/chunkstore.py
+++ b/katdal/chunkstore.py
@@ -287,6 +287,13 @@ class ChunkStore(object):
         return name.split(cls.NAME_SEP, maxsplit)
 
     @classmethod
+    def chunk_id_str(cls, slices):
+        """Chunk identifier in string form (e.g. '00012_01024_00000')."""
+        index = [s.start for s in slices]
+        return '_'.join(["{:0{width}d}".format(i, width=cls.NAME_INDEX_WIDTH)
+                         for i in index])
+
+    @classmethod
     def chunk_metadata(cls, array_name, slices, chunk=None, dtype=None):
         """Turn array name and chunk identifier into chunk name and shape.
 
@@ -330,10 +337,7 @@ class ChunkStore(object):
             raise BadChunk('Array {!r}: chunk ID {} contains non-unit strides'
                            .format(array_name, slices))
         # Construct chunk name from array_name + slices
-        index = [s.start for s in slices]
-        idxstr = '_'.join(["{:0{width}d}".format(i, width=cls.NAME_INDEX_WIDTH)
-                           for i in index])
-        chunk_name = cls.join(array_name, idxstr)
+        chunk_name = cls.join(array_name, cls.chunk_id_str(slices))
         if chunk is not None and chunk.shape != shape:
             raise BadChunk('Chunk {!r}: shape {} implied by slices does not '
                            'match actual shape {}'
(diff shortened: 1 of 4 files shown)


ludwigschwardt:
Warning: the previous commit refactored the original one quite a bit, maybe read the PR as a unit...

katdal/chunkstore.py
```diff
@@ -286,6 +286,13 @@ def split(cls, name, maxsplit=-1):
         """Split chunk name into components based on supported separator."""
         return name.split(cls.NAME_SEP, maxsplit)
 
+    @classmethod
+    def chunk_id_str(cls, slices):
+        """Chunk identifier in string form (e.g. '00012_01024_00000')."""
+        index = [s.start for s in slices]
+        return '_'.join(["{:0{width}d}".format(i, width=cls.NAME_INDEX_WIDTH)
+                         for i in index])
```
bmerry: Doesn't really seem necessary to create the `index` intermediate list - it could all be done in one line comprehension.

katdal/chunkstore.py
```diff
@@ -437,6 +441,58 @@ def put_dask_array(self, array_name, array, offset=()):
...
+
+        Parameters
+        ----------
+        array_name : string
+            Identifier of array in chunk store
+
+        Returns
+        -------
+        chunk_ids : list of string
+            List of chunk identifier strings (e.g. '00012_01024_00000')
```
bmerry: Maybe document the `NotImplementedError` case. By the way, have you thought about doing `return NotImplemented`, which I think is what Python uses for unimplemented operator overloads? It would probably take more code though (since you'd have to explicitly handle it in has_array, rather than letting an exception bubble out).

katdal/chunkstore.py
```diff
@@ -437,6 +441,58 @@ def put_dask_array(self, array_name, array, offset=()):
...
+            If the underlying store does not have an efficient implementation
+        """
+        # Obtain ID strings of all chunks in store associated with array_name
+        # This might not be implemented by underlying store
+        store_ids = set(self.list_chunk_ids(array_name))
+        # Turn chunks + offset into list of expected chunk ID strings
+        slices = da.core.slices_from_chunks(chunks)
+        if offset:
+            offset_slices = tuple(slice(start, None) for start in offset)
+            slices = [da.core.fuse_slice(offset_slices, s) for s in slices]
```
bmerry: How much does this add to runtime? fuse_slice seems like a rather large hammer for this, but if it's not slowing things down then it's fine.

katdal/chunkstore_npy.py
```diff
@@ -93,6 +93,13 @@ def has_chunk(self, array_name, slices, dtype):
         filename = os.path.join(self.path, chunk_name) + '.npy'
         return os.path.exists(filename)
 
+    def list_chunk_ids(self, array_name):
+        """See the docstring of :meth:`ChunkStore.list_chunk_ids`."""
+        array_dir = os.path.join(self.path, array_name)
+        # Strip the .npy extension to get the chunk ID string
+        return [fn[:-4] for fn in os.listdir(array_dir)]
```
bmerry: This can go wrong if there happens to be other stuff in the directory apart from the .npy files. Maybe add `if fn.endswith('.npy')` to the list comprehension.

katdal/chunkstore.py
```diff
@@ -437,6 +441,58 @@ def put_dask_array(self, array_name, array, offset=()):
...
+            Identifier of array in chunk store
+
+        Returns
+        -------
+        chunk_ids : list of string
+            List of chunk identifier strings (e.g. '00012_01024_00000')
+        """
+        raise NotImplementedError
+
+    def has_array(self, array_name, chunks, offset=()):
```
bmerry: Should this maybe be a private function, with `has_dask_array` being the public, will-always-work frontend? Otherwise I can easily see someone accidentally using this version without being fully aware that they have to handle the not implemented case.

katdal/chunkstore_s3.py
```diff
@@ -172,6 +172,17 @@ def has_chunk(self, array_name, slices, dtype):
...
 
+    def list_chunk_ids(self, array_name):
+        """See the docstring of :meth:`ChunkStore.list_chunk_ids`."""
+        bucket, prefix = self.split(array_name, 1)
+        paginator = self.client.get_paginator('list_objects')
+        page_iter = paginator.paginate(Bucket=bucket, Prefix=prefix,
+                                       PaginationConfig={'PageSize': 10000})
+        keys = [item['Key'] for page in page_iter for item in page['Contents']]
+        # Strip the array name and .npy extension to get the chunk ID string
+        return [key[len(prefix) + 1:-4] for key in keys]
```
bmerry: As for the Npy store, maybe check that the filename actually conforms to the scheme and it isn't some random junk that got added.

katdal/test/test_chunkstore.py
```diff
@@ -241,3 +249,14 @@ def test_dask_array_put_parts_get_whole(self):
         # Now store the last quarter and check that complete array is correct
         self.put_dask_array('big_y2', np.s_[3:8, 30:60, 0:2])
         self.get_dask_array('big_y2')
+
+    def test_list_of_chunk_ids(self):
```
bmerry: Remove `of` so that test name matches the method being tested.

katdal/chunkstore.py
ludwigschwardt: Sure, that would be my first instinct too, but I'm trying to be gentler these days to other readers of the code :-)

In this particular case, I can buy your arguments - reverting to old habits...

katdal/chunkstore.py
ludwigschwardt: I saw in tab completion that there are two NotImplemented's and wondered about it...

The return version seems to be specialised to the operator case where the comparison can happen thousands of times, and result in more code as you mention.

I flirted with the idea of `has_array` falling back to `has_dask_array().compute()` if `list_chunk_ids` is not there, but then it all gets quite cyclical.

katdal/chunkstore.py
ludwigschwardt: I don't know. The offsets are not used much - mostly in `vis_writer` where each heap is chunked up separately and written into the larger array. Oh wait, you ditched `put_dask_array` :-)

I see how you did this in `vis_writer` - I'll pilfer that then...

katdal/chunkstore_npy.py
ludwigschwardt: OK, will add this test.

I was trying not to do too much (in the spirit of removing checks from has_chunk), since this list can also be pretty large (million elements?) and 99.999999% of the time that test will not be needed.

katdal/chunkstore.py
ludwigschwardt: I intend to use this in katdal one of these days, to verify the presence of all the chunks in the dataset. I was upset when I found out that a 10000-element `da.from_array`-based `dask.Array` can still take 7 seconds to express, so for some functionality I'd rather call `has_array`.

The main use of `has_dask_array` is when building up the flags dask array. Eventually the dataset selection will zoom in on a specific set of chunks to compute, and `da.map_blocks` (coupled with dask arrays for the flag components) simplifies this a lot.

The not implemented case is rados (pfft) and dict. The latter could still override `has_array` instead of `list_chunk_ids` to get efficiency.

katdal/chunkstore_s3.py
ludwigschwardt: Done half-heartedly.

katdal/test/test_chunkstore.py
ludwigschwardt: Good idea.

ludwigschwardt:
On the code path test coverage:

In `ChunkStoreTestBase.has_dask_array` in `test_chunk_store.py` I always try to test `has_dask_array` as well as `has_array`, which should cover both cases.

PR #159, opened 2018-05-30
Enforce timeout when constructing S3 chunk store

ludwigschwardt:
On initialisation of S3ChunkStore a connection is made to the S3
endpoint URL to check that it is up. While useful, this can lead to
interminable delays while doing DNS lookups for `archive-gw-1.kat.ac.za`
(at least from my admittedly pathological home setup via sshuttle).

The worst part is that DNS lookup via `socket.gethostbyname()` ignores
the usual timeout mechanisms. Therefore introduce a hard cutoff by
running the blocking initialisation code in a separate thread. The
constructed store (or an exception) is returned via a queue. This
smacks of `concurrent.futures` but avoids adding yet another dependency
on Python 2 just for this (revisit when Python 3 only...).

Test that the initialisation happens within the expected time period.

### DIFF
--- a/katdal/chunkstore_s3.py
+++ b/katdal/chunkstore_s3.py
@@ -20,6 +20,7 @@ import contextlib
 import io
 import threading
 import Queue
+import sys
 import urlparse
 
 import numpy as np
@@ -148,8 +149,8 @@ class S3ChunkStore(ChunkStore):
             """Construct chunk store and return it (or exception) via queue."""
             try:
                 queue.put(cls._from_url(url, timeout, **kwargs))
-            except BaseException as exc:
-                queue.put(exc)
+            except BaseException:
+                queue.put(sys.exc_info())
 
         thread = threading.Thread(target=_from_url, args=(url, timeout),
                                   kwargs=kwargs)
@@ -164,10 +165,11 @@ class S3ChunkStore(ChunkStore):
             raise StoreUnavailable('Timed out, possibly due to DNS lookup '
                                    'of {} stalling'.format(hostname))
         else:
-            if isinstance(result, BaseException):
-                raise result
-            else:
+            if isinstance(result, cls):
                 return result
+            else:
+                # Assume result is (exception type, exception value, traceback)
+                raise result[0], result[1], result[2]
 
     def get_chunk(self, array_name, slices, dtype):
         """See the docstring of :meth:`ChunkStore.get_chunk`."""
--- a/katdal/chunkstore_s3.py
+++ b/katdal/chunkstore_s3.py
@@ -18,6 +18,9 @@
 
 import contextlib
 import io
+import threading
+import Queue
+import urlparse
 
 import numpy as np
 try:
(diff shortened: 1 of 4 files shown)


katdal/test/test_chunkstore_s3.py
```diff
@@ -120,7 +120,14 @@ def array_name(self, path):
...
         # Drastically reduce the default botocore timeout of nearly 7 seconds
         assert_raises(StoreUnavailable, S3ChunkStore.from_url,
-                      'http://apparently.invalid/', timeout=0.1)
+                      'http://apparently.invalid/',
+                      timeout=0.1, extra_timeout=0)
+        # Some pathological DNS setups (sshuttle?) ignore the original timeout,
+        # preferring to wait 25-30 seconds instead... Check that it is fixed.
+        assert_raises(StoreUnavailable, S3ChunkStore.from_url,
+                      'http://a-valid-domain-is-somehow-harder.kat.ac.za/',
+                      timeout=0.1, extra_timeout=1)
```
bmerry: I don't see how this is testing a slow `getaddrinfo` - surely you need to mock it?

katdal/test/test_chunkstore_s3.py
```diff
@@ -120,7 +120,14 @@ def array_name(self, path):
         bucket = 'katdal-unittest'
         return self.store.join(bucket, path)
 
+    @timed(0.1 + 0.1 + 1 + 0.05)
```
bmerry: I would rather split this into two separate timed tests. At the moment if it fails you won't know which one was slow.

katdal/chunkstore_s3.py
```diff
@@ -131,6 +110,65 @@ def from_url(cls, url, timeout=10, **kwargs):
...
+        # XXX This is a poor man's attempt at concurrent.futures functionality
+        # (avoiding extra dependency on Python 2, revisit when Python 3 only)
+        queue = Queue.Queue()
+
+        def _from_url(url, timeout, **kwargs):
+            """Construct chunk store and return it (or exception) via queue."""
+            try:
+                queue.put(cls._from_url(url, timeout, **kwargs))
+            except BaseException as exc:
+                queue.put(exc)
```
bmerry: This will lose the exception traceback on Python 2.

katdal/test/test_chunkstore_s3.py
ludwigschwardt: How does one mock? :-P Never mind, I'll figure it out.

I've been testing it by running it on my pathological setup at home but mocking is the sensible way. You'll make an avid tester out of me yet!

katdal/test/test_chunkstore_s3.py
ludwigschwardt: Point taken.

katdal/chunkstore_s3.py
ludwigschwardt: How do I keep the traceback? sys.exc_info() trickery?

bmerry: Yes. Take a look through the concurrent.futures backport.

katdal/test/test_chunkstore_s3.py
ludwigschwardt: Mocked.

katdal/chunkstore_s3.py

PR #160, opened 2018-05-31
Robustify and add metadata fallback

ludwigschwardt:
This is a small platter of fixes to make katdal more robust:

- Fix a fatal bug in mvftoms (see [COMM-318](https://skaafrica.atlassian.net/browse/COMM-318))
- Map a requests ReadTimeout to StoreUnavailable
- Retry S3 connections twice, to cater for network glitches
- Open the RDB metadata even if the chunk store is missing or inaccessible
- Warn when timestamps and/or data is missing (but continue opening what you have)
- Allow simple overrides of key parameters via kwargs to `katdal.open` as well as query strings in the original URL

### DIFF
--- a/katdal/chunkstore_s3.py
+++ b/katdal/chunkstore_s3.py
@@ -93,7 +93,7 @@ class S3ChunkStore(ChunkStore):
         if timeout is not None:
             config_kwargs['read_timeout'] = int(timeout)
             config_kwargs['connect_timeout'] = int(timeout)
-            config_kwargs['retries'] = {'max_attempts': 0}
+            config_kwargs['retries'] = {'max_attempts': 2}
         # Split keyword arguments into config settings and create_client args
         for k, v in kwargs.items():
             if k in botocore.config.Config.OPTION_DEFAULTS:
--- a/katdal/chunkstore_s3.py
+++ b/katdal/chunkstore_s3.py
@@ -76,7 +76,7 @@ class S3ChunkStore(ChunkStore):
             raise _botocore_import_error
         error_map = {EndpointConnectionError: StoreUnavailable,
                      ConnectionError: StoreUnavailable,
-                     requests.exceptions.RequestsException: StoreUnavailable,
+                     requests.exceptions.RequestException: StoreUnavailable,
                      client.exceptions.NoSuchKey: ChunkNotFound,
                      client.exceptions.NoSuchBucket: ChunkNotFound}
         super(S3ChunkStore, self).__init__(error_map)
--- a/katdal/chunkstore_s3.py
+++ b/katdal/chunkstore_s3.py
@@ -76,7 +76,7 @@ class S3ChunkStore(ChunkStore):
             raise _botocore_import_error
         error_map = {EndpointConnectionError: StoreUnavailable,
                      ConnectionError: StoreUnavailable,
-                     requests.exceptions.ReadTimeout: StoreUnavailable,
+                     requests.exceptions.RequestsException: StoreUnavailable,
                      client.exceptions.NoSuchKey: ChunkNotFound,
                      client.exceptions.NoSuchBucket: ChunkNotFound}
         super(S3ChunkStore, self).__init__(error_map)
--- a/katdal/chunkstore_s3.py
+++ b/katdal/chunkstore_s3.py
@@ -36,6 +36,7 @@ except ImportError as e:
 else:
     import botocore.config
     import botocore.session
+    import botocore.vendored.requests as requests
     from botocore.exceptions import (ConnectionError, EndpointConnectionError,
                                      NoCredentialsError, ClientError)
 
@@ -75,6 +76,7 @@ class S3ChunkStore(ChunkStore):
             raise _botocore_import_error
         error_map = {EndpointConnectionError: StoreUnavailable,
                      ConnectionError: StoreUnavailable,
+                     requests.exceptions.ReadTimeout: StoreUnavailable,
                      client.exceptions.NoSuchKey: ChunkNotFound,
                      client.exceptions.NoSuchBucket: ChunkNotFound}
         super(S3ChunkStore, self).__init__(error_map)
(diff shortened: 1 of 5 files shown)


katdal/datasources.py
```diff
@@ -164,7 +164,8 @@ def name(self):
         return name
 
 
-def view_capture_stream(telstate, capture_block_id=None, stream_name=None):
+def view_capture_stream(telstate, capture_block_id=None, stream_name=None,
+                        **kwargs):
```
bmerry: kwargs needs to be added to the docstring. It doesn't seem to be used anywhere - is the intent that it's just there to swallow arguments that aren't `capture_block_id` or `stream_name`?

katdal/datasources.py
```diff
@@ -247,6 +248,37 @@ def _shorten_key(telstate, key):
...
+            return NpyFileChunkStore(store_path)
+        try:
+            s3_endpoint_url = s3_endpoint_url or telstate['s3_endpoint_url']
+        except KeyError as e:
+            raise StoreUnavailable('[{}] {}'.format(type(e).__name__, e))
+        else:
+            return S3ChunkStore.from_url(s3_endpoint_url)
+    except StoreUnavailable as e:
+        logger.warn('StoreUnavailable: %s', e)
+        logger.warn('Chunk store unavailable - opening metadata only')
```
bmerry: Maybe just log one line e.g. `Chunk store unavailable (%s) - opening metadata only`.

katdal/chunkstore_s3.py
```diff
@@ -75,6 +76,7 @@ def __init__(self, client):
             raise _botocore_import_error
         error_map = {EndpointConnectionError: StoreUnavailable,
                      ConnectionError: StoreUnavailable,
+                     requests.exceptions.ReadTimeout: StoreUnavailable,
```
bmerry: Shouldn't this catch `requests.exceptions.RequestsException`? There are lots of exception cases other than ReadTimeout (http://docs.python-requests.org/en/master/_modules/requests/exceptions/)

katdal/datasources.py
ludwigschwardt: Correct. I was going to update the docs some day... Let's make it today then.

katdal/chunkstore_s3.py
ludwigschwardt: Thanks, I was looking for such a more general connection-related catch-all but looked on the wrong side (the botocore client side, that is). I'll use that.

katdal/datasources.py
ludwigschwardt: I had it exactly like that but did not like the output. The %s can be quite long and an eye-strain, making one miss the important bit at the end (`opening metadata only`).

katdal/chunkstore_s3.py
bmerry: One thought: does error_map work with the class hierarchy or do you need an exact type match?

katdal/datasources.py
bmerry: Maybe move the %s to the end then? My thinking is that Elasticsearch treats each log message as a document, so when split into two lines like this you get two documents which could end up interleaved with log messages from other processes etc.

katdal/chunkstore_s3.py
ludwigschwardt: Both of sorts, actually. It first looks up the exact type and if not found, looks for the first key in error_map that is a base class of the exception.

bmerry:
I have a vaguely uneasy feeling about passing arbitrary URL parameters in `**kwargs` everywhere. It means that passing a bad URL can cause crashes due to using an argument that is already the name of a positional argument (causing a crash because an argument is given twice) or is an argument that was never intended to take a string.

An alternative would be to pass the URL parameters (and kwargs to `katdal.open`) as a specific argument to the lower-level functions, which could pull out values that they've whitelisted for overrides. They could also delete those arguments from the dict, and then at the end if the dict isn't empty you could warn about unused arguments. I realise this reduces flexibility though since it prevents overriding anything you didn't think to whitelist, so if you prefer to keep it as is I won't object.

bmerry:
Re metadata-only observations: presumably you're adding this support because there are users that are only interested in the metadata and don't care if the data is available. To support them, why not add an option to `katdal.open` to indicate that only the metadata is desired, in which case you don't even try to open the chunk store and don't print a warning. If the option is tri-state (don't want data, want data, or None for existing behaviour), then users who indicate they want the data can get to see  the StoreUnavailable exception instead of getting a TypeError when they try to read the data.

ludwigschwardt:
These are all good opinions that I had myself to varying levels of conviction.

On the URL kwargs: I am also slightly concerned about it. I'm thinking of putting it out there to see if anyone will use it and revoking it if too troublesome.

On the `katdal.open` kwargs: No qualms, it always worked like that (i.e. magically).

On metadata-only scenarios: I like your idea of a `metadata_only` parameter. The user basically promises not to touch `vis` and friends, which definitely simplifies the error handling.

ludwigschwardt:
OK, I've addressed most concerns and reworked the whole metadata-only idea.

katdal/datasources.py
```diff
@@ -249,35 +250,51 @@ def _shorten_key(telstate, key):
...
+    Raises
+    ------
+    KeyError
+        If telstate lacks critical keys
+    :exc:`katdal.chunkstore.StoreUnavailable`
+        If the chunk store could not be constructed
+    """
+    # Use overrides if provided, regardless of URL and telstate (NPY first)
+    if npy_store_path and os.path.isdir(npy_store_path):
+        return NpyFileChunkStore(npy_store_path)
```
bmerry: If this is a user override, I think you should fail hard if it doesn't exist, rather than silently falling back to other detection logic.

katdal/datasources.py
```diff
@@ -249,35 +250,51 @@ def _shorten_key(telstate, key):
...
+    if s3_endpoint_url:
+        return S3ChunkStore.from_url(s3_endpoint_url, **kwargs)
+    # NPY chunk store is an option if the dataset is an RDB file
+    if url_parts.scheme == 'file':
+        # Look for adjacent data directory (presumably containing NPY files)
+        rdb_path = os.path.abspath(url_parts.path)
+        store_path = os.path.dirname(os.path.dirname(rdb_path))
+        if not store_path:
+            store_path = os.path.curdir
```
bmerry: Under what conditions could `store_path` be false, given that you started with an absolute path and `os.path.dirname('/') == '/'`?

PR #161, opened 2018-06-05
Fix concatenated datasets

ludwigschwardt:
This hopefully restores the usability of concatenated datasets.

Two fixes:
  - When a sub-dataset has an empty selection on a specific sensor, it got the dtype and shape wrong, which in turn polluted the concatenated sensor.
  - This, however, was a symptom of a bigger issue, namely that identical subarrays and spectral windows were not picked up across the multiple datasets, resulting in a disjoint concatenated dataset (the worst kind).

The one outstanding issue is concatenation of cal products but I'll deal with this as part of applycal.

This addresses JIRA ticket [SR-1193](https://skaafrica.atlassian.net/browse/SR-1193).

### DIFF
--- a/katdal/dataset.py
+++ b/katdal/dataset.py
@@ -39,21 +39,6 @@ class BrokenFile(Exception):
     """Data set could not be loaded because file is inconsistent or misses critical bits."""
 
 
-def array_equal(a1, a2):
-    """True if two arrays have the same shape and elements, False otherwise.
-
-    This is meant to be identical to :func:`numpy.array_equal` but should also
-    work for variable-sized arrays containing strings. See the discussion at
-    http://mail.scipy.org/pipermail/numpy-discussion/2007-February/025967.html.
-
-    """
-    try:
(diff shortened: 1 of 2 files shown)


katdal/dataset.py
```diff
@@ -88,21 +74,34 @@ def __init__(self, ants, corr_products):
...
+               (len(self.ants), len(self.inputs), len(self.corr_products),
+                id(self))
+
+    @property
+    def _description(self):
+        """Complete string representation, used internally for comparisons."""
+        ants = '\n'.join(ant.description for ant in self.ants)
+        corrprods = ' '.join('%s,%s' % (inpA, inpB)
+                             for inpA, inpB in self.corr_products)
+        return '\n'.join((ants, corrprods))
```
bmerry: You're no longer comparing `inputs`. Could that lead to incorrect comparisons if two Subarrays have the same antennas and corrprods but the inputs are listed in a different order?

PR #162, opened 2018-06-11
Fix multiple boolean indexing in dask indexer

ludwigschwardt:
This never worked due to a missing initialisation and was only tested now by @lmagnus...


ludwigschwardt:
Added a unit test.

PR #163, opened 2018-06-11
Fix sensor props for v4 observation labels

ludwigschwardt:
The compscan labels have always been exposed via the 'Observation/label'
sensor, which crucially allows repeats. In v4 land this sensor is
initialised from the 'obs_label' telstate sensor, but that sensor is
extracted without the same sensor props. Add the relevant props.

I've toyed with the idea of getting obs_label props directly from the
existing Observation/label props, but I think there is no more need for
the `transform=str` bit, for one, so rather be explicit.


spassmoor:
This works

PR #164, opened 2018-06-12
Incorporate L1 flags into L0 product

ludwigschwardt:
If telstate has the 'sdp_archived_streams' key and it lists additional
streams (only in the full RDB), take each stream's 'chunk_info' dict
and let it override the corresponding entries in the default L0 one.
This specifically allows the additional 'sdp_l1_flags' stream to
override the 'flags' array in 'sdp_l0'.

In the process, refactor the chunk_info entries to contain the relevant
prefix aka base_name aka chunk_name, which is the prefix to be added
to the basic array name to find the array in the chunk store. This
makes it trivial to mix and match arrays between arbitrary S3 buckets
and NPY subdirectories. It's a pity this was not the original design
as the 'chunk_name' telstate key could then be chucked...

This addresses JIRA ticket [SR-1157](https://skaafrica.atlassian.net/browse/SR-1157).

### DIFF
--- a/katdal/datasources.py
+++ b/katdal/datasources.py
@@ -249,6 +249,14 @@ def _shorten_key(telstate, key):
     return ''
 
 
+def _ensure_prefix_is_set(chunk_info, telstate):
+    """Augment `chunk_info` with chunk name prefix if not set."""
+    for info in chunk_info.values():
+        if 'prefix' not in info:
+            info['prefix'] = telstate['chunk_name']
+    return chunk_info
+
+
 def _infer_chunk_store(url_parts, telstate, npy_store_path=None,
                        s3_endpoint_url=None, **kwargs):
     """Construct chunk store automatically from dataset URL and telstate.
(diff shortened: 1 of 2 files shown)


katdal/datasources.py
```diff
@@ -343,8 +342,31 @@ def __init__(self, telstate, chunk_store=None, timestamps=None,
         if chunk_store is None:
             data = None
         else:
-            data = ChunkStoreVisFlagsWeights(
-                chunk_store, telstate['chunk_name'], telstate['chunk_info'])
+            # Augment chunk_info with chunk name prefix of current stream
+            prefix = telstate['chunk_name']
+            chunk_info = telstate['chunk_info']
+            for info in chunk_info.values():
+                info['prefix'] = prefix
```
bmerry: Do you want to make this replace only if it doesn't exist? That would allow us to update viswriter/flagwriter to populate this chunk_info entry in future.

sratcliffe:
Agree with @bmerry - we don't need to make this mechanism completely generic (especially for now). For the foreseeable future the l1_flags is the only override we need to worry about.

bmerry:
> Out of curiosity, what is telstate['sdp_l1_flags_calibrations_applied'] equal to ['cal'] for?

It tells you what set of calibration solutions were applied before flagging - mostly as a future-proofing against a time when we also have self-cal solutions.

ludwigschwardt:
This can only go in once the archive has been cleansed... see JIRA ticket [SR-1235](https://skaafrica.atlassian.net/browse/SR-1235).

ludwigschwardt:
The archive has been cleansed... proceed.

PR #167, opened 2018-06-20
Authenticate S3ChunkStore using OpenID Connect

bmerry:
Apart from the changes to support authentication, it now also
- actually makes defusedxml optional (previously it was listed as optional in setup.py, but would probably have failed if it wasn't installed)
- removes the timeout on the connection opening process, since that process may now involve user interaction.


katdal/auth.py
```diff
@@ -0,0 +1,169 @@
...
+"""
+
+from __future__ import print_function
+import json
+import requests
+import threading
+import urlparse
+import logging
+
+import oic
```
ludwigschwardt: Out of curiosity, why `oic` and not `pyoidc`?

katdal/auth.py
```diff
@@ -0,0 +1,169 @@
...
+                print(login_url)
+                print()
+                code = raw_input('Enter the code: ')
+                auth_json = json.dumps({"state": self._state, "nonce": nonce, "code": code})
+                self._client.parse_response(AuthorizationResponse, info=auth_json,
+                                            sformat="json")
+                access = self._client.do_access_token_request(state=self._state,
+                                                              request_args={"code": code})
+                if isinstance(access, TokenErrorResponse):
+                    print(access.get('error_description', 'unknown error'))
```
ludwigschwardt: Does this turn into `StoreUnavailable` eventually, or how are errors reported?

setup.py
```diff
@@ -59,10 +59,9 @@
                         'katsdptelstate[rdb]', 'dask[array]'],
       extras_require={
           'ms': ['python-casacore >= 2.2.1', 'numba'],
-          's3': ['requests', 'defusedxml'],
           # rados is not in PyPI but available as Debian package python-rados
           'rados': ['rados'],
-          # katsdpauth is currently only available via GitHub
-          'auth': ['katsdpauth'],
+          's3': ['requests', 'defusedxml', 'oic'],
+          'auth': [],
```
ludwigschwardt: It is OK to remove this now.

katdal/auth.py
bmerry: For some reason the package is pyoidc but the module is oic. One of those annoying ones.

katdal/auth.py
bmerry: It goes around the while loop again and asks the user to try again.

setup.py
bmerry: Oh good, I did that in the other branch now.

bmerry:
Closing the PR since we're not going to deploy this at this time. I'll keep the branch alive for now.

PR #168, opened 2018-06-27
Add support for token-based authentication

bmerry:
This is to pair with ska-sa/katsdpauth#5

### DIFF
--- a/katdal/chunkstore_s3.py
+++ b/katdal/chunkstore_s3.py
@@ -61,7 +61,8 @@ class _BearerAuth(requests.auth.AuthBase):
         self._token = token
 
     def __call__(self, r):
-        r.headers['Authorization'] = 'Bearer ' + token
+        r.headers['Authorization'] = 'Bearer ' + self._token
+        return r
 
 
 def _raise_for_status(response):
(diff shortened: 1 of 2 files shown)


katdal/chunkstore_s3.py
```diff
@@ -166,29 +186,28 @@ def from_url(cls, url, timeout=10, extra_timeout=1, **kwargs):
         extra_timeout : int or float, optional
             Additional timeout, useful to terminate e.g. slow DNS lookups
             without masking read / connect errors (ignored if `timeout` is None)
+        token : str
+            Bearer token to authenticate
         kwargs : dict
-            Extra keyword arguments: config settings or create_client arguments
+            Extra keyword arguments (unused)
```
sratcliffe: Just my personal view on this: Either remove these, or log a warning if someone uses them. Providing an interface that doesn't do anything feels a bit off...

ludwigschwardt: It needs to be there at the moment, since it's part of the glorious free-for-all `katdal.open(**kwargs)` chain... The `kwargs` could contain keys such as `capture_block_id` and `stream_name` which are not meant for the `S3ChunkStore` but passed in nonetheless from the top.

There are better ways of doing this but the priority seems to be low.

bmerry: I've tried that before but Ludwig pointed out that removing it would break the way katdal.open works: it takes kwargs and passes them to multiple places that may have an interest in taking config from them. If one makes it a warning, then you end up getting the warning when kwargs are passed that are legitimate for other consumers.

I still don't like it because misspelt keyword args get silently dropped, but fixing that needs a higher-level redesign.

katdal/chunkstore_s3.py
```diff
@@ -52,6 +53,18 @@ def send(self, request, stream=False, timeout=None, *args, **kwargs):
         return super(_TimeoutHTTPAdapter, self).send(request, stream, timeout, *args, **kwargs)
 
 
+class _BearerAuth(requests.auth.AuthBase):
+    def __init__(self, token):
+        # Character set from RFC 6750
```
ludwigschwardt: Are we keeping OAuth 2.0 compatible tokens then?

bmerry: RFC 6750 isn't exclusive to OAuth 2.0. It can be used anywhere that a single token is used for authentication.

PR #170, opened 2018-07-04
Optimise DaskLazyIndexer for small reads from large files

bmerry:
The implementation of the dask array optimizer is tuned for cases where
most of the graph is retained, which doesn't match the use for lazy
indexing to load small pieces at a time.

### DIFF
--- a/katdal/lazy_indexer.py
+++ b/katdal/lazy_indexer.py
@@ -21,6 +21,7 @@ import threading
 
 import numpy as np
 import dask.array as da
+import dask.optimization
 
 # TODO support advanced integer indexing with non-strictly increasing indices (i.e. out-of-order and duplicates)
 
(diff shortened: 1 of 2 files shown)


bmerry:
With this change, dss-hs-1 is running mvftoms -f at 600MiB/s (with 1 dump at a time). If I disable it doing anything after katdal loading it reads at about 1500MB/s, so it's now mostly limited by the transpose performance. It doesn't seem to push the CPUs to 100%, so it is probably possible to improve the scheduling (maybe just by throwing more threads at it) to get more performance.

bmerry:
@sjperkins you might find this interesting if you're loading data from katdal with dask.

sjperkins:
Thanks. Something to watch out for.

katdal/lazy_indexer.py
```diff
@@ -443,11 +444,26 @@ def dataset(self):
...
+
+        This is functionally equivalent to ``self.dataset[keep]``, but it culls
+        unnecessary nodes from the graph, which makes it cheaper to compute if
+        only a small piece of the graph is needed.
+        """
+        # dask does culling anyway as part of optimization, but it first calls
+        # ensure_dict, which copies all the keys, presumably to speed up the
+        # case where most keys are retained. A lazy indexer is normally used to
+        # fetch a small part of the data.
+        kept = self.dataset[keep]
```
ludwigschwardt: How does this differ from the removed line 450?

katdal/lazy_indexer.py
```diff
@@ -443,11 +444,26 @@ def dataset(self):
...
+        kept.dask = dask.optimization.cull(kept.dask, kept.__dask_keys__())[0]
```
ludwigschwardt: Does this prevent further optimizations?

katdal/lazy_indexer.py
bmerry: It's exactly the same. The magic is in the next line.

katdal/lazy_indexer.py
bmerry: No, and in fact it'll end up going through the cull process a second time, which is a bit wasteful, but overall it's still a major improvement unless you're passing the whole file through dask.

sratcliffe:
Thank Bruce - nice find !

Is there anything further to be gained by doing more than one timestamp at a time in mvftoms ?

bmerry:
> Is there anything further to be gained by doing more than one timestamp at a time in mvftoms ?

I think there is still a small speedup available, but it's more likely to be 10% than 2x (and that's just on the time spent reading the data, which is now less than half the total time).

PR #171, opened 2018-07-05
Speed up flagging of lost data

bmerry:
Instead of synthesizing a sequence of masks (which are mostly zero), it
looks at the list of missing chunks, and computes the pieces of each
flag chunk that need the mask applied. A single map_blocks is then used
to apply the mask only where needed. This gives a 10x speedup.

### DIFF
--- a/katdal/chunkstore.py
+++ b/katdal/chunkstore.py
@@ -458,46 +458,8 @@ class ChunkStore(object):
         """
         raise NotImplementedError
 
-    def has_array(self, array_name, chunks, offset=()):
-        """Check if array is in the store.
-
-        This is an optional optimised version of :meth:`has_dask_array`.
-
-        Parameters
-        ----------
-        array_name : string
-            Identifier of array in chunk store
-        chunks : tuple of tuples of ints
-            Chunk specification
-        offset : tuple of int, optional
-            Offset to add to each dimension when addressing chunks in store
-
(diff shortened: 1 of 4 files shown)


katdal/chunkstore.py
```diff
@@ -477,62 +477,25 @@ def has_array(self, array_name, chunks, offset=()):
         success : :class:`numpy.ndarray` object
             Array of bools indicating presence of each chunk
 
-        Raises
-        ------
-        NotImplementedError
-            If the underlying store does not have an efficient implementation
+        Notes
+        -----
+        If the underlying store implements :meth:`list_chunk_ids`, that is
```
sratcliffe: @ludwigschwardt - why not just mandate that ALL stores implement list_chunk_ids (since the important ones already do). I see that the dict store has neither list_chunk_ids nor has_chunk...

ludwigschwardt: Dict store cannot support list_chunk_ids. It has no set chunks and allows arbitrary chunking via getitem. It does support has_chunk (look at the base class).

bmerry: Maybe that's something we should change, so that it works more like the others i.e. for each array it stores a dictionary mapping chunk ID to a numpy array?

ludwigschwardt:
Finally had a proper look at this - nice!

PR #172, opened 2018-07-25
Let L1 flag dumps match those of L0

ludwigschwardt:
The number of dumps in the `sdp_l1_flags` stream may differ from that of the source `sdp_l0` stream, since the chunks are produced in separate unsynchronised writers (`flag_writer` vs `vis_writer`) which may experience different load conditions and dropouts. Consider the `sdp_l0` stream to be the gold standard (it has the visibilities, after all) and extend or truncate the number of dumps in the `sdp_l1_flags` chunk info to match it.

This addresses JIRA ticket [MKAIV-1220](https://skaafrica.atlassian.net/browse/MKAIV-1220).

### DIFF
--- a/katdal/datasources.py
+++ b/katdal/datasources.py
@@ -272,8 +272,8 @@ def _upgrade_chunk_info(chunk_info, improved_chunk_info):
         time_chunks = chunks[0]
         n_improved_dumps = sum(time_chunks)
         if n_improved_dumps != n_original_dumps:
-            logger.debug('Original %r array has %d dumps while improved '
-                         'version has %d - forcing it to %d dumps', key,
+            logger.debug("Original '%s' array has %d dumps while improved "
+                         "version has %d - forcing it to %d dumps", key,
                          n_original_dumps, n_improved_dumps, n_original_dumps)
             # Extend/truncate improved version to match original number of dumps
             if n_improved_dumps < n_original_dumps:
@@ -285,8 +285,8 @@ def _upgrade_chunk_info(chunk_info, improved_chunk_info):
         shape = improved_info['shape']
         improved_info['shape'] = (sum(time_chunks),) + shape[1:]
         if improved_info['shape'] != original_info['shape']:
-            raise ValueError('Original {!r} array has shape {} while improved'
-                             'version has shape {}, even after fixing dumps'
+            raise ValueError("Original '{}' array has shape {} while improved"
+                             "version has shape {}, even after fixing dumps"
                              .format(key, original_info['shape'],
                                      improved_info['shape']))
         chunk_info[key] = improved_info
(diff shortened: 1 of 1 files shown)


ludwigschwardt:
Yes, I know, there are no tests for this... I need to test the whole TelstateDataSource at some point.

katdal/datasources.py
```diff
@@ -263,6 +263,36 @@ def _ensure_prefix_is_set(chunk_info, telstate):
...
 
 
+def _upgrade_chunk_info(chunk_info, improved_chunk_info):
+    """Replace chunk info items with better ones while preserving number of dumps."""
+    for key, improved_info in improved_chunk_info.items():
+        original_info = chunk_info.get(key, improved_info)
+        n_original_dumps = original_info['shape'][0]
+        chunks = improved_info['chunks']
+        time_chunks = chunks[0]
+        n_improved_dumps = sum(time_chunks)
```
bmerry: Looks a bit odd that n_original_dumps is taken from shape but n_improved_dumps is taken from summing chunks.

katdal/datasources.py
```diff
@@ -263,6 +263,36 @@ def _ensure_prefix_is_set(chunk_info, telstate):
...
+        if n_improved_dumps != n_original_dumps:
+            logger.debug("Original '%s' array has %d dumps while improved "
+                         "version has %d - forcing it to %d dumps", key,
+                         n_original_dumps, n_improved_dumps, n_original_dumps)
+            # Extend/truncate improved version to match original number of dumps
+            if n_improved_dumps < n_original_dumps:
+                time_chunks += (n_original_dumps - n_improved_dumps) * (1,)
+            else:
+                while time_chunks and sum(time_chunks) > n_original_dumps:
```
bmerry: Shouldn't be necessary to check if time_chunks is non-empty: sum of an empty sequence is 0.

katdal/datasources.py
```diff
@@ -263,6 +263,36 @@ def _ensure_prefix_is_set(chunk_info, telstate):
...
+                    time_chunks = time_chunks[:-1]
```
bmerry: This algorithm is quadratic time (both because you're summing each time, and because you're creating a new tuple each time). You can make it linear by iterating through the chunks and keeping a running total until the running total is >= n_original_dumps.

katdal/datasources.py
ludwigschwardt: I realise that, but added that later in case something *really* goes wrong (let's say for some bizarre reason `n_original_dumps` is negative).

katdal/datasources.py
ludwigschwardt: I had `n_improved_dumps = improved_info['shape'][0]` at first, then I removed the line completely, just doing `sum(time_chunks)` wherever needed. Then I relented and added the line back in. I'll go back to the original then - it seems better now.

katdal/datasources.py
ludwigschwardt: Good point - I was being lazy.

ludwigschwardt:
Nice! I have always found a need for that version of cumsum.

PR #173, opened 2018-08-03
Python 3 support

bmerry:
The first two commits are automated changes made by futurize (with a little manual intervention), and probably make sense to review individually.

The remaining commits have a bit of back-and-forth on how to convert things to native strings, and it might be quicker to review as a block.

I'm marking as don't merge yet because I want to revert 5a48a78 and update h5py in katsdpdockerbase instead, but it's ready for review.


ludwigschwardt:
This addresses JIRA ticket [SR-1210](https://skaafrica.atlassian.net/browse/SR-1210).

PR #174, opened 2018-08-16
Support AWS-style authentication to S3ChunkStore

bmerry:
The token-based authentication authenticates with our JWT-based proxy.
This adds credential authentication using an access key and secret key.
It relies on botocore to generate the signatures, which is added as an
optional dependency ("s3credentials").

Also adds a create_array method to ChunkStore, which should be used
before calling put_chunk or put_dask_array to ensure that the bucket
(S3ChunkStore) or directory (NpyFileChunkStore) is created, and that the
bucket policy allows public access if requested.

### DIFF
--- a/katdal/chunkstore.py
+++ b/katdal/chunkstore.py
@@ -211,6 +211,21 @@ class ChunkStore(object):
             chunk_name, shape = self.chunk_metadata(array_name, slices)
             return np.zeros(shape, dtype)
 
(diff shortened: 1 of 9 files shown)


bmerry:
This will break the current vis_writer.py script since it expects put_chunk to create the directory, but it pins a specific version of katdal so it won't cause issues in production.

PR #177, opened 2018-08-23
Add mark_complete and is_complete to ChunkStore

bmerry:
This was being handled by katsdpdatawriter, but in a way that would only
work for NpyFileChunkStore. This will allow katsdpdatawriter's internals
to be chunk store agnostic.

It's not implemented for RadosChunkStore, because I don't have a
convenient way to test that. We should consider just deleting
RadosChunkStore since it's never used nor planned to be used.

### DIFF
--- a/katdal/chunkstore.py
+++ b/katdal/chunkstore.py
@@ -289,6 +289,29 @@ class ChunkStore(object):
         else:
             return True
 
+    def mark_complete(self, array_name):
+        """Write a special object to indicate that `array_name` is finished.
+
+        This operation is idempotent.
+
+        The `array_name` need not correspond to any array written with
+        :meth:`put_chunk`. This has no effect on katdal, but a producer can
+        call this method to provide a hint to a consumer that no further data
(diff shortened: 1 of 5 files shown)


ludwigschwardt:
Remind me again - this is for the benefit of command-line tools like rclone and vis_trawlers that don't connect to telstate, right? Otherwise the completion can coincide with the presence of chunk_info or other keys.

bmerry:
> Remind me again - this is for the benefit of command-line tools like rclone and vis_trawlers that don't connect to telstate, right? Otherwise the completion can coincide with the presence of chunk_info or other keys.

Exactly. I think connecting to telstate may be problematic due to lifetime problems i.e. telstate disappearing when the subarray is done, whereas vis trawler runs on its own independent schedule.

bmerry:
Urgh, Jenkins is being very annoying about timeouts again. I have found some evidence of dodgy DNS on sdp-services though (one of its DNS servers doesn't exist), so hopefully Martin can fix that tomorrow and then we'll see if that improves things.

PR #178, opened 2018-08-28
Add new requests requirements

bmerry:
Needed due to ska-sa/katsdpdockerbase#18.


bmerry:
I'd missed certifi, so I've added it directly to master.

PR #180, opened 2018-08-28
Upgrade DaskLazyIndexer

ludwigschwardt:
This prepares to prepare for applycal...

The DaskLazyIndexer is given a bit of a facelift, with explicit oindex semantics (see JIRA [SR-1323](https://skaafrica.atlassian.net/browse/SR-1323)) and a way to add transforms after instantiation.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -6,6 +6,7 @@
 # Packages
 *.egg
 *.egg-info
+*.eggs
 dist
 build
 eggs
--- a/.gitignore
+++ b/.gitignore
@@ -6,10 +6,10 @@
 # Packages
 *.egg
 *.egg-info
-*.eggs
 dist
 build
 eggs
+.eggs
 parts
 bin
 var
--- a/katdal/lazy_indexer.py
+++ b/katdal/lazy_indexer.py
@@ -95,6 +95,20 @@ def _simplify_index(indices, shape):
     return tuple(out)
 
 
+def _dask_oindex(x, indices):
+    """Perform outer indexing on dask array `x`, one dimension at a time.
+
+    It is assumed that `indices` is suitably normalised (no ellipsis, etc.)
+    """
+    axis = 0
+    for index in indices:
+        x = da.take(x, index, axis=axis)
+        # If axis wasn't dropped by a scalar index:
+        if not isinstance(index, Integral):
+            axis += 1
+    return x
+
+
 def _dask_getitem(x, indices):
     """Index a dask array, with N-D fancy index support and better performance.
 
@@ -119,14 +133,7 @@ def _dask_getitem(x, indices):
     try:
         out = x[indices]
     except NotImplementedError:
-        # Perform outer indexing, one dimension at a time
-        out = x
-        axis = 0
-        for index in indices:
-            out = da.take(out, index, axis=axis)
-            # If axis wasn't dropped by a scalar index:
-            if not isinstance(index, Integral):
-                axis += 1
+        out = _dask_oindex(x, indices)
     # dask does culling anyway as part of optimization, but it first calls
     # ensure_dict, which copies all the keys, presumably to speed up the
     # case where most keys are retained. A lazy indexer is normally used to
--- a/katdal/lazy_indexer.py
+++ b/katdal/lazy_indexer.py
@@ -36,10 +36,10 @@ def _simplify_index(shape, indices):
 
     In the current implementation, boolean numpy arrays which select contiguous
     ranges are converted to slices. Note that when indexing along multiple axes
-    with arrays, this may change the semantics of the indexing (see this `NEP`_
-    for details).
+    with arrays, this may change the semantics of the indexing (see NumPy's
+    `NEP 21`_ for details).
 
-    .. _NEP: https://gist.github.com/seberg/976373b6a2b7c4188591
+    .. _NEP 21: http://www.numpy.org/neps/nep-0021-advanced-indexing.html
 
     If the expression is invalid in some way (e.g. wrong length array index)
     the original index is returned so that that an exception will be raised
@@ -74,6 +74,39 @@ def _simplify_index(shape, indices):
                     index = slice(selected[0], selected[-1] + 1)
             out.append(index)
     return tuple(out)
+
+
+def _dask_getitem(x, keep):
+    """Index a dask array, with N-D fancy index support and better performance.
+
+    This is a drop-in replacement for ``x[keep]`` that goes one further
+    by implementing "N-D fancy indexing" which is still unsupported in dask.
+    If `keep` contains multiple fancy indices, perform outer (`oindex`)
+    indexing. This behaviour deviates from NumPy, which performs the more
+    general (but also more obtuse) vectorized (`vindex`) indexing in this case.
+    See  NumPy `NEP 21`_, dask/dask#433 and h5py/h5py#652 for more details.
+
+    .. _NEP 21: http://www.numpy.org/neps/nep-0021-advanced-indexing.html
+
+    In addition, this optimises performance by culling unnecessary nodes from
+    the dask graph after indexing, which makes it cheaper to compute if only a
+    small piece of the graph is needed, and by collapsing fancy indices in
+    `keep` to slices where possible (which also implies oindex semantics).
+    """
+    keep = _simplify_index(x.shape, keep)
+    try:
+        kept = x[keep]
+    except NotImplementedError:
+        # Dask does not like multiple boolean indices: go one dim at a time
+        kept = x
+        for dim, keep_per_dim in enumerate(keep):
+            kept = da.take(kept, keep_per_dim, axis=dim)
+    # dask does culling anyway as part of optimization, but it first calls
+    # ensure_dict, which copies all the keys, presumably to speed up the
+    # case where most keys are retained. A lazy indexer is normally used to
+    # fetch a small part of the data.
+    kept.dask = dask.optimization.cull(kept.dask, kept.__dask_keys__())[0]
+    return kept
 
 
 # -------------------------------------------------------------------------------------------------
@@ -421,7 +454,6 @@ class DaskLazyIndexer(object):
     """
     def __init__(self, dataset, keep=(), transforms=()):
         self.name = getattr(dataset, 'name', '')
-        keep = _simplify_index(dataset.shape, keep)
         # Fancy indices can be mutable arrays, so take a copy to protect
         # against the caller mutating the array before we apply it.
         self.keep = copy.deepcopy(keep)
@@ -435,33 +467,12 @@ class DaskLazyIndexer(object):
         """Array after first-stage indexing and transformation."""
         with self._lock:
             if self._dataset is None:
-                try:
-                    dataset = self._orig_dataset[self.keep]
-                except NotImplementedError:
-                    # Dask does not like multiple boolean indices: go one dim at a time
-                    dataset = self._orig_dataset
-                    for dim, keep_per_dim in enumerate(self.keep):
-                        dataset = da.take(dataset, keep_per_dim, axis=dim)
+                dataset = _dask_getitem(self._orig_dataset, self.keep)
                 for transform in self.transforms:
                     dataset = transform(dataset)
                 self._dataset = dataset
                 self._orig_dataset = None
             return self._dataset
-
-    def dask_getitem(self, keep):
-        """Index the dataset and return a dask array.
-
-        This is functionally equivalent to ``self.dataset[keep]``, but it culls
-        unnecessary nodes from the graph, which makes it cheaper to compute if
-        only a small piece of the graph is needed.
-        """
-        # dask does culling anyway as part of optimization, but it first calls
-        # ensure_dict, which copies all the keys, presumably to speed up the
-        # case where most keys are retained. A lazy indexer is normally used to
-        # fetch a small part of the data.
-        kept = self.dataset[keep]
-        kept.dask = dask.optimization.cull(kept.dask, kept.__dask_keys__())[0]
-        return kept
 
     def __getitem__(self, keep):
         """Extract a selected array from the underlying dataset.
@@ -482,7 +493,7 @@ class DaskLazyIndexer(object):
         out : :class:`numpy.ndarray`
             Extracted output array (computed from the final dask version)
         """
-        kept = self.dask_getitem(keep)
+        kept = _dask_getitem(self.dataset, keep)
         # Workaround for https://github.com/dask/dask/issues/3595
         # This is equivalent to kept.compute(), but does not
         # allocate excessive memory.
--- a/katdal/lazy_indexer.py
+++ b/katdal/lazy_indexer.py
@@ -402,24 +402,22 @@ class DaskLazyIndexer(object):
     ----------
     dataset : :class:`dask.Array`
         The full dataset, from which a subset is chosen by `keep`.
-    keep : tuple
-        Index expression. This object wraps `dataset[keep]`, although
-        fancy indices are applied independently per axis.
-    transforms : sequence
+    keep : tuple, optional
+        Index expression describing first-stage selection (i.e. as applied by
+        :meth:`katdal.DataSet.select`). This object wraps `dataset[keep]`,
+        although fancy indices are applied independently per axis.
+    transforms : sequence of function, signature ``array = f(array)``, optional
         Transformations that are applied after indexing by `keep` but
         before indexing on this object. Each transformation is a callable
         that takes a dask array and returns another dask array.
 
     Attributes
     ----------
+    name : str
+        The name of the (full) underlying dataset, useful for reporting
     dataset : :class:`dask.Array`
-        The dask array that is accessed by indexing. It can be used
-        directly to perform dask computations.
-    keep : tuple
-        The index expression that is used to compute :attr:`dataset`.
-        It may be an alternative form to that given in the constructor.
-    transforms : list
-        The transformations given in the constructor.
+        The dask array that is accessed by indexing (after applying `keep` and
+        `transforms`). It can be used directly to perform dask computations.
     """
     def __init__(self, dataset, keep=(), transforms=()):
         self.name = getattr(dataset, 'name', '')
@@ -434,6 +432,7 @@ class DaskLazyIndexer(object):
 
     @property
     def dataset(self):
+        """Array after first-stage indexing and transformation."""
         with self._lock:
             if self._dataset is None:
                 try:
@@ -450,7 +449,7 @@ class DaskLazyIndexer(object):
             return self._dataset
 
     def dask_getitem(self, keep):
-        """Index the array and return a dask array.
+        """Index the dataset and return a dask array.
 
         This is functionally equivalent to ``self.dataset[keep]``, but it culls
         unnecessary nodes from the graph, which makes it cheaper to compute if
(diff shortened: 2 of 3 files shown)


katdal/test/test_lazy_indexer.py
```diff
@@ -73,8 +73,64 @@ def test_too_many_axes(self):
...
+        # Restrict ourselves to more practical 1-D boolean indices
+        bindx2 = np.zeros(7, dtype=np.bool_)
+        bindx2[0] = True
+        bindx3 = np.zeros(8, dtype=np.bool_)
+        bindx3[0] = True
+        bindx3[3] = True
+        self._check_shape(np.s_[:, 0, bindx2, bindx3], (5, 1, 2))
+        self._check_shape(np.s_[0, :, bindx2, bindx3], (6, 1, 2))
+        self._check_shape(np.s_[[0], :, bindx2, bindx3], (1, 6, 1, 2))
+        self._check_shape(np.s_[:, [0, 1], bindx2, bindx3], (5, 2, 1, 2))
```
bmerry: Checking just the shape is a weak test. How about checking values? It should generally be possible to get to the same value by indexing the numpy array several times.

katdal/test/test_lazy_indexer.py
```diff
@@ -73,8 +73,64 @@ def test_too_many_axes(self):
...
+
```
bmerry: Since bindx2 will be simplified, and `[0, 1]` could potentially be simplified (even if it isn't yet), in future these tests might not actually cover the oindexing fallback path at all.

Also add this test: it currently fails because the oindexing code doesn't correctly account for scalar indices:
```
self._check_shape(np.s_[:, 0, [0, 2], [1, 3, 5]], (5, 2, 3))
```
I think some tests with `np.newaxis` and `...` are needed too.

katdal/test/test_lazy_indexer.py
```diff
@@ -89,3 +145,11 @@ def test_stage1_multiple_boolean_indices(self):
         stage1 = tuple([True] * d for d in self.data.shape)
         indexer = DaskLazyIndexer(self.data_dask, stage1)
         np.testing.assert_array_equal(indexer[:], self.data)
+
+    def test_stage2_multiple_boolean_indices(self):
+        stage1 = tuple([True] * d for d in self.data.shape)
+        indexer = DaskLazyIndexer(self.data_dask, stage1)
+        stage2 = tuple([True] * 2 + [False] * (d - 2) for d in self.data.shape)
```
bmerry: This is going to be reduced by the simplifier. Should maybe have a separate test case that won't be simplified to a non-fancy case.

katdal/lazy_indexer.py
```diff
@@ -76,6 +76,39 @@ def _simplify_index(shape, indices):
...
+    `keep` to slices where possible (which also implies oindex semantics).
+    """
+    keep = _simplify_index(x.shape, keep)
+    try:
+        kept = x[keep]
+    except NotImplementedError:
+        # Dask does not like multiple boolean indices: go one dim at a time
+        kept = x
+        for dim, keep_per_dim in enumerate(keep):
+            kept = da.take(kept, keep_per_dim, axis=dim)
```
bmerry: This was suitable for stage1 where the indices were all boolean arrays, but it's inadequate for the general case. Scalar indices, `np.newaxis` and `...` will all need to be handled.

bmerry:
Incidentally, what do you think about contributing an `oindex` attribute to dask itself? I'm not sure if that would be easier or harder, but possibly one could reuse a lot of the existing machinery for dealing with the different cases, and of course other people would then review and maintain it.

.gitignore
```diff
@@ -6,6 +6,7 @@
 # Packages
 *.egg
 *.egg-info
+*.eggs
```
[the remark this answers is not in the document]
ludwigschwardt: Yes yes - too much copy and paste! (it worked though) Reverting to what katpoint has.

katdal/lazy_indexer.py
```diff
@@ -434,38 +464,48 @@ def __init__(self, dataset, keep=(), transforms=()):
...
+        This applies the given second-stage index on top of the current
+        dataset, which already has a first-stage index and optional transforms
+        applied to it. The indexer also finally stops being lazy and triggers
+        dask computation to arrive at the output array.
+
+        Parameters
+        ----------
+        keep : tuple of int or slice or sequence of int or sequence of bool
+            Second-stage index as a valid index or slice specification
+            (supports arbitrary slicing or advanced indexing on any dimension)
```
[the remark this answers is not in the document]
ludwigschwardt: Good point.

katdal/lazy_indexer.py
```diff
@@ -76,6 +76,39 @@ def _simplify_index(shape, indices):
...
 
+def _dask_getitem(x, keep):
+    """Index a dask array, with N-D fancy index support and better performance.
+
+    This is a drop-in replacement for ``x[keep]`` that goes one further
+    by implementing "N-D fancy indexing" which is still unsupported in dask.
+    If `keep` contains multiple fancy indices, perform outer (`oindex`)
+    indexing. This behaviour deviates from NumPy, which performs the more
+    general (but also more obtuse) vectorized (`vindex`) indexing in this case.
+    See  NumPy `NEP 21`_, dask/dask#433 and h5py/h5py#652 for more details.
```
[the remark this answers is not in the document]
ludwigschwardt: Will do.

katdal/test/test_lazy_indexer.py
```diff
@@ -73,8 +73,64 @@ def test_too_many_axes(self):
...
+        self._test_with(np.s_[:, [0, 1], bindx], IndexError)  # dask failure
+
+    def test_outer_indexing(self):
+        self._check_shape(np.s_[:, [0], [0, 1], :], (5, 1, 2, 8))
+        self._check_shape(np.s_[:, [0], :, [0, 1]], (5, 1, 7, 2))
+        self._check_shape(np.s_[:, [0], 0, :], (5, 1, 8))
+        self._check_shape(np.s_[:, [0], :, 0], (5, 1, 7))
+        # Restrict ourselves to more practical 1-D boolean indices
+        bindx2 = np.zeros(7, dtype=np.bool_)
+        bindx2[0] = True
```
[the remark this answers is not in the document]
ludwigschwardt: I numbered them according to the dimension they act on, but it is probably better to be 1-based.

katdal/test/test_lazy_indexer.py
```diff
@@ -73,8 +73,64 @@ def test_too_many_axes(self):
...
+        npy = self.data[indices]
+        us = _dask_getitem(self.data_dask, indices).compute()
+        np.testing.assert_array_equal(us, npy)
+
+    def _test_with(self, indices, expected_error=None):
+        if expected_error is None:
+            self._compare_indexing(indices)
+        else:
+            with assert_raises(expected_error):
+                self._compare_indexing(indices)
```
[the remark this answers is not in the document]
ludwigschwardt: Completely redid the unit tests.

katdal/test/test_lazy_indexer.py
```diff
@@ -73,8 +73,64 @@ def test_too_many_axes(self):
...
+
+    def test_legacy_fancy_indexing(self):
+        self._test_with(np.s_[[0], ...])
+        self._test_with(np.s_[:, [0], ...])
+        self._test_with(np.s_[:, [0], [0], :], AssertionError)  # future error
+        self._test_with(np.s_[:, [0], :, [0]], AssertionError)  # future error
+        self._test_with(np.s_[:, [0], 0, :])
+        self._test_with(np.s_[:, [0], :, 0], AssertionError)  # future error
```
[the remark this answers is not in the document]
ludwigschwardt: I was trying to copy the examples in NEP 21 as a benchmark for outer indexing but it is not actually that helpful.

katdal/test/test_lazy_indexer.py
ludwigschwardt: Point taken.

ludwigschwardt:
Ready to be reviewed again...

katdal/test/test_lazy_indexer.py
```diff
@@ -73,19 +74,182 @@ def test_too_many_axes(self):
...
+        # Handle dropped dimensions
+        if not _is_integer(k):
+            dim += 1
+    return result
+
+
+class TestNumPyOIndex(object):
+    """Test the :func:`~numpy_oindex` function."""
+    def setup(self):
+        shape = (10, 10, 10, 10)
```
bmerry: Use a different size on each axis so that if a code path that uses the size on a dimension picks the wrong dimension it will get picked up.

katdal/test/test_lazy_indexer.py
```diff
@@ -73,19 +74,182 @@ def test_too_many_axes(self):
         self._test_index_error(np.s_[0, 0, 0, 0])
 
 
+def slice_to_range(s, l):
+    return range(*s.indices(l))
```
bmerry: Should add `from builtins import range` so that you get the same behaviour in Python 2 and 3.

katdal/test/test_lazy_indexer.py
```diff
@@ -73,19 +74,182 @@ def test_too_many_axes(self):
...
+        self._test_with(np.s_[:, [0], [0, 1], :])
+        self._test_with(np.s_[:, [0], :, [0, 1]])
+        # booleans
+        pick_one = np.zeros(10, dtype=np.bool_)
+        pick_one[6] = True
+        self._test_with(np.s_[:, [True, False] * 5, pick_one, :])
+        self._test_with(np.s_[:, [False, True] * 5, :, pick_one])
+        # slices
+        self._test_with(np.s_[0:2, 2:4, 4:6, 6:8])
+        self._test_with(np.s_[-8:-6, -4:-2, slice(3, 10, 2), -2:])
```
bmerry: Any particular reason you've written `slice(3, 10, 2)` and not `3:10:2`?

katdal/test/test_lazy_indexer.py
```diff
@@ -73,19 +74,182 @@ def test_too_many_axes(self):
...
+        # single ints
+        self._test_with(np.s_[:, [0], 0, :])
+        self._test_with(np.s_[:, [0], :, 0])
+        self._test_with(np.s_[:, [0], -1, :])
+        self._test_with(np.s_[:, [0], :, -1])
+        # newaxis
+        self._test_with(np.s_[np.newaxis, :, [0], :, 0])
+        self._test_with(np.s_[:, [0], np.newaxis, 0, :])
```
bmerry: Could you also add a case with an integer index before newaxis e.g. `0, np.newaxis, 1, np.newaxis, 2`.

katdal/test/test_lazy_indexer.py
```diff
@@ -73,19 +74,182 @@ def test_too_many_axes(self):
...
+        # the lot
+        self._test_with(np.s_[..., 2:5, [4, 6], np.newaxis, 0],
+                        np.s_[:, 2:5, [4, 6], np.newaxis, 0])
+
```
bmerry: Could you share the list of test cases between this test and TestDaskGetitem? Maybe ellipsis cases might need some special handling, but it looks like there is a lot of duplication.

katdal/lazy_indexer.py
```diff
@@ -434,38 +478,52 @@ def __init__(self, dataset, keep=(), transforms=()):
...
+                dataset = _dask_getitem(self._orig_dataset, self.keep)
                 for transform in self.transforms:
                     dataset = transform(dataset)
                 self._dataset = dataset
                 self._orig_dataset = None
             return self._dataset
 
-    def dask_getitem(self, keep):
-        """Index the array and return a dask array.
+    def add_transform(self, transform):
```
bmerry: This function needs a unit test, covering both cases (call before or after the lazy computation is done).

katdal/test/test_lazy_indexer.py
ludwigschwardt: Done.

katdal/lazy_indexer.py
```diff
@@ -32,61 +32,59 @@
...
-                bool_array = False
-            if bool_array:
-                selected = np.nonzero(index)[0]
-                if not len(selected):
-                    index = slice(0, 0)
-                elif selected[-1] - selected[0] + 1 == len(selected):
-                    index = slice(selected[0], selected[-1] + 1)
-            out.append(index)
+            # If there is 1-D fancy index on this axis, try to convert to slice
+            if isinstance(index, np.ndarray) and index.ndim == 1:
```
bmerry: What would happen if someone passed a non-integer, non-boolean ndarray in? While it won't be legal, we ideally want to have it fail in the same way after `_simplify_index`.

katdal/lazy_indexer.py
```diff
@@ -32,61 +32,59 @@
...
+                try:
+                    index = _range_to_slice(index)
+                except ValueError:
+                    pass
+                else:
+                    index = da.slicing.normalize_slice(index, length)
```
bmerry: Is this needed? It's not part of the contract, and when the simplified index is used with `da.take` won't dask normalize it anyway?

katdal/test/test_lazy_indexer.py
```diff
@@ -123,77 +157,34 @@ def numpy_oindex_lite(x, keep):
...
 
     def test_outer_indexing(self):
         self._test_with(())
         self._test_with(2)
         self._test_with((2, 3, 4, 5))
         # ellipsis
-        self._test_with(np.s_[[0], ...])
-        self._test_with(np.s_[:, [0], ...])
+        self._test_with(np.s_[[0], ...], np.s_[[0], :, :, :])
+        self._test_with(np.s_[:, [0], ...], np.s_[:, [0], :, :])
```
bmerry: Maybe have a test where `...` is in the middle rather than at one end.

katdal/test/test_lazy_indexer.py
bmerry: Isn't this test only testing the fast path? I think all those those fancy indices will get turned into slices by `_simplify_index`. The same applies to many of the other tests now that you're simplifying strided accesses.

katdal/lazy_indexer.py
```diff
@@ -76,6 +77,52 @@ def _simplify_index(shape, indices):
...
+    the dask graph after indexing, which makes it cheaper to compute if only a
+    small piece of the graph is needed, and by collapsing fancy indices in
+    `keep` to slices where possible (which also implies oindex semantics).
+    """
+    keep = _simplify_index(x.shape, keep)
+    try:
+        kept = x[keep]
+    except NotImplementedError:
+        # Unpack ellipsis first as da.take does not like it
+        keep = da.slicing.normalize_index(keep, x.shape)
```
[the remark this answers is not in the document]
ludwigschwardt: Indeed - it's well employed now.

katdal/lazy_indexer.py
```diff
@@ -434,38 +478,52 @@ def __init__(self, dataset, keep=(), transforms=()):
...
-        # ensure_dict, which copies all the keys, presumably to speed up the
-        # case where most keys are retained. A lazy indexer is normally used to
-        # fetch a small part of the data.
-        kept = self.dataset[keep]
-        kept.dask = dask.optimization.cull(kept.dask, kept.__dask_keys__())[0]
-        return kept
+        with self._lock:
+            if self._dataset is not None:
+                self._dataset = transform(self._dataset)
+            self.transforms.append(transform)
```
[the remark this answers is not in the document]
ludwigschwardt: Done.

katdal/lazy_indexer.py
```diff
@@ -434,38 +478,52 @@ def __init__(self, dataset, keep=(), transforms=()):
...
+        applied to it. The indexer also finally stops being lazy and triggers
+        dask computation to arrive at the output array.
+
+        Both indexing stages perform "outer" indexing (aka oindex), which
+        indexes each dimension independently. This is especially relevant for
+        advanced or fancy indexing.
+
+        Parameters
+        ----------
+        keep : tuple of int or slice or sequence of int or sequence of bool
```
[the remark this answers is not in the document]
ludwigschwardt: Done.

katdal/test/test_lazy_indexer.py
ludwigschwardt: Done.

katdal/test/test_lazy_indexer.py
ludwigschwardt: Done.

katdal/test/test_lazy_indexer.py
ludwigschwardt: Don't think so, fixed.

katdal/test/test_lazy_indexer.py
ludwigschwardt: Done.

katdal/lazy_indexer.py
ludwigschwardt: Done.

katdal/lazy_indexer.py
ludwigschwardt: Both NumPy and dask's checks result in an `IndexError`. I've added a test for this, showing that `_simplify_index` and NumPy have matching errors.

katdal/lazy_indexer.py
ludwigschwardt: I think it is.

Part of the contract is a normalised index (via `_simplify_index`), which normalises slices internally. I looked at `da.take` and it does not seem to have much checking (on the `axis` parameter is validated, for one).

katdal/test/test_lazy_indexer.py
ludwigschwardt: Done.

katdal/lazy_indexer.py
```diff
@@ -434,38 +478,52 @@ def __init__(self, dataset, keep=(), transforms=()):
...
-        """Index the array and return a dask array.
+    def add_transform(self, transform):
+        """Add another transform to the end of the transform chain.
 
-        This is functionally equivalent to ``self.dataset[keep]``, but it culls
-        unnecessary nodes from the graph, which makes it cheaper to compute if
-        only a small piece of the graph is needed.
+        The `transform` is a callable that takes a dask array and returns
+        another dask array. Apply it directly to the current dataset if
+        first-stage indexing and initial transformation has already happened.
```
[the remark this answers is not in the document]
ludwigschwardt: Improved the docstring.

katdal/test/test_lazy_indexer.py
```diff
@@ -238,10 +267,28 @@ def test_stage1_multiple_boolean_indices(self):
...
 
     def test_stage2_multiple_boolean_indices(self):
         stage1 = tuple([True] * d for d in self.data.shape)
         stage2 = tuple([True] * 4 + [False] * (d - 4) for d in self.data.shape)
         self._test_with(stage1, stage2)
         stage2 = tuple([True, False] * (d // 2) for d in self.data.shape)
         self._test_with(stage1, stage2)
+        stage1 = np.s_[UNEVEN, 2 * UNEVEN, :24]
+        stage2 = np.s_[:3, [1, 2, 3, 4, 6, 9], [8, 6, 4, 2, 0]]
+        self._test_with(stage1, stage2)
```
bmerry: This last check doesn't match the name of the test (multiple boolean indices).

bmerry:
I accidentally discovered that the `test_wrong_length` test fails with dask<0.18.2. I haven't checked if that's because the normalisation functions you're relying on aren't doing their checking or if it's the reference path that's failing to check things.

katdal/test/test_lazy_indexer.py
ludwigschwardt: Fixed.

ludwigschwardt:
On `test_wrong_length`: In dask < 0.18.2 `check_index` did not check the length of boolean indices. Oops! After simplification the sin is washed away, so definitely something to fix.

I could depend on `dask>=0.18.2` but it is quite new (2018-07-23). Or I could add a workaround to be phased out. What is your preference?

bmerry:
> I could depend on dask>=0.18.2 but it is quite new (2018-07-23). Or I could add a workaround to be phased out. What is your preference?

Currently everything using katsdpdockerbase is on an older version. I guess see whether you think it will be less effort to work around it or to do the upgrade and check that other dask users (particularly katsdpcal) aren't broken by it.

ludwigschwardt:
Another round of fixes done (I've also added shiny new \_\_str\_\_ and \_\_repr\_\_ methods).

ludwigschwardt:
Yay!

PR #181, opened 2018-08-31
Optimise S3ChunkStore.put_chunk

bmerry:
Writing the numpy array to a BytesIO and passing that BytesIO to
requests is very slow, possibly because httplib reads it 8KB at a time.
The new approach is a bit more fragile but is zero-copy if the input
array is C-contiguous.


bmerry:
Rats, it looks like the support for an iterable doesn't existing Python 2.7. I'll revisit on Monday to add a fallback path.

PR #182, opened 2018-09-03
Optimise S3ChunkStore.put_chunk

bmerry:
Writing the numpy array to a BytesIO and passing that BytesIO to
requests is very slow, possibly because httplib reads it 8KB at a time.
The new approach is a bit more fragile but is zero-copy if the input
array is C-contiguous.

### DIFF
--- a/katdal/chunkstore_s3.py
+++ b/katdal/chunkstore_s3.py
@@ -126,7 +126,7 @@ class _Multipart(object):
     @property
     def len(self):
         """Total content length (retrieved by requests to set Content-Length)"""
-        return sum(len(item) for item in self.items)
+        return sum(memoryview(item).nbytes for item in self.items)
 
 
 def _raise_for_status(response):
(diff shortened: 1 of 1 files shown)


bmerry:
This is attempt 2. This time it has Python 2 support, although via a slower path (probably still faster than the original code, although I haven't benchmarked it).

katdal/chunkstore_s3.py
```diff
@@ -338,16 +359,35 @@ def create_array(self, array_name):
             with self._request(None, 'PUT', policy_url, data=json.dumps(policy)):
                 pass
 
+    @classmethod
+    def _numpy_header(cls, chunk):
+        fp = io.BytesIO()
+        header_fields = np.lib.format.header_data_from_array_1_0(chunk)
+        np.lib.format.write_array_header_1_0(fp, header_fields)
```
ludwigschwardt: I used to wonder about forcing the headers to v1 in my previous attempts at this... I guess we would not have overly elaborate dtypes in our chunk store, so v2 headers would be overkill (and less compatible).

katdal/chunkstore_s3.py
```diff
@@ -338,16 +359,35 @@ def create_array(self, array_name):
...
         url = self._chunk_url(chunk_name)
-        fp = io.BytesIO()
-        np.lib.format.write_array(fp, chunk, allow_pickle=False)
-        md5 = base64.b64encode(hashlib.md5(fp.getvalue()).digest())
-        fp.seek(0)
+        # Note: don't use ascontiguousarray as it turns 0D into 1D.
+        # See https://github.com/numpy/numpy/issues/5300
+        chunk = np.asarray(chunk, order='C')
+        # Need an array of bytes to so that _Multipart.len gives the right value
+        chunk_view = memoryview(np.ravel(chunk).view(np.uint8))
```
ludwigschwardt: Why do you need `ravel` and a `uint8` view? I would have thought that `chunk_view = memoryview(chunk)` is sufficient, given that chunk satisfies the buffer protocol and its buffer is already C-contiguous.

katdal/chunkstore_s3.py
bmerry: Yes I seem to remember us having discussions about this. I can't remember what we settled on the first time around. I figure it's a problem we can solve if we ever actually run into it.

katdal/chunkstore_s3.py
bmerry: The problem is that len(memoryview(chunk)) returns the number of items, not the number of bytes, which would break `_Multipart` since it uses `len` to get the number of bytes. The ravel is probably not necessary, I just wasn't sure what `len` would return for a multi-dimensional array.

Thinking about it more though, I think it might be better to handle this in `_Multipart` itself, so that you just put the chunk directly into `_Multipart`, and `_Multipart` takes care of wrapping things into memory views and figuring out their lengths.

bmerry: Done as described above.

bmerry:
It looks like that last simplification somehow made things even faster (went from 1.3 GB/s to 2 GB/s!), so I'm going to merge.

PR #187, opened 2018-09-18
Handle stream inheritance

bmerry:
If a vis or flags stream has an `inherit` member in telstate, the
telstate view will be constructed to have that stream in the telstate
view chain. This supports SR-1348 / ska-sa/katsdpdata#126.

view_capture_stream now just does the work of constructing a telstate
from capture_block_id and stream_name, and the "smart" version is now
view_l0_capture_stream (name chosen since it asserts that the type is
sdp.l0).

### DIFF
--- a/katdal/datasources.py
+++ b/katdal/datasources.py
@@ -173,17 +173,56 @@ class DataSource(object):
         return name
 
 
-def view_capture_stream(telstate, capture_block_id=None, stream_name=None,
-                        **kwargs):
-    """Create telstate view based on capture block ID and stream name.
+def view_capture_stream(telstate, capture_block_id, stream_name):
+    """Create telstate view based on given capture block ID and stream name.
 
-    This figures out the appropriate capture block ID and L0 stream name from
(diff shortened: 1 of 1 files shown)


katdal/datasources.py
```diff
@@ -359,13 +396,11 @@ def _upgrade_flags(chunk_info, telstate):
         logger.debug('No additional flag capture streams found: %s', e)
         return chunk_info
     for s in archived_streams:
-        telstate_s = telstate.root().view(s)
-        if telstate_s['stream_type'] != 'sdp.flags' or \
-           stream_name not in telstate_s['src_streams']:
+        telstate_cs = view_capture_stream(telstate, capture_block_id, s)
+        if telstate_cs['stream_type'] != 'sdp.flags' or \
```
ludwigschwardt: I guess this assumes that stream `s` has an explicit `stream_type` key or only inherits from streams of the same type (is that how it works?).

bmerry: Basically yes. I would think inheriting from a different type would be a bad idea though.

PR #189, opened 2018-09-25
Scale weights by autocorrelation power during read

bmerry:
If the telstate key 'need_weights_power_scale' is set on the L0 stream, the weights are divided by the product of the autocorrelation visibilities. This will be matched with changes to katsdpingest and katsdpdata to set this flag and remove the equivalent scaling done in ingest.

Relates to SR-1230.

### DIFF
--- a/katdal/datasources.py
+++ b/katdal/datasources.py
@@ -164,6 +164,9 @@ def weight_power_scale(block, auto_indices, index1, index2, out=None, tmp=None):
     for i in range(block.shape[0]):
         for j in range(block.shape[1]):
             for k in range(len(auto_indices)):
+                # Have to use np.divide here, because 1 / x raises
+                # ZeroDivisionError on divide by zero whereas we want it
+                # to produce a non-finite value.
                 auto_scale[k] = np.divide(1, block[i, j, auto_indices[k]].real)
             for k in range(block.shape[2]):
                 p = auto_scale[index1[k]] * auto_scale[index2[k]]
--- a/katdal/datasources.py
+++ b/katdal/datasources.py
@@ -103,7 +103,7 @@ def _apply_data_lost(orig_flags, lost, block_id):
     return flags
 
 
-def _corrprod_to_autocorr(corrprod):
+def corrprod_to_autocorr(corrprod):
     """Find the autocorrelation indices of correlation products.
 
     Parameters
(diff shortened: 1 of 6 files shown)


katdal/datasources.py
```diff
@@ -102,6 +103,80 @@ def _apply_data_lost(orig_flags, lost, block_id):
     return flags
 
 
+def corrprod_to_autocorr(corrprod):
```
ludwigschwardt: I'm really sensing the imminent birth of `utils.py` :-) This stuff would be very useful to `applycal` too.

katdal/datasources.py
```diff
@@ -102,6 +103,80 @@ def _apply_data_lost(orig_flags, lost, block_id):
...
+    for i in range(block.shape[0]):
+        for j in range(block.shape[1]):
+            for k in range(len(auto_indices)):
+                auto_scale[k] = np.divide(1, block[i, j, auto_indices[k]].real)
+            for k in range(block.shape[2]):
+                p = auto_scale[index1[k]] * auto_scale[index2[k]]
+                # If either or both of the autocorrelations has zero power then
+                # there is likely something wrong with the system. Set the
+                # weight to very close to zero (not actually zero, since that
+                # can cause divide-by-zero problems downstream).
```
ludwigschwardt: I like this idea for `applycal` too.

katdal/datasources.py
```diff
@@ -102,6 +103,80 @@ def _apply_data_lost(orig_flags, lost, block_id):
...
+        reuse the space across multiple calls). It must have the same shape
+        as `auto_indices` and dtype ``np.float32``.
+    """
+    auto_scale = np.empty(len(auto_indices), np.float32) if tmp is None else tmp
+    power_scale = np.empty(block.shape, np.float32) if out is None else out
+    bad_weight = np.float32(2.0**-32)
+    for i in range(block.shape[0]):
+        for j in range(block.shape[1]):
+            for k in range(len(auto_indices)):
+                auto_scale[k] = np.divide(1, block[i, j, auto_indices[k]].real)
```
ludwigschwardt: What is the difference between this and `1 / block...`?

katdal/datasources.py
```diff
@@ -102,6 +103,80 @@ def _apply_data_lost(orig_flags, lost, block_id):
     return flags
 
 
+def corrprod_to_autocorr(corrprod):
+    """Find the autocorrelation indices of correlation products.
+
+    Parameters
+    ----------
+    corrprod : sequence of 2-tuples or ndarray
```
ludwigschwardt: I would call this `corrprods` to match the rest of the code base.

katdal/datasources.py
```diff
@@ -144,6 +223,17 @@ def __init__(self, store, chunk_info):
...
         # Combine low-resolution weights and high-resolution weights_channel
         weights = darray['weights'] * darray['weights_channel'][..., np.newaxis]
+        # Scale weights according to power
+        if corrprods is not None:
+            assert len(corrprods) == vis.shape[2]
+            # Ensure that we have only a single chunk on the baseline axis.
+            if len(vis.chunks[2]) > 1:
+                vis = vis.rechunk({2: vis.shape[2]})
+            auto_indices, index1, index2 = corrprod_to_autocorr(corrprods)
+            power_scale = da.map_blocks(weight_power_scale, vis, dtype=np.float32,
```
ludwigschwardt: Again, a lot like `applycal`...

katdal/datasources.py
```diff
@@ -421,9 +511,13 @@ def __init__(self, telstate, chunk_store=None, timestamps=None,
             data = None
         else:
             chunk_info = telstate['chunk_info']
+            if True or telstate.get('need_weights_power_scale', False):
```
ludwigschwardt: The `True` is just for testing, right?

katdal/datasources.py
bmerry: Good question. The latter raises ZeroDivisionError (might be a numba quirk), but I want to deal with that lower down (np.isfinite check). I've added a comment to explain it.

katdal/datasources.py
bmerry: Makes sense. Should I keep it where it is for now and let you shuffle it around after this PR is merged?

ludwigschwardt: Yes, I think so. I still want to ponder the connections with `applycal` as well.

katdal/datasources.py
bmerry: Just the parameter, or do you prefer to rename the function too?

ludwigschwardt: I only thought of the parameter, but good question... Probably both (you could also rename the function `autocorr_indices` for more options :-))

katdal/datasources.py
ludwigschwardt: I get similar behaviour to `np.divide` with straight NumPy arrays, e.g.

```Python
1 / np.array(0)
```

The ZeroDivisionError only appears when NumPy is not involved:

```
1 / 0
```

ludwigschwardt:
Just checking whether np.divide is still needed (1 / np.array(...) seems to work too)...

bmerry:
> Just checking whether np.divide is still needed (1 / np.array(...) seems to work too)...

Nice idea. I've implemented that.

PR #191, opened 2018-09-25
Fix S3ChunkStore for Python 3.6

bmerry:
While testing on my home machine, I ran into unit test failures because
code inside Python 3.6's httplib does
```
if not chunk:
```
and if chunk is an ndarray then raises ValueError.

Wrap it in a memoryview so that truthiness is more sensible. I haven't
measured the performance impact, but the initial development of the
_Multipart optimisation used a memoryview and performed acceptably
well.


bmerry:
I'll re-run my benchmark when I'm back from leave, before merging.

bmerry:
I've tested and it doesn't appear to have harmed performance at all (if anything it's faster, but my sample size is very very small).

PR #192, opened 2018-10-08
Make SpectralWindow do channel juggling

bmerry:
Can now compute a new spectral window with a (contiguous) subset of
channels or a new channelisation of the bandwidth.

Closes SR-1369.

### DIFF
--- a/katdal/dataset.py
+++ b/katdal/dataset.py
@@ -140,19 +140,28 @@ class SpectralWindow(object):
         Type of downconversion (-1 => lower sideband, +1 => upper sideband)
     band : {'L', 'UHF', 'S', 'X', 'Ku'}, optional
         Name of receiver / band
+    bandwidth : float, optional
+        The bandwidth of the whole spectral window, in Hz. If specified,
+        `channel_width` is ignored and computed from the bandwidth. If not
+        specified, bandwidth is computed from the channel width. Specifying
+        this is a good idea if the channel width cannot be exactly represented
+        in floating point.
 
     Attributes
     ----------
     channel_freqs : array of float, shape (*F*,)
         Centre frequency of each frequency channel (assuming LSB mixing), in Hz
-    bandwidth : float
-        Total bandwidth across all channels
     """
 
     def __init__(self, centre_freq, channel_width, num_chans, product=None,
-                 sideband=-1, band='L'):
+                 sideband=-1, band='L', bandwidth=None):
+        if bandwidth is None:
+            bandwidth = channel_width * num_chans
+        else:
+            channel_width = bandwidth / num_chans
         self.centre_freq = centre_freq
         self.channel_width = channel_width
+        self.bandwidth = bandwidth
         self.num_chans = num_chans
         self.product = product if product is not None else ''
         self.sideband = sideband
@@ -166,8 +175,10 @@ class SpectralWindow(object):
         with self._channel_freqs_lock:
             if self._channel_freqs is None:
                 # Don't subtract half a channel width as channel 0 is centred on 0 Hz in baseband
-                self._channel_freqs = self.centre_freq + self.sideband * self.channel_width * (
-                    np.arange(self.num_chans) - self.num_chans // 2)
+                # We use self.bandwidth and self.num_chans to avoid rounding
+                # errors that might accumulate if channel_width is inexact.
+                self._channel_freqs = self.centre_freq + self.sideband * self.bandwidth * (
+                    np.arange(self.num_chans) - self.num_chans // 2) / self.num_chans
             return self._channel_freqs
 
     def __repr__(self):
@@ -177,20 +188,18 @@ class SpectralWindow(object):
                (self.band if self.band else 'unknown',
                 repr(self.product) if self.product else 'unknown',
                 self.centre_freq / 1e6,
-                self.num_chans * self.channel_width / 1e6,
+                self.bandwidth / 1e6,
                 self.num_chans, id(self))
 
     @property
     def _description(self):
         """Complete hashable representation, used internally for comparisons."""
         # Pick values that enable a sensible ordering of spectral windows
+        # Using self.bandwidth is generally redundant but may play a role in
+        # obscure rounding cases.
         return (self.centre_freq,
                 -self.channel_width, self.num_chans, self.sideband,
-                self.band, self.product)
-
-    @property
-    def bandwidth(self):
-        return self.channel_width * self.num_chans
+                self.band, self.product, -self.bandwidth)
 
     def __eq__(self, other):
         """Equality comparison operator."""
@@ -224,9 +233,15 @@ class SpectralWindow(object):
         if not (0 <= first < last <= self.num_chans):
             raise IndexError('channel indices out of range')
         channel_shift = (first + last) // 2 - self.num_chans // 2
+        num_chans = last - first
+        # We use self.bandwidth and self.num_chans to avoid rounding errors
+        # that might accumulate if channel_width is inexact.
+        centre_freq = self.centre_freq \
+            + channel_shift * self.bandwidth * self.sideband / self.num_chans,
         return SpectralWindow(
-            self.centre_freq + channel_shift * self.channel_width * self.sideband,
-            self.channel_width, last - first, self.product, self.sideband, self.band)
+            centre_freq, self.channel_width, num_chans,
+            self.product, self.sideband, self.band,
+            bandwidth=self.bandwidth * num_chans / self.num_chans)
 
     def rechannelise(self, num_chans):
         """Get a new :class:`SpectralWindow` with a different number of channels.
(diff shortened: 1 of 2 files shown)


katdal/dataset.py
```diff
@@ -190,6 +210,46 @@ def __hash__(self):
...
+        The returned :class:`SpectralWindow` covers the same frequencies as
+        channels [first, last) of the original.
+
+        Raises
+        ------
+        IndexError
+            If [first, last) is not a (non-empty) subinterval of the channels
+        """
+        if not (0 <= first < last <= self.num_chans):
+            raise IndexError('indices out of range')
```
ludwigschwardt: I would make this slightly more specific (e.g. `channel indices out of range`), since peeps tend not to look at the traceback but rather at the error message :-)

katdal/dataset.py
```diff
@@ -190,6 +210,46 @@ def __hash__(self):
         """Base hash on description tuple, just like equality operator."""
         return hash(self._description)
 
+    def subset(self, first, last):
```
ludwigschwardt: I prefer the name `slice` or `subrange` as `subset` seems too generic, but my conviction is relatively weak on this.

ludwigschwardt:
It is also almost time to split `SpectralWindow`, `Subarray`, etc out to their own modules or a `utilities` module, since `dataset.py` is quite cumbersome. I guess once the unit tests land :-)

katdal/dataset.py
```diff
@@ -190,6 +210,46 @@ def __hash__(self):
...
+        channels [first, last) of the original.
+
+        Raises
+        ------
+        IndexError
+            If [first, last) is not a (non-empty) subinterval of the channels
+        """
+        if not (0 <= first < last <= self.num_chans):
+            raise IndexError('indices out of range')
+        channel_shift = (first + last) // 2 - self.num_chans // 2
```
ludwigschwardt: In the old `katsdpingest` code you mention:

```
The implementation is careful to avoid introducing rounding errors where the answer is an exactly representable value.
```

Is this still true? (It seems to be)

bmerry: Yes and no. The new functions avoid introducing any rounding errors, but when constructing a SpectralWindow you have to compute channel_width as `bandwidth / num_chans` which can introduce rounding errors.

One idea I've had is to give the SpectralWindow constructor an optional `bandwidth=` arg, which supersedes channel_width. That would allow any calculations regarding bandwidth to use the exact bandwidth, rather than reconstituting it from the (possibly inexact) channel width. It's a somewhat theoretical point, since CBF always have power-of-two channel counts and hence the division is always exact.

katdal/dataset.py
bmerry: I was also not totally happy with subset. I considered slice, but I dislike functions that have the same name as a built-in type (if only because my editor gives misleading syntax highlighting in that case). Subrange sounds reasonable. Another option would be to allow `__getitem__` e.g. spw[first:last].

ludwigschwardt: I also like `__getitem__`, as long as strides raise an exception (in case users expect more than we can provide :-))

katdal/dataset.py
ludwigschwardt: I was going to make a similar suggestion... I'd be even more blunt and just change the API to expect a `bandwidth` parameter, since I don't *think* anyone else ever instantiated a `SpectralWindow` explicitly...

katdal/dataset.py
bmerry: I started implementing `__getitem__`, but then realised that might lead to expectations that one can do things like `spw[:-2]` which would complicate the implementation, so I'm going with subrange.

katdal/dataset.py
bmerry: Done.

mattieudv:
What is the big deal with providing strides? Isn't a better interim
solution to just simply load the whole array and then stride it, as opposed
to (the more challenging) only load strided data.

katdal/dataset.py
bmerry: I think @sjperkins would disagree: https://github.com/ska-sa/katsdppipelines/blob/0f594dd0a3962e880bc042e879b5c68692308b6c/katsdpcontim/katacomb/katacomb/mock_dataset.py#L369.

But since we both think the concept is good I'll work on my original suggestion.

katdal/test/test_dataset.py
```diff
@@ -0,0 +1,73 @@
...
+        usb = self.usb.rechannelise(2)
+        assert_array_equal(usb.channel_freqs, [980.0, 1010.0])
+
+    def test_rechannelise_to_odd(self):
+        lsb = self.lsb.rechannelise(3)
+        assert_array_equal(lsb.channel_freqs, [1025.0, 1005.0, 985.0])
+        usb = self.usb.rechannelise(3)
+        assert_array_equal(usb.channel_freqs, [975.0, 995.0, 1015.0])
+        odd = self.odd.rechannelise(1)
+        assert_array_equal(odd.channel_freqs, [1000.0])
```
ludwigschwardt: Thinking about this saga again, shouldn't this be 995.0? The complete band ranges from 965 to 1025 with 995 in the middle.

katdal/test/test_dataset.py
```diff
@@ -0,0 +1,73 @@
...
+
+class TestSpectralWindow(object):
+    def setUp(self):
+        self.lsb = SpectralWindow(1000.0, 10.0, 6, sideband=-1, product='lsb')
+        self.usb = SpectralWindow(1000.0, 10.0, 6, sideband=1, band='X')
+        self.odd = SpectralWindow(1000.0, 10.0, 5, sideband=1)
+
+    def test_channel_freqs(self):
+        assert_array_equal(self.lsb.channel_freqs,
+                           [1030.0, 1020.0, 1010.0, 1000.0, 990.0, 980.0])
```
ludwigschwardt: I like these tests very much! Nice and clear.

katdal/dataset.py
ludwigschwardt: Ah, forgot to check that power user :-)

bmerry:
@mattieudv a SpectralWindow represents a contiguous range of frequencies, so you can't skip over channels.

katdal/test/test_dataset.py
bmerry: No, it ranges from 975 to 1025, which makes 1000 the middle.

ludwigschwardt: Whoops, scratch that - I confused `self.odd` and `self.usb`.

ludwigschwardt:
I like what's there but feel free to add the `bandwidth` parameter as well.

katdal/dataset.py
bmerry: I've implemented something and pushed it, but it's impossible to exactly guarantee all the desirable properties at once (in particular, subrange should ideally preserve channel_width and the frequencies of the selected channels, while rechannelise should preserve bandwidth), so I'm not sure if it's really a win that justifies the extra code complexity. Let me know what you think.

katdal/dataset.py
```diff
@@ -224,9 +233,15 @@ def subrange(self, first, last):
...
+        # We use self.bandwidth and self.num_chans to avoid rounding errors
+        # that might accumulate if channel_width is inexact.
+        centre_freq = self.centre_freq \
+            + channel_shift * self.bandwidth * self.sideband / self.num_chans,
         return SpectralWindow(
-            self.centre_freq + channel_shift * self.channel_width * self.sideband,
-            self.channel_width, last - first, self.product, self.sideband, self.band)
+            centre_freq, self.channel_width, num_chans,
+            self.product, self.sideband, self.band,
+            bandwidth=self.bandwidth * num_chans / self.num_chans)
```
ludwigschwardt: If `subrange` should ideally preserve `channel_width`, why override it with a `bandwidth` parameter?

bmerry: Because it should also ideally compute an accurate bandwidth. For example, if the origin bandwidth is 1000Hz and you select half the channels, you should end up with 500Hz. If there were 14 channels and you preserved channel_width, you'd end up with a bandwidth of 1000/14*7, which might have rounding errors. I've mostly been favouring bandwidth over channel_width because that's what is stored in telstate.

ludwigschwardt: And will preserving `channel_width` help preserve the channel frequencies? A good question whether the latter is more NB than bandwidth or not.

bmerry: Not necessarily, because floating point is't distributive i.e. a channel frequency previously calculated as `old_center + i * channel_width` is now calculated as `(old_center + shift * channel_width) + (i - shift) * channel_width`. I think the only way to ensure that subrange doesn't alter channel frequencies is to store a reference channel and frequency and keep those the same in the return value, rather than always making the centre the reference.

ludwigschwardt: I see... I guess that's the rationale behind e.g. the FITS WCS keywords CRVALx and CRPIXx. More design choices to ponder...

PR #194, opened 2018-10-16
Refactor applycal, split out SpectralWindow and chuck RadosChunkStore

ludwigschwardt:
These are mostly in preparation of more "make L1" work.

- Let `applycal` handle multi-part cal products, and provide telstate access for that.
- Split out `SpectralWindow` into its own module to make it more amenable for use in `applycal`.
- Get rid of the obsolete `RadosChunkStore` (as well as the `h5toobj` script, which is ancient and potentially misleading by now).


katdal/__init__.py
```diff
@@ -240,7 +242,9 @@
 _dataset, _concatdata, _sensordata = dataset, concatdata, sensordata
 _h5datav1, _h5datav2, _h5datav3 = h5datav1, h5datav2, h5datav3
 _categorical, _lazy_indexer = categorical, lazy_indexer
-del dataset, concatdata, h5datav1, h5datav2, h5datav3, sensordata, categorical, lazy_indexer
+_spectral_window, _visdatav4 = spectral_window, visdatav4
+del dataset, concatdata, sensordata, h5datav1, h5datav2, h5datav3
+del categorical, lazy_indexer, spectral_window, visdatav4
```
bmerry: Can we just get rid of these? It leads to some bizarre things (see my [PyCon lightning talk](http://www.brucemerry.org.za/pycon2018/imports.slides.html#/2).

katdal/applycal.py
```diff
@@ -22,41 +22,66 @@
...
-    `cache`, then extracts the delay time series of the input specified by
-    `index` (in the form (pol, ant)) and builds a categorical sensor for
-    the corresponding complex correction terms (channelised by `freqs`).
+    This takes care of stitching together multiple parts of the product
+    if this is indicated in the `attrs` dict.
+    """
+    key = 'cal_product_' + product
+    try:
+        parts = int(attrs[key + '_parts'])
+    except (KeyError, TypeError):
```
bmerry: When would you get TypeError?

katdal/__init__.py
ludwigschwardt: Indeed, I've always wanted to... I'll make a new PR that removes the `del`s and rather put the public API in an `__all__` list.

katdal/applycal.py
ludwigschwardt: You should ask yourself - I copied that code from `LoaderKatdal._load_cal_product` :-)

I suspect you wanted to ensure that `cal_product_X_parts` is an int, but why would it not be? I'm in favour of just doing `KeyError` here if you are.

ludwigschwardt: Done.

PR #196, opened 2018-10-19
Use katsdptelstate.decode_value to decode telstate values

bmerry:
Telstate now provides a decode_value function to decode the stored
values. The values in future may cease to be pickles, so this
abstraction will then be necessary. A nice benefit even before switching
the encoding is that katdal no longer needs to duplicate the magic
incantations necessary to safely decode pickles written by Python 2, and
completely removes the pickle module from katdal.

### DIFF
--- a/katdal/h5datav3.py
+++ b/katdal/h5datav3.py
@@ -28,17 +28,14 @@ from collections import Counter
 import numpy as np
 import h5py
 import katpoint
-try:
-    import cPickle as pickle
-except ImportError:
-    import pickle
+import katsdptelstate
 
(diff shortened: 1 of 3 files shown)


bmerry:
@mserylak @marisageyer you might want to take note of this too for loading telescope state data from the beamformer HDF5 files. You can probably just use `katdal.sensordata.telstate_decode` rather than trying to keep up with changes in katsdptelstate.

mserylak:
@bmerry my thoughts exactly. I will test this soon and update relevant scripts (hopefully soon).

PR #199, opened 2018-10-31
Make the sensor cache safe for multiple readers

bmerry:
There were two issues:
- TelstateSensorData has an internal cache, which was not locked.
- SensorCache.get mutates its internal structures (in the case of
virtual sensors, that includes the set of keys present), again
without a lock.

Remove the bandaid on get_cal_product which should no longer be needed.

### DIFF
--- a/katdal/applycal.py
+++ b/katdal/applycal.py
@@ -21,7 +21,6 @@ from builtins import range, zip
 from functools import partial
 import copy
 import logging
-import threading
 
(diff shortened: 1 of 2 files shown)


ludwigschwardt:
This fixes the synchronisation issue mentioned in JIRA ticket [SR-1217](https://skaafrica.atlassian.net/browse/SR-1217).

PR #205, opened 2018-11-09
Expose S3 bucket name on data source

ludwigschwardt:
The `ChunkStoreVisFlagsWeights` class gains a `prefix` attribute
which corresponds to the S3 bucket name for the visibility data
(considered the main payload and therefore representative of the
data set). This aids the archive metadata extractor by removing the
need to scavenge around in `chunk_info`.

This addresses JIRA ticket SR-1401.

### DIFF
--- a/katdal/datasources.py
+++ b/katdal/datasources.py
@@ -194,12 +194,12 @@ class ChunkStoreVisFlagsWeights(VisFlagsWeights):
 
     Attributes
     ----------
-    prefix : string
-        Prefix of correlator_data / visibility array (i.e. the S3 bucket name)
+    vis_prefix : string
+        Prefix of correlator_data / visibility array, viz. its S3 bucket name
     """
     def __init__(self, store, chunk_info, corrprods):
         self.store = store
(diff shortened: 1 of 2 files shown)


katdal/datasources.py
```diff
@@ -191,9 +191,15 @@ class ChunkStoreVisFlagsWeights(VisFlagsWeights):
         Correlation products. If given, the weights for baseline (inp1, inp2)
         will be divided by the square root of the product of the corresponding
         autocorrelations vis[inp1,inp1] and vis[inp2,inp2].
+
+    Attributes
+    ----------
+    prefix : string
+        Prefix of correlator_data / visibility array (i.e. the S3 bucket name)
```
bmerry: I'd feel happier if we call this something like `correlator_data_prefix`, so that it's explicit that you shouldn't depend on using it for the other arrays. Otherwise the unwary might start using it like that, which would then be a source of future bugs e.g. code might accidentally use the L0 instead of the L1 flags.

PR #206, opened 2018-11-09
bugfix: open_data_source() swallowing an exception

adriaanph:
open_data_source() must not swallow DataSourceNotFound exception for URL's that don't have "file" protocol.


ludwigschwardt:
Good catch!

PR #207, opened 2018-11-23
Fix attribute lookup regression on H5 files

ludwigschwardt:
If an attribute in an HDF5v3 data set is not found in the `attrs` section of the `TelescopeState group`, one should check the sensors as well.

When #196 introduced telstate-level decoding, UnpicklingErrors was replaced by DecodeErrors but unfortunately the KeyError in this except clause got lost in the process.

This addresses JIRA ticket SR-1503.


bmerry:
Oops, I'm guessing that was my fault :-)

mserylak:
Tested last night. Works as advertised.

PR #209, opened 2019-01-09
Optionally 'switch off' l1-flags

KimMcAlpine:
Optionally don't include the l1-flags when opening the observation. This is useful when using the pipeline simulator to run on an mvf4 file, typically this is useful to debug problems with the pipeline. Switching off the l1 flags more faithfully represents the flags that would have been seen by the pipeline during the real observation.


KimMcAlpine:
I've made the suggested changes.
I also looked at just using the rdb-lite option as an input to the simulator, but it doesn't have the noise diode sensor information which I need for any replay of a delay-cal observation.

PR #213, opened 2019-01-29
Pass capture_block_id and stream_name to _upgrade_flags

bmerry:
Fixes SR-1584. This does make some breaking changes to the low-level
public API: view_l0_capture_stream returns the CBID and stream name, and
TelstateDataSource.__init__ takes them. I searched Github and didn't
find any code that calls these functions, so we are probably okay.

Also adds a very basic unit test that ensures the upgrade occurs even
without the telstate keys for capture_block_id/stream_name.

### DIFF
--- a/katdal/datasources.py
+++ b/katdal/datasources.py
@@ -335,6 +335,10 @@ def view_l0_capture_stream(telstate, capture_block_id=None, stream_name=None,
     -------
     telstate : :class:`~katdal.sensordata.TelstateToStr` object
         Telstate with a view that incorporates capture block, stream and combo
+    capture_block_id : string
+        Actual capture block ID used
+    stream_name : string
+        Actual L0 stream name used
 
     Raises
     ------
@@ -367,7 +371,7 @@ def view_l0_capture_stream(telstate, capture_block_id=None, stream_name=None,
                                                  expected_type))
     logger.info('Using capture block %r and stream %r',
                 capture_block_id, stream_name)
-    return telstate
+    return telstate, capture_block_id, stream_name
 
 
 def _shorten_key(telstate, key):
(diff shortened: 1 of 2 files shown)


katdal/datasources.py
```diff
@@ -497,6 +499,7 @@ def _upgrade_flags(chunk_info, telstate):
            stream_name not in telstate_cs['src_streams']:
             continue
         # Look for chunk metadata in appropriate capture_stream telstate view
+        logger.info('Upgrading flags to use %s instead of %s', s, stream_name)
```
ludwigschwardt: Nice.

katdal/datasources.py
```diff
@@ -518,6 +521,10 @@ class TelstateDataSource(DataSource):
     ----------
     telstate : :class:`katsdptelstate.TelescopeState` object
         Telescope state with appropriate views
+    capture_block_id : string
+        Capture block ID
+    stream_name : string
+        Name of the L0 stream
```
ludwigschwardt: Would this necessarily be L0 (just checking)?

katdal/test/test_datasources.py
```diff
@@ -187,3 +244,34 @@ def test_missing_chunks_uneven_chunking(self):
                 'weights_channel': (1, 7),
                 'flags': (4, 15, 30)
             })
+
+
+class TestTelstateDataSource(object):
+    def setup(self):
+        self.tempdir = tempfile.mkdtemp()
+        self.store = NpyFileChunkStore(self.tempdir)
```
ludwigschwardt: What about a `DictChunkStore` to avoid temp files? Although, NPY's are probably more representative...

katdal/datasources.py
bmerry: It's the same argument passed to (or returned from) `view_l0_capture_stream`, so I assume so.

katdal/test/test_datasources.py
bmerry: DictChunkStore is read-only, so wouldn't work with `put_fake_dataset`.

PR #214, opened 2019-01-29
Add more tests of _upgrade_flags

bmerry:
This is in preparation for SR-1330. It tests cases where the L1 flags
stream has a different number of dumps, giving 100% line coverage of
_upgrade_chunk_info.

### DIFF
--- a/katdal/test/test_datasources.py
+++ b/katdal/test/test_datasources.py
@@ -57,10 +57,10 @@ def to_dask_array(x, chunks=None):
 def put_fake_dataset(store, prefix, shape, chunk_overrides=None, array_overrides={}, flags_only=False):
     """Write a fake dataset into the chunk store."""
     if flags_only:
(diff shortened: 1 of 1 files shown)


katdal/test/test_datasources.py
```diff
@@ -57,10 +57,10 @@ def to_dask_array(x, chunks=None):
 def put_fake_dataset(store, prefix, shape, chunk_overrides=None, array_overrides={}, flags_only=False):
     """Write a fake dataset into the chunk store."""
     if flags_only:
-        data = {'flags': np.full(shape, 2, dtype=np.uint8)}
+        data = {'flags': np.random.randint(0, 7, shape, dtype=np.uint8)}
```
ludwigschwardt: Do you need to fix the random seed?

bmerry: Good idea. Done.

PR #215, opened 2019-01-29
Add missing import of dask.sharedict

bmerry:
In previous dask versions it was automatically imported, but now needs
to be imported explicitly. It is also deprecated in dask 1.1, but that's
a problem for another day.

### DIFF
--- a/katdal/chunkstore.py
+++ b/katdal/chunkstore.py
@@ -27,6 +27,7 @@ import io
 import numpy as np
 import dask
 import dask.array as da
+import dask.sharedict
 import toolz
 
 


bmerry:
Or add a dependency on dask>=1.1, so that when someone installs katdal they get upgraded? I guess we'll need to make sure dask doesn't cause any problems with our other packages though, particularly katdal. I don't remember the details but I also seem to recall that the lazy indexing had some subtleties in graph optimisation to ensure that indexing a small piece doesn't end up making copies of the full graph.

ludwigschwardt:
Sure, the dependency is clearly the way of the future, but I'm wary depending on a 12-day old version :-)

bmerry:
Fair enough.

PR #216, opened 2019-01-29
Change padding strategy for upgrade_chunk_info

bmerry:
The previous strategy was to take the number of dumps in the original
info as golden, and extend or truncate the improved info to fit. That
won't work so well with multi-dump chunks.

Instead, set the number of dumps to be the largest amongst all arrays
(after upgrades), and extend any array that is shorter to ensure a
match. That means that some files may now appear to have extra dumps for
which all the original data is missing, which will then be flagged.

Another side effect is that when not specifying a chunk-store OR
timestamps, the upgrade process is still done just to determine how many
dumps there are.

This should theoretically work with multi-dump chunks, but that has yet
to be tested.


bmerry:
@ludwigschwardt I'll add some unit tests for multi-dump chunks later, but I wanted to get an opinion on the approach first since if the semantics change I don't want to have to rewrite tests.

PR #217, opened 2019-01-30
Add TelstateDataSource tests with multi-dump chunks

bmerry:
This is hopefully the final piece of SR-1330.

### DIFF
--- a/katdal/test/test_datasources.py
+++ b/katdal/test/test_datasources.py
@@ -35,6 +35,9 @@ from katdal.chunkstore_npy import NpyFileChunkStore
 from katdal.datasources import ChunkStoreVisFlagsWeights, TelstateDataSource, view_l0_capture_stream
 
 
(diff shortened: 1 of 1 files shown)


katdal/test/test_datasources.py
```diff
@@ -54,7 +57,7 @@ def to_dask_array(x, chunks=None):
     return da.from_array(x, chunks)
 
 
-def put_fake_dataset(store, prefix, shape, chunk_overrides=None, array_overrides={}, flags_only=False):
+def put_fake_dataset(store, prefix, shape, chunk_overrides=None, array_overrides=None, flags_only=False):
```
ludwigschwardt: I know `None` is generally better than an empty container as a default value, but do you gain anything from using it in this function?

bmerry: Once I was adding similar arguments to a bunch of other functions and chaining them through I didn't want to have to remember which ones should default to `None` and which to `{}`.

PR #219, opened 2019-02-06
Add max_dim_elements option to generate_chunks

bmerry:
This will help with the spectral vis writer, allowing the number of
channels to be limited.

The algorithm was changed to determine an (integral) chunk shape, and then
divide everything into chunks of that shape (with smaller left-overs at
the end).

The behaviour changes slightly: when not using power_of_two, the chunks
are now all the same size except the last which is smaller (like with
power_of_two), rather than having chunks of size X+1 and X. This
simplifies the code, and I don't think !power_of_two is being used much
if at all any more.

### DIFF
--- a/katdal/chunkstore.py
+++ b/katdal/chunkstore.py
@@ -70,8 +70,7 @@ def generate_chunks(shape, dtype, max_chunk_size, dims_to_split=None,
         (the last chunk size along each dimension will potentially be smaller)
     max_dim_elements : dict, optional
         Maximum number of elements on each dimension (each key is a dimension
-        index). Behaviour is undefined if any dimensions are not also in
-        `dims_in_split`.
+        index). Dimensions that are not in `dims_to_split` are ignored.
 
     Returns
     -------
@@ -98,22 +97,21 @@ def generate_chunks(shape, dtype, max_chunk_size, dims_to_split=None,
     for dim in dims_to_split:
         if cur_elements <= max_elements:
             break      # We have already split enough to meet the budget
-        if dim >= len(shape):
-            continue
-        trg_size_real = dim_elements[dim] * max_elements / cur_elements
-        if trg_size_real < 1:
-            trg_size = 1
+        trg_elements_real = dim_elements[dim] * max_elements / cur_elements
+        if trg_elements_real < 1:
+            trg_elements = 1
         elif power_of_two:
-            trg_size = _floor_power_of_two(trg_size_real)
+            trg_elements = _floor_power_of_two(trg_elements_real)
         else:
             # Try to split into a number of equal-as-possible sized pieces
-            pieces = int(np.ceil(shape[dim] / trg_size_real))
+            pieces = int(np.ceil(shape[dim] / trg_elements_real))
             # Note: np.ceil rather than np.floor here means the max_chunk_size
             # could be breached. It's done like this for backwards
             # compatibility.
-            trg_size = int(np.floor(shape[dim] / pieces))
-        cur_elements = cur_elements // dim_elements[dim] * trg_size
-        dim_elements[dim] = trg_size
+            trg_elements = int(np.floor(shape[dim] / pieces))
+        # Update cur_elements to still be the product of dim_elements
+        cur_elements = cur_elements // dim_elements[dim] * trg_elements
+        dim_elements[dim] = trg_elements
 
     return da.core.blockdims_from_blockshape(shape, dim_elements)
 
--- a/katdal/chunkstore.py
+++ b/katdal/chunkstore.py
@@ -28,7 +28,6 @@ import numpy as np
 import dask
 import dask.array as da
 import dask.sharedict
-import toolz
 
 
 class ChunkStoreError(Exception):
@@ -52,8 +51,20 @@ def _floor_power_of_two(x):
     return 2 ** int(np.floor(np.log2(x)))
 
 
+def _generate_chunks_axis(length, chunk):
+    """Generate single-axis chunk specification.
+
+    It splits `length` into pieces of length `chunk`, possibly with one
+    smaller chunk left over.
+    """
+    if length % chunk == 0:
+        return (chunk,) * (length // chunk)
+    else:
+        return (chunk,) * (length // chunk) + (length % chunk,)
+
+
 def generate_chunks(shape, dtype, max_chunk_size, dims_to_split=None,
-                    power_of_two=False):
+                    power_of_two=False, max_chunk_sizes=None):
     """Generate dask chunk specification from ndarray parameters.
 
     Parameters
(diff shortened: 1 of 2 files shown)


katdal/chunkstore.py
```diff
@@ -77,32 +91,43 @@ def generate_chunks(shape, dtype, max_chunk_size, dims_to_split=None,
...
             continue
-        if num_chunks > shape[dim] / 2:
-            # Split the dimension into the maximum number of chunks
-            chunk_sizes = (1,) * shape[dim]
+        trg_size_real = chunk_sizes[dim] * max_elements / cur_elements
+        if trg_size_real < 1:
+            trg_size = 1
+        elif power_of_two:
+            trg_size = _floor_power_of_two(trg_size_real)
+            print(dim, trg_size_real, trg_size)
```
ludwigschwardt: Leftover debug print.

bmerry:
All good comments, thanks. Should be addressed in latest commit.

katdal/chunkstore.py
```diff
@@ -69,6 +80,10 @@ def generate_chunks(shape, dtype, max_chunk_size, dims_to_split=None,
     power_of_two : bool, optional
         True if chunk size should be rounded down to a power of two
         (the last chunk size along each dimension will potentially be smaller)
+    max_dim_elements : dict, optional
+        Maximum number of elements on each dimension (each key is a dimension
+        index). Behaviour is undefined if any dimensions are not also in
+        `dims_in_split`.
```
ludwigschwardt: dims_to_split

katdal/chunkstore.py
```diff
@@ -77,32 +92,43 @@ def generate_chunks(shape, dtype, max_chunk_size, dims_to_split=None,
...
+                dim_elements[i] = max_dim_elements[i]
+    cur_elements = int(np.prod(dim_elements))
+    # The ideal number of elements per chunk to achieve requested chunk size
+    # (can be float).
+    max_elements = max_chunk_size / np.dtype(dtype).itemsize
     # Split the array greedily along each dimension, in order of `dims_to_split`
     for dim in dims_to_split:
+        if cur_elements <= max_elements:
+            break      # We have already split enough to meet the budget
         if dim >= len(shape):
```
ludwigschwardt: This is not expected, right? Should you not rather do an `IndexError` or similar? Unless `dims_to_split` is merely a suggestion.

katdal/chunkstore.py
```diff
@@ -77,32 +92,43 @@ def generate_chunks(shape, dtype, max_chunk_size, dims_to_split=None,
...
-        chunks[dim] = chunk_sizes
-        # Update number of remaining chunks to realise
-        num_chunks /= len(chunk_sizes)
-    return tuple(chunks)
+            # Try to split into a number of equal-as-possible sized pieces
+            pieces = int(np.ceil(shape[dim] / trg_size_real))
+            # Note: np.ceil rather than np.floor here means the max_chunk_size
+            # could be breached. It's done like this for backwards
+            # compatibility.
+            trg_size = int(np.floor(shape[dim] / pieces))
```
ludwigschwardt: So `trg_size` is almost but not quite `int(trg_size_real)`?

katdal/chunkstore.py
```diff
@@ -69,6 +80,10 @@ def generate_chunks(shape, dtype, max_chunk_size, dims_to_split=None,
     power_of_two : bool, optional
         True if chunk size should be rounded down to a power of two
         (the last chunk size along each dimension will potentially be smaller)
+    max_dim_elements : dict, optional
+        Maximum number of elements on each dimension (each key is a dimension
+        index). Behaviour is undefined if any dimensions are not also in
```
ludwigschwardt: I would say dimensions not in `dims_to_split` are ignored.

katdal/chunkstore.py
```diff
@@ -52,8 +51,20 @@ def _floor_power_of_two(x):
     return 2 ** int(np.floor(np.log2(x)))
 
 
+def _generate_chunks_dim(length, chunk):
```
ludwigschwardt: Here is the dask version from 4 years ago (see dask/dask#57) ...
```Python
def blockdims_from_blockshape(shape, blockshape):
    """
     >>> blockdims_from_blockshape((10, 10), (4, 3))
    ((4, 4, 2), (3, 3, 3, 1))
    """
    return tuple((bd,) * (d // bd) + ((d % bd,) if d % bd else ())
                              for d, bd in zip(shape, blockshape))
```

katdal/chunkstore.py
```diff
@@ -77,32 +92,43 @@ def generate_chunks(shape, dtype, max_chunk_size, dims_to_split=None,
...
+        max_dim_elements = {}
+
+    dim_elements = list(shape)
+    for i in dims_to_split:
+        if i in max_dim_elements and max_dim_elements[i] < shape[i]:
+            if power_of_two:
+                dim_elements[i] = _floor_power_of_two(max_dim_elements[i])
+            else:
+                dim_elements[i] = max_dim_elements[i]
+    cur_elements = int(np.prod(dim_elements))
```
ludwigschwardt: So you start with the maximum possible number of elements and work your way down to `max_elements`? :-) I think of `max_elements` more like `trg_elements`.

katdal/chunkstore.py
```diff
@@ -77,32 +92,43 @@ def generate_chunks(shape, dtype, max_chunk_size, dims_to_split=None,
...
-        # Update number of remaining chunks to realise
-        num_chunks /= len(chunk_sizes)
-    return tuple(chunks)
+            # Try to split into a number of equal-as-possible sized pieces
+            pieces = int(np.ceil(shape[dim] / trg_size_real))
+            # Note: np.ceil rather than np.floor here means the max_chunk_size
+            # could be breached. It's done like this for backwards
+            # compatibility.
+            trg_size = int(np.floor(shape[dim] / pieces))
+        cur_elements = cur_elements // dim_elements[dim] * trg_size
```
ludwigschwardt: My head is still spinning because of all the multiplies and divides :-)

katdal/chunkstore.py
bmerry: I was just leaving in the semantics you wrote in the first place. But I'm happy to remove it. I don't think it's necessary to explicitly check it, since it's lead to an IndexError anyway.

katdal/chunkstore.py
bmerry: Yes. For example if the shape is 100 and trg_size_real is 60, then pieces is 2 and trg_size becomes 50, giving a (50, 50) split instead of a (60, 40) split. I think it ends up being more-or-less the same as the original code.

katdal/chunkstore.py
bmerry: This is just updating `cur_elements` so that it remains the product of `dim_elements`. I'll add a comment to that effect.

PR #223, opened 2019-02-12
Prepare for v0.12

ludwigschwardt:
And the last episode of the season.

### DIFF
--- a/NEWS.rst
+++ b/NEWS.rst
@@ -3,12 +3,12 @@ History
 
 0.12 (2019-02-12)
 -----------------
-* Make L1 product by applying calibration corrections aka applycal (#194 - #198)
+* Optionally make L1 product by applying calibration corrections (#194 - #198)
 * Let default reference antenna in v4 datasets be "array" antenna (#202, #220)
-* Use katsdptelstate v0.7 features: msgpack, memory backend (#196, #201, #212)
+* Use katsdptelstate v0.7: generic encodings, memory backend (#196, #201, #212)
 * Prepare for multi-dump chunks (#213, #214, #216, #217, #219)
 * Allow L1 flags to be ignored (#209, #210)
-* Fix usage of deprecated dask features (#204, #215)
+* Deal with deprecated dask features (#204, #215)
 * Remove RADOS chunk store (it's all via S3 from here on)
 
 0.11 (2018-10-15)
--- a/NEWS.rst
+++ b/NEWS.rst
@@ -1,6 +1,16 @@
 History
 =======
 
+0.12 (2019-02-12)
+-----------------
+* Make L1 product by applying calibration corrections aka applycal (#194 - #198)
+* Let default reference antenna in v4 datasets be "array" antenna (#202, #220)
+* Use katsdptelstate v0.7 features: msgpack, memory backend (#196, #201, #212)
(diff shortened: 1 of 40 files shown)


NEWS.rst
```diff
@@ -1,6 +1,16 @@
 History
 =======
 
+0.12 (2019-02-12)
+-----------------
+* Make L1 product by applying calibration corrections aka applycal (#194 - #198)
```
bmerry: This makes it sounds like it happens automatically, which might lead people astray.

NEWS.rst
```diff
@@ -1,6 +1,16 @@
...
+* Let default reference antenna in v4 datasets be "array" antenna (#202, #220)
+* Use katsdptelstate v0.7 features: msgpack, memory backend (#196, #201, #212)
```
bmerry: We're not using msgpack anywhere, just using the encode/decode helpers.

NEWS.rst
```diff
@@ -1,6 +1,16 @@
...
+* Prepare for multi-dump chunks (#213, #214, #216, #217, #219)
+* Allow L1 flags to be ignored (#209, #210)
+* Fix usage of deprecated dask features (#204, #215)
```
bmerry: I would say *some* deprecated dask features, since we're still using sharedict.

setup.py
```diff
@@ -18,12 +18,12 @@
 
 from __future__ import print_function, division, absolute_import
 
-import os.path
+import os
```
bmerry: Why?

setup.py
```diff
@@ -58,8 +58,8 @@
       python_requires='>=2.7,!=3.0,!=3.1,!=3.2',
       setup_requires=['katversion'],
       use_katversion=True,
-      install_requires=['numpy', 'katpoint', 'h5py>=2.3',
-                        'katsdptelstate[rdb]', 'dask[array] >= 0.18.2', 'numba',
+      install_requires=['numpy', 'katpoint', 'h5py >= 2.3', 'numba',
+                        'katsdptelstate[rdb] >= 0.7', 'dask[array] >= 0.18.2',
```
bmerry: Does it still work with dask 0.18.2, or are we using any newer features?

setup.py
ludwigschwardt: There is some [confusion](https://stackoverflow.com/questions/2724348) around this. I somehow mistakenly recalled that Python 3 encouraged `os` above `os.path`. I'm happy to change it back - more explicit that way.

bmerry: I would change it back, but I'm not too fussed.

NEWS.rst
ludwigschwardt: Fixed.

NEWS.rst
ludwigschwardt: Sure.

NEWS.rst
ludwigschwardt: Reworded.

setup.py
ludwigschwardt: Changed back.

setup.py
ludwigschwardt: I can't see any newer features. I am still keen to resolve the `HighLevelGraph` warning soon for the record, as I don't like scaring the users.

PR #227, opened 2019-02-20
Add rechunker script

bmerry:
It currently only supports output to local filesystem (and read from S3
isn't tested either), and should be considered an engineering
development tool rather than a user tool (which is why it doesn't get
installed by default).

### DIFF
--- a/katdal/datasources.py
+++ b/katdal/datasources.py
@@ -601,6 +601,8 @@ class TelstateDataSource(DataSource):
             timestamps = t0 + np.arange(n_dumps) * int_time
         # Metadata and timestamps with or without data
         DataSource.__init__(self, metadata, timestamps, data)
+        self.capture_block_id = capture_block_id
+        self.stream_name = stream_name
 
     @classmethod
     def from_url(cls, url, chunk_store='auto', upgrade_flags=True, **kwargs):
--- a/katdal/datasources.py
+++ b/katdal/datasources.py
@@ -464,8 +464,8 @@ def _align_chunk_info(chunk_info):
     return chunk_info
 
 
-def _infer_chunk_store(url_parts, telstate, npy_store_path=None,
-                       s3_endpoint_url=None, **kwargs):
+def infer_chunk_store(url_parts, telstate, npy_store_path=None,
+                      s3_endpoint_url=None, array='correlator_data', **kwargs):
     """Construct chunk store automatically from dataset URL and telstate.
 
     Parameters
@@ -478,6 +478,8 @@ def _infer_chunk_store(url_parts, telstate, npy_store_path=None,
         Top-level directory of NpyFileChunkStore (overrides the default)
     s3_endpoint_url : string, optional
         Endpoint of S3 service, e.g. 'http://127.0.0.1:9000' (overrides default)
+    array : string, optional
+        Array within the bucket from which to determine the prefix
     kwargs : dict, optional
         Extra keyword arguments, typically meant for other methods and ignored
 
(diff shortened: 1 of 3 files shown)


scripts/mvf_rechunk.py
```diff
@@ -0,0 +1,231 @@
...
+    def _make_lost(self, block_info):
+        loc = block_info['array-location']
+        shape = [l[1] - l[0] for l in loc]
+        if self.has_data[block_info['chunk-location']]:
+            return np.zeros(shape, np.uint8)
+        else:
+            return np.full(shape, DATA_LOST, np.uint8)
+
+
+def get_chunk_store(source, telstate, array):
```
ludwigschwardt: What's the main difference - simpler interface? separate store per array?

scripts/mvf_rechunk.py
```diff
@@ -0,0 +1,231 @@
...
+            # Older files have dtype as an object that can't be encoded in msgpack
+            dtype = np.dtype(array_info['dtype'])
+            array_info['dtype'] = np.lib.format.dtype_to_descr(dtype)
+            arrays[(stream_name, array_name)] = Array(stream_name, array_name, store, array_info)
+
+    # Apply DATA_LOST bits to the flags arrays. This is a less efficient approach than
+    # datasources.py, but much simpler.
+    for stream_name in streams:
+        flags_array = arrays.get((stream_name, 'flags'))
+        if flags_array:
```
ludwigschwardt: I prefer the following to reduce indentation (I see it's already used to good effect further into the loops):

```Python
if not flags_array:
    continue
```

katdal/datasources.py
```diff
@@ -601,6 +601,8 @@ def __init__(self, telstate, capture_block_id, stream_name,
             timestamps = t0 + np.arange(n_dumps) * int_time
         # Metadata and timestamps with or without data
         DataSource.__init__(self, metadata, timestamps, data)
+        self.capture_block_id = capture_block_id
```
ludwigschwardt: Looks useful in general.

scripts/mvf_rechunk.py
```diff
@@ -0,0 +1,231 @@
...
+
+    return streams
+
+
+def main():
+    args = parse_args()
+    dask.config.set(num_workers=args.workers)
+
+    # Lightweight open with no data - just to create telstate and identify the CBID
+    ds = TelstateDataSource.from_url(args.source, upgrade_flags=False, chunk_store=None)
```
ludwigschwardt: Glad to see the `upgrade_flags` param in use!

scripts/mvf_rechunk.py
```diff
@@ -0,0 +1,231 @@
...
+    url_parts = urllib.parse.urlparse(args.source, scheme='file')
+    dest_file = os.path.join(args.dest, args.new_prefix or cbid, os.path.basename(url_parts.path))
+    os.makedirs(os.path.dirname(dest_file), exist_ok=True)
+    writer = RDBWriter(client=telstate.backend)
+    writer.save(dest_file)
+
+
+if __name__ == '__main__':
+    try:
+        main()
```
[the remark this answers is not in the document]
bmerry: I've always done it because [Guido does](https://www.artima.com/weblogs/viewpost.jsp?thread=4829). Apart from the advice there, I found another advantage on StackOverflow: it keeps the variables used by the main code local to `main`, so that they don't accidentally leak into other functions.

scripts/mvf_rechunk.py
bmerry: Both I guess.

The code duplication is a bit of a smell though. I'm thinking perhaps I should make `_infer_chunk_store` more generic (take `array` argument that defaults to `correlator_data`) and public, and just call it from here.

scripts/mvf_rechunk.py
bmerry: Done.

scripts/mvf_rechunk.py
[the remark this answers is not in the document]
ludwigschwardt: Ah, so Guido talks about "a main() function that's easy to invoke in other contexts, e.g. from the interactive Python prompt when you feel like experimenting." We don't really import `main` elsewhere.

And the localised variables are exactly why I *don't* like `main()` 😀 Instead of importing `main()` in IPython, I prefer to `%run` the script and then inspect the local namespace after the script is done for debugging and enlightenment. Since we don't import scripts, for me the risk of leakage is low while the convenience of non-main is high...

Just an opinion, happy to keep this as is.

PR #229, opened 2019-02-22
Optimise reads from S3

bmerry:
A variety of nasty tricks to make requests go faster.

### DIFF
--- a/katdal/chunkstore_s3.py
+++ b/katdal/chunkstore_s3.py
@@ -72,6 +72,36 @@ _BUCKET_POLICY = {
 }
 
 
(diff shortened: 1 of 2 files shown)


bmerry:
This should be tested against an HTTPS server before merging; @ctgschollar is getting the HTTPS proxy set up again.

bmerry:
I've tested it against the HTTPS proxy and it still works, although performance doesn't really improve (it's about 200-250MB/s and shows poor scaling). It's not clear if that's a limitation of the proxy itself or of the Python SSL implementation.

PR #231, opened 2019-03-01
mvftoms: put weights into WEIGHT_SPECTRUM column

bmerry:


bmerry:
@sjperkins I'm creating the WEIGHT_SPECTRUM column by cargo-culting the existing code. If I'm Doing It Wrong(tm) let me know.

PR #232, opened 2019-03-04
Inclusion of AVN datafile format.

james-smith-za:
Inclusion of AVN datafile format for reduction of single-dish observations in Ghana.

### DIFF
--- a/katdal/h5datav2_5.py
+++ b/katdal/h5datav2_5.py
@@ -1,0 +1,631 @@
+"""Data accessor class for HDF5 files produced by AVN DBE."""
+
+import logging
+
+import numpy as np
+import h5py
+import katpoint
+
+from .dataset import DataSet, WrongVersion, BrokenFile, Subarray, \
+    DEFAULT_SENSOR_PROPS, DEFAULT_VIRTUAL_SENSORS, _robust_target
+from .spectral_window import SpectralWindow
+from .sensordata import RecordSensorData, SensorCache
+from .categorical import CategoricalData, sensor_to_categorical
+from .lazy_indexer import LazyIndexer, LazyTransform
+
+logger = logging.getLogger(__name__)
+
+# Simplify the scan activities to derive the basic state of the antenna (slewing, scanning, tracking, stopped)
+SIMPLIFY_STATE = {'scan_ready': 'slew', 'scan': 'scan', 'scan_complete': 'scan', 'track': 'track', 'slew': 'slew'}
+
+SENSOR_PROPS = dict(DEFAULT_SENSOR_PROPS)
+SENSOR_PROPS.update({
+    '*activity': {'greedy_values': ('slew', 'stop'), 'initial_value': 'slew',
+                  'transform': lambda act: SIMPLIFY_STATE.get(act, 'stop')},
+    '*target': {'initial_value': '', 'transform': _robust_target},
+    # These float sensors are actually categorical by nature as they represent user settings
+    'RFE/center-frequency-hz': {'categorical': True},
+    'RFE/rfe7.lo1.frequency': {'categorical': True},
+    '*attenuation': {'categorical': True},
+    '*attenuator.horizontal': {'categorical': True},
+    '*attenuator.vertical': {'categorical': True},
+    'RFE/rfe.band.select.LCP': {'categorical': True},
+    'RFE/rfe.band.select.RCP': {'categorical': True},
+    'RFE/rfe.lo-intermediate.5GHz.frequency': {'categorical': True},
+    'RFE/rfe.lo-intermediate.6_7GHz.frequency': {'categorical': True},
+    'RFE/rfe.lo-final.frequency': {'categorical': True}
+})
+
+SENSOR_ALIASES = {
+    'nd_roach': 'roach.noise.diode.on',
+}
+
+def _calc_azel(cache, name, ant="ant1"):
+    """Calculate virtual (az, el) sensors from actual ones in sensor cache."""
+    real_sensor = 'Antennas/%s/%s' % (ant, 'pos.actual-scan-azim' if name.endswith('az') else 'pos.actual-scan-elev')
+    cache[name] = sensor_data = katpoint.deg2rad(cache.get(real_sensor))
+    return sensor_data
+
+VIRTUAL_SENSORS = dict(DEFAULT_VIRTUAL_SENSORS)
+VIRTUAL_SENSORS.update({'Antennas/{ant}/az': _calc_azel, 'Antennas/{ant}/el': _calc_azel})
+
+FLAG_NAMES = ('reserved0', 'static', 'cam', 'reserved3', 'detected_rfi', 'predicted_rfi', 'reserved6', 'reserved7')
+FLAG_DESCRIPTIONS = ('reserved - bit 0', 'predefined static flag list', 'flag based on live CAM information',
+                     'reserved - bit 3', 'RFI detected in the online system', 'RFI predicted from space based pollutants',
+                     'reserved - bit 6', 'reserved - bit 7')
+
+#--------------------------------------------------------------------------------------------------
+#--- Utility functions
+#--------------------------------------------------------------------------------------------------
+
+
+def make_pmodel_string(param_list):
+    """Take the h5 dataset that stores the AVN pointing model and convert it to a
+       string in the format that katpoint expects.
+    """
+    if not isinstance(param_list, h5py.Dataset):
+        raise ValueError("Error! param_list isn't an HDF5 dataset.")
+    model_string = ""
+    for param in param_list:
+        model_string += str(param) + " "
+    model_string = model_string[:-1]  # To take off the space on the end.
+    return model_string
+
+
+def get_single_value(group, name):
+    """Return single value from attribute or dataset with given name in group.
+
+    If `name` is an attribute of the HDF5 group `group`, it is returned,
+    otherwise it is interpreted as an HDF5 dataset of `group` and the last value
+    of `name` is returned. This is meant to retrieve static configuration values
+    that potentially get set more than once during capture initialisation, but
+    then does not change during actual capturing.
+
+    Parameters
+    ----------
+    group : :class:`h5py.Group` object
+        HDF5 group to query
+    name : string
+        Name of HDF5 attribute or dataset to query
+
+    Returns
+    -------
+    value : object
+        Attribute or last value of dataset
+
+    """
+    return group.attrs[name] if name in group.attrs else group[name][-1]
+
+
+def dummy_dataset(name, shape, dtype, value):
+    """Dummy HDF5 dataset containing a single value.
+
+    This creates a dummy HDF5 dataset in memory containing a single value. It
+    can have virtually unlimited size as the dataset is highly compressed.
+
+    Parameters
+    ----------
+    name : string
+        Name of dataset
+    shape : sequence of int
+        Shape of dataset
+    dtype : :class:`numpy.dtype` object or equivalent
+        Type of data stored in dataset
+    value : object
+        All elements in the dataset will equal this value
+
+    Returns
+    -------
+    dataset : :class:`h5py.Dataset` object
+        Dummy HDF5 dataset
+
+    """
+    # It is important to randomise the filename as h5py does not allow two writable file objects with the same name
+    # Without this randomness katdal can only open one file requiring a dummy dataset
+    random_string = ''.join(['%02x' % (x,) for x in np.random.randint(256, size=8)])
+    dummy_file = h5py.File('%s_%s.h5' % (name, random_string), driver='core', backing_store=False)
+    return dummy_file.create_dataset(name, shape=shape, maxshape=shape, dtype=dtype, fillvalue=value, compression='gzip')
+
+#--------------------------------------------------------------------------------------------------
+#--- CLASS :  H5DataV2_5
+#--------------------------------------------------------------------------------------------------
+
+
+class H5DataV2_5(DataSet):
+    """Load HDF5 format version 2.5 file produced by AVN DBE.
+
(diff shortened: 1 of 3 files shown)


ludwigschwardt:
Excellent!

katdal/h5datav2_5.py
```diff
@@ -0,0 +1,631 @@
...
+def _calc_azel(cache, name, ant="ant1"):
+    """Calculate virtual (az, el) sensors from actual ones in sensor cache."""
+    real_sensor = 'Antennas/%s/%s' % (ant, 'pos.actual-scan-azim' if name.endswith('az') else 'pos.actual-scan-elev')
+    cache[name] = sensor_data = katpoint.deg2rad(cache.get(real_sensor))
+    return sensor_data
+
+VIRTUAL_SENSORS = dict(DEFAULT_VIRTUAL_SENSORS)
+VIRTUAL_SENSORS.update({'Antennas/{ant}/az': _calc_azel, 'Antennas/{ant}/el': _calc_azel})
+
+FLAG_NAMES = ('reserved0', 'static', 'cam', 'reserved3', 'detected_rfi', 'predicted_rfi', 'reserved6', 'reserved7')
```
ludwigschwardt: The flags have moved to a separate `.flags` module but you could always clear this up afterwards if you wish.

katdal/h5datav2_5.py
```diff
@@ -0,0 +1,631 @@
...
+def make_pmodel_string(param_list):
+    """Take the h5 dataset that stores the AVN pointing model and convert it to a
+       string in the format that katpoint expects.
+    """
+    if not isinstance(param_list, h5py.Dataset):
+        raise ValueError("Error! param_list isn't an HDF5 dataset.")
+    model_string = ""
+    for param in param_list:
+        model_string += str(param) + " "
+    model_string = model_string[:-1]  # To take off the space on the end.
```
ludwigschwardt: A more compact way would be

```Python
model_string = ' '.join(str(param) for param in param_list)
```

which pretty much removes the need for this function.

katdal/h5datav2_5.py
```diff
@@ -0,0 +1,631 @@
...
+        accumulation_length = get_single_value(config_group['DBE'], 'accum_length')  # Accumulation length in number of FPGA frames.
+        coarse_size = get_single_value(config_group["DBE"], "dbe.fft.coarse.size")
+        fine_size = get_single_value(config_group["DBE"], "dbe.fft.fine.size")
+        sampling_frequency = 800e6  # Have to hardcode this for the time being.
+        self.dump_period  = float(accumulation_length) * coarse_size / sampling_frequency
+        if fine_size != 0:
+            self.dump_period *= fine_size
+
+        # Obtain visibility data and timestamps
+        self._vis         = data_group['VisData'][:-1,:,:]
```
ludwigschwardt: Why drop the last data point - duplicate DBE output like in KAT7?

katdal/h5datav2_5.py
```diff
@@ -0,0 +1,631 @@
...
+        expected_dumps = (self._timestamps[num_dumps - 1] - self._timestamps[0]) / self.dump_period + 1
+        # The expected_dumps should always be an integer (like num_dumps), unless the timestamps and/or dump period
+        # are messed up in the file, so the threshold of this test is a bit arbitrary (e.g. could use > 0.5)
+        irregular = abs(expected_dumps - num_dumps) >= 0.01
+        if irregular:
+            # Warn the user, as this is anomalous
+            logger.warning(("Irregular timestamps detected in file '%s': "
+                           "expected %.3f dumps based on dump period and start/end times, got %d instead") %
+                           (filename, expected_dumps, num_dumps))
+            if quicklook:
```
ludwigschwardt: Do you still need quicklook for AVN? In other words, are your timestamps interspersed throughout the HDF5 file or written as a single dataset at the end?

katdal/h5datav2_5.py
```diff
@@ -0,0 +1,631 @@
...
+        self.ref_ant = script_ants[0] if not ref_ant else ref_ant
+        # Original list of correlation products as pairs of input labels
+        corrprods = get_single_value(config_group["DBE"], "vis_ordering").split(',')
+        if len(corrprods) != self._vis.shape[2]:
+            raise BrokenFile('Number of data labels (containing expected antenna names) '
+                             'received from h5 file (%d) differs from number of power products in data (%d)' %
+                             (len(corrprods), self._vis.shape[2]))
+        # Get the corrprod labels into the format that KatDAL wants, the v2.5 files only give ll and rr.
+        # This will change it into  "ant1lant1l,ant1rant1r" which katdal expects.
+        corrprods = [('ant1' + corrprods[0][0], 'ant1' + corrprods[0][1]),
```
ludwigschwardt: Is it always 'ant1'? Maybe better to do a constant at the top for this.

katdal/h5datav2_5.py
```diff
@@ -0,0 +1,631 @@
...
+            diameter = config_group["Antennas"][antenna].attrs['diameter']
+            delay_model = None
+            pointing_model = make_pmodel_string(config_group["Antennas"][antenna]['pointing-model-params'])
+            beamwidth = config_group["Antennas"][antenna].attrs['beamwidth']
+
+            ants.append(katpoint.Antenna(name, latitude, longitude, altitude, diameter, delay_model, pointing_model, beamwidth))
+
+        # Temporarily casting this to string. I edited a file by hand to change the name and it seems to now read the
+        # attr as a np.ndarray
+        #ants = [katpoint.Antenna(str(config_group['Antennas'][name].attrs['description']))
```
ludwigschwardt: You should remove commented out code if not applicable anymore.

katdal/h5datav2_5.py
```diff
@@ -0,0 +1,631 @@
...
+    '*activity': {'greedy_values': ('slew', 'stop'), 'initial_value': 'slew',
+                  'transform': lambda act: SIMPLIFY_STATE.get(act, 'stop')},
+    '*target': {'initial_value': '', 'transform': _robust_target},
+    # These float sensors are actually categorical by nature as they represent user settings
+    'RFE/center-frequency-hz': {'categorical': True},
+    'RFE/rfe7.lo1.frequency': {'categorical': True},
+    '*attenuation': {'categorical': True},
+    '*attenuator.horizontal': {'categorical': True},
+    '*attenuator.vertical': {'categorical': True},
+    'RFE/rfe.band.select.LCP': {'categorical': True},
```
ludwigschwardt: Are these really floats?

katdal/h5datav2_5.py
```diff
@@ -0,0 +1,631 @@
...
+            if band_select[i] == "0":
+                centre_freq[i] = centre_freq_5GHz[i]
+            elif band_select[i] == "1":
+                centre_freq[i] = centre_freq_6p7GHz[i]
+            else:
+                raise BrokenFile("Unknown band selection, seems to be neither 5 GHz nor 6.7 GHz.")
+
+        # If we aren't in wideband mode, then we need to do some additional tweaking:
+        if (fine_size != 0):
+            centre_freq -= 200e6  # Because we're going to count from the bottom of the band.
```
ludwigschwardt: It's safer to use fractions of `sampling_frequency` here.

katdal/h5datav2_5.py
```diff
@@ -0,0 +1,631 @@
...
+        Returns
+        -------
+        targets : :class:'katpoint.Catalogue' object
+            All targets in file
+
+        """
+        f, version = H5DataV2_5._open(filename)
+        # Use the delay-tracking centre as the one and only target
+        # Try two different sensors for the DBE target
+        target_list = f['MetaData/Sensors/Antennas/ant1/target']
```
ludwigschwardt: This looks like the pointing centre, not the delay-tracking centre.

katdal/h5datav2_5.py
```diff
@@ -0,0 +1,631 @@
...
+        sampling_frequency = 800e6  # Have to hardcode this for the time being.
+        self.dump_period  = float(accumulation_length) * coarse_size / sampling_frequency
+        if fine_size != 0:
+            self.dump_period *= fine_size
+
+        # Obtain visibility data and timestamps
+        self._vis         = data_group['VisData'][:-1,:,:]
+        self._stokes      = data_group['StokesData']
+        self._timestamps  = data_group['Timestamps']
+        self._time_av_ll  = data_group["Left Power time average"]
```
ludwigschwardt: Out of curiosity, how are these averages calculated?

katdal/h5datav2_5.py
```diff
@@ -0,0 +1,631 @@
...
+    @property
+    def time_av_u(self):
+        def _extract_time_av_u(u, keep):
+            return u.astype(np.float64)
+        extract_time_av_u = LazyTransform('extract_time_av_u', _extract_time_av_u, lambda shape: shape, np.float64)
+        return LazyIndexer(self._time_av_u, transforms=[extract_time_av_u])
+
+    @property
+    def u(self):
+        raise NotImplementedError("File is AVN Single-dish format, uv-plane is not applicable.")
```
ludwigschwardt: Or you could just return 0 :-)

katdal/h5datav2_5.py
james-smith-za: We don't actually need flags, that was just something that was in h5datav2.py which I copied previously. I will remove the line, I don't think I use it.

katdal/h5datav2_5.py
james-smith-za: Yes, the way that the data from the ROACH is ingested at the moment inserts a line of zeros into the file. It's challenging to remove from that end, I found it easier just to ignore it on this end.

katdal/h5datav2_5.py
james-smith-za: I remember you talked about this. It's written as a single datafile at the end. Should this just be removed?

katdal/h5datav2_5.py
james-smith-za: Noted.

katdal/h5datav2_5.py
james-smith-za: This might be some left-overs from the old v2 file that I copied. I might need to check that I understand properly what's going on there.

katdal/h5datav2_5.py
ludwigschwardt: Then also remove anything mentioning flags. The default `DataSet.flags` property will raise `NotImplementedError` so as long as you don't access it you should be fine.

The same goes for any mentions for weights...

katdal/h5datav2_5.py
ludwigschwardt: Our original HDF5 files took a long while to open with katdal (if very large), mostly because the timestamps have to be read and the timestamps were interspersed throughout the file. In v3 we wrote the timestamps as a single contiguous dataset just before closing the file, which allowed katdal to open the file quickly.

So if your files already open "quickly" (say within 1-2 seconds) with `katdal.open`, I would remove the whole quicklook machinery.

katdal/h5datav2_5.py
ludwigschwardt: Also, what if the antenna name in the file attributes differs from 'ant1' (not likely but plausible)?

katdal/h5datav2_5.py
james-smith-za: Would that be valid? I'm concerned that if a user unwittingly uses this, it may cause unexpected weird results.

ludwigschwardt: It should be. That's where single-dish data goes...

katdal/h5datav2_5.py
james-smith-za: The ingest software just averages across the entire spectrum for each dump, keeps it in memory and writes the entire vector as an H5 dataset at the same time as it's writing timestamps, etc. It's just a rough-and-ready piece of data to use for quick checks of the H5 file if you don't have KatDAL handy.

katdal/h5datav2_5.py
james-smith-za: Okay. Would that be true of u, v and w? Or just u?

katdal/h5datav2_5.py
james-smith-za: Done, commit 04bc711.

katdal/h5datav2_5.py
james-smith-za: Good suggestion, thanks. Fixed in commit c4255d.

katdal/h5datav2_5.py
ludwigschwardt: All of them.

katdal/h5datav2_5.py
james-smith-za: Removed in d8ea341.

katdal/h5datav2_5.py
james-smith-za: Done in ca0e58d. Have I understood correctly?

katdal/h5datav2_5.py
james-smith-za: Corrected in 9246136.

katdal/h5datav2_5.py
james-smith-za: Noted. Changed in 56e9a20.

katdal/h5datav2_5.py
james-smith-za: Noted. Removed in 51587b5.

katdal/h5datav2_5.py
james-smith-za: Reworked in b6ff28a.

katdal/h5datav2_5.py
james-smith-za: Corrected in c2daa49.

james-smith-za:
@ludwigschwardt I realise this is not a priority right now, but this PR has been dangling for quite a while.

I don't really have an opinion either way - if more work is needed to clean it up, or if it should just be closed. What are your thoughts?

PR #233, opened 2019-03-08
Load RDB files straight from archive

ludwigschwardt:
If given an HTTP URL, interpret it as the location of an RDB dump file in an S3 object store. Use (or abuse) an S3ChunkStore object to access it, which handles the necessary authorization. Download the RDB object as a binary blob into memory and open it directly using the latest katsdptelstate. Then also use this S3ChunkStore as the actual chunk store in the absence of any overrides.

To enable the (ab)use of `S3ChunkStore`, upgrade the `_request` method to its public API by dropping the underscore.

This addresses JIRA ticket [SR-1443](https://skaafrica.atlassian.net/browse/SR-1443).

### DIFF
--- a/katdal/datasources.py
+++ b/katdal/datasources.py
@@ -36,13 +36,14 @@ import numba
 from .sensordata import TelstateSensorData, TelstateToStr
 from .chunkstore_s3 import S3ChunkStore
 from .chunkstore_npy import NpyFileChunkStore
+from .chunkstore import StoreUnavailable
 from .flags import DATA_LOST
 
 
 logger = logging.getLogger(__name__)
 
 
-class DataSourceNotFound(Exception):
+class DataSourceNotFound(OSError):
     """File associated with DataSource not found or server not responding."""
 
 
@@ -623,42 +624,48 @@ class TelstateDataSource(DataSource):
         kwargs : dict, optional
             Extra keyword arguments passed to telstate view and chunk store init
         """
-        url_parts = urllib.parse.urlparse(url, scheme='file')
+        u = urllib.parse.urlparse(url, scheme='file')
         # Merge key-value pairs from URL query with keyword arguments
         # of function (the latter takes precedence)
-        url_kwargs = dict(urllib.parse.parse_qsl(url_parts.query))
+        url_kwargs = dict(urllib.parse.parse_qsl(u.query))
         url_kwargs.update(kwargs)
         kwargs = url_kwargs
         # Extract Redis database number if provided
         db = int(kwargs.pop('db', '0'))
-        if url_parts.scheme == 'file':
+        if u.scheme == 'file':
             # RDB dump file
             telstate = katsdptelstate.TelescopeState(katsdptelstate.memory.MemoryBackend())
             try:
-                telstate.load_from_file(url_parts.path)
-            except OSError as err:
-                raise DataSourceNotFound(str(err))
-        elif url_parts.scheme == 'redis':
+                telstate.load_from_file(u.path)
+            except OSError as e:
+                raise DataSourceNotFound(str(e))
+        elif u.scheme == 'redis':
             # Redis server
             try:
-                telstate = katsdptelstate.TelescopeState(url_parts.netloc, db)
+                telstate = katsdptelstate.TelescopeState(u.netloc, db)
             except katsdptelstate.ConnectionError as e:
                 raise DataSourceNotFound(str(e))
-        elif url_parts.scheme in {'http', 'https'}:
-            # Open endpoint as an S3 object store (with auth info in kwargs)
-            endpoint = urllib.parse.urlunparse(url_parts[:2] + 4 * ('',))
-            rdb_store = S3ChunkStore.from_url(endpoint, **kwargs)
-            rdb_path = urllib.parse.urlunparse(url_parts[:3] + 3 * ('',))
+        elif u.scheme in {'http', 'https'}:
+            # Treat endpoint as an S3 object store (with auth info in kwargs)
+            store_url = (u.scheme, u.netloc, '', '', '', '')
+            store_url = urllib.parse.urlunparse(store_url)
+            # Strip off parameters, query strings and fragments to get basic URL
+            rdb_url = (u.scheme, u.netloc, u.path, '', '', '')
+            rdb_url = urllib.parse.urlunparse(rdb_url)
             telstate = katsdptelstate.TelescopeState(katsdptelstate.memory.MemoryBackend())
-            with rdb_store.request('RDB', 'GET', rdb_path) as response:
-                telstate.load_from_file(io.BytesIO(response.content))
+            try:
+                rdb_store = S3ChunkStore.from_url(store_url, **kwargs)
+                with rdb_store.request('', 'GET', rdb_url) as response:
+                    telstate.load_from_file(io.BytesIO(response.content))
+            except StoreUnavailable as e:
+                raise DataSourceNotFound(str(e))
             if chunk_store == 'auto' and not kwargs.get('s3_endpoint_url'):
                 chunk_store = rdb_store
         telstate, capture_block_id, stream_name = view_l0_capture_stream(telstate, **kwargs)
         if chunk_store == 'auto':
-            chunk_store = infer_chunk_store(url_parts, telstate, **kwargs)
-        return cls(telstate, capture_block_id, stream_name,
-                   chunk_store, source_name=url_parts.geturl(), upgrade_flags=upgrade_flags)
+            chunk_store = infer_chunk_store(u, telstate, **kwargs)
+        return cls(telstate, capture_block_id, stream_name, chunk_store,
+                   source_name=u.geturl(), upgrade_flags=upgrade_flags)
 
 
 def open_data_source(url, **kwargs):
--- a/katdal/datasources.py
+++ b/katdal/datasources.py
@@ -646,8 +646,9 @@ class TelstateDataSource(DataSource):
             except katsdptelstate.ConnectionError as e:
                 raise DataSourceNotFound(str(e))
         elif u.scheme in {'http', 'https'}:
-            # Treat endpoint as an S3 object store (with auth info in kwargs)
-            store_url = (u.scheme, u.netloc, '', '', '', '')
+            # Treat URL prefix as an S3 object store (with auth info in kwargs)
+            store_path = os.path.dirname(os.path.dirname(u.path))
+            store_url = (u.scheme, u.netloc, store_path, '', '', '')
             store_url = urllib.parse.urlunparse(store_url)
             # Strip off parameters, query strings and fragments to get basic URL
             rdb_url = (u.scheme, u.netloc, u.path, '', '', '')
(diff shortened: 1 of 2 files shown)


katdal/datasources.py
```diff
@@ -632,7 +633,7 @@ def from_url(cls, url, chunk_store='auto', upgrade_flags=True, **kwargs):
         db = int(kwargs.pop('db', '0'))
         if url_parts.scheme == 'file':
             # RDB dump file
-            telstate = katsdptelstate.TelescopeState(katsdptelstate.memory.MemoryBackend())
+            telstate = katsdptelstate.TelescopeState()
```
bmerry: Actually I see that the change in telstate to make this the default didn't make it into the 0.7 release. I'd suggest reverting this change until we have a new telstate release (and then make katdal depend on it), since there will probably be users using PyPI katsdptelstate and git master katdal.

katdal/datasources.py
```diff
@@ -643,6 +644,16 @@ def from_url(cls, url, chunk_store='auto', upgrade_flags=True, **kwargs):
                 telstate = katsdptelstate.TelescopeState(url_parts.netloc, db)
             except katsdptelstate.ConnectionError as e:
                 raise DataSourceNotFound(str(e))
+        elif url_parts.scheme in {'http', 'https'}:
+            # Open endpoint as an S3 object store (with auth info in kwargs)
+            endpoint = urllib.parse.urlunparse(url_parts[:2] + 4 * ('',))
+            rdb_store = S3ChunkStore.from_url(endpoint, **kwargs)
+            rdb_path = urllib.parse.urlunparse(url_parts[:3] + 3 * ('',))
+            telstate = katsdptelstate.TelescopeState()
+            with rdb_store.request('RDB', 'GET', rdb_path) as response:
```
bmerry: Should a failure get turned into `DataSourceNotFound`?

katdal/datasources.py
```diff
@@ -643,6 +644,16 @@ def from_url(cls, url, chunk_store='auto', upgrade_flags=True, **kwargs):
                 telstate = katsdptelstate.TelescopeState(url_parts.netloc, db)
             except katsdptelstate.ConnectionError as e:
                 raise DataSourceNotFound(str(e))
+        elif url_parts.scheme in {'http', 'https'}:
+            # Open endpoint as an S3 object store (with auth info in kwargs)
+            endpoint = urllib.parse.urlunparse(url_parts[:2] + 4 * ('',))
+            rdb_store = S3ChunkStore.from_url(endpoint, **kwargs)
+            rdb_path = urllib.parse.urlunparse(url_parts[:3] + 3 * ('',))
```
bmerry: I think we should respect the nesting level of the request path, so that if the URL given is `https://foo.bar/level1/level2/123456789/123456789_sdp_l0.rdb', then we should construct the chunk store for `https://foo.bar/level1/level2`. We probably won't need it ourselves, but conceivable one might have a reverse proxy in front of a chunk store that maps it to a subpath.

katdal/datasources.py
ludwigschwardt: I like that.

katdal/datasources.py

ludwigschwardt:
Ready for more comments.

katdal/datasources.py
```diff
@@ -36,13 +36,14 @@
...
 from .chunkstore_npy import NpyFileChunkStore
+from .chunkstore import StoreUnavailable
 from .flags import DATA_LOST
 
 
 logger = logging.getLogger(__name__)
 
 
-class DataSourceNotFound(Exception):
+class DataSourceNotFound(OSError):
```
bmerry: If you do this, you should pass an errno and strerror to the constructor when available (e.g. when catching a real OSError and raising). I'm not 100% sure it's a good idea - it's normally used for system calls.

katdal/datasources.py
```diff
@@ -623,42 +622,49 @@ def from_url(cls, url, chunk_store='auto', upgrade_flags=True, **kwargs):
         kwargs : dict, optional
             Extra keyword arguments passed to telstate view and chunk store init
         """
-        url_parts = urllib.parse.urlparse(url, scheme='file')
+        u = urllib.parse.urlparse(url, scheme='file')
```
bmerry: Any particular reason for making this variable name less self-documenting?

katdal/datasources.py
```diff
@@ -623,42 +622,49 @@ def from_url(cls, url, chunk_store='auto', upgrade_flags=True, **kwargs):
...
+            store_url = urllib.parse.urlunparse(store_url)
+            # Strip off parameters, query strings and fragments to get basic URL
+            rdb_url = (u.scheme, u.netloc, u.path, '', '', '')
+            rdb_url = urllib.parse.urlunparse(rdb_url)
+            telstate = katsdptelstate.TelescopeState(katsdptelstate.memory.MemoryBackend())
+            try:
+                rdb_store = S3ChunkStore.from_url(store_url, **kwargs)
+                with rdb_store.request('', 'GET', rdb_url) as response:
+                    telstate.load_from_file(io.BytesIO(response.content))
+            except StoreUnavailable as e:
```
bmerry: Why not `ChunkStoreError`? Presumably if the smoke S3ChunkStore construction passes but the rdb file is missing you'll get `ChunkNotFound` rather than `StoreUnavailable`.

katdal/datasources.py
```diff
@@ -623,42 +622,49 @@ def from_url(cls, url, chunk_store='auto', upgrade_flags=True, **kwargs):
...
+                raise DataSourceNotFound(str(e))
             if chunk_store == 'auto' and not kwargs.get('s3_endpoint_url'):
```
bmerry: There may be uses for disabling this behaviour. For example, if one wants to rerun the continuum imager by hand, you'll want to use the RDB file from the archive, but the data will be in the on-site Ceph store. Maybe have another magic value like 'original' ignore this bit of code and apply `infer_chunk_store`.

katdal/datasources.py
ludwigschwardt: `StoreUnavailable` has been an `OSError` since the start... Also, all instances of `DataSourceNotFound` were raised after either an actual `OSError` or a `ConnectionError`, which is also an `OSError`. So I'm not 100% sure it's a bad idea :-)

katdal/datasources.py
ludwigschwardt: Reducing line lengths to fit into my editor window.

katdal/datasources.py
ludwigschwardt: I've now checked and a missing RDB is indeed a `ChunkStoreError`. I'll catch both.

katdal/datasources.py
ludwigschwardt: Reverted.

katdal/datasources.py
ludwigschwardt: Would you not have to specify `s3_endpoint_url` to point to the on-site Ceph store in this case?

katdal/datasources.py
bmerry: In that case it's probably a good idea, but we should pass through the (errno, strerror) pair from the underlying OSError.

katdal/datasources.py
bmerry: Yeah, I think meaningful variable names are more important than not have to scroll.

katdal/datasources.py
```diff
@@ -622,49 +622,49 @@ def from_url(cls, url, chunk_store='auto', upgrade_flags=True, **kwargs):
...
-            rdb_url = (u.scheme, u.netloc, u.path, '', '', '')
+            rdb_url = (url_parts.scheme, url_parts.netloc, url_parts.path, '', '', '')
             rdb_url = urllib.parse.urlunparse(rdb_url)
             telstate = katsdptelstate.TelescopeState(katsdptelstate.memory.MemoryBackend())
             try:
                 rdb_store = S3ChunkStore.from_url(store_url, **kwargs)
                 with rdb_store.request('', 'GET', rdb_url) as response:
                     telstate.load_from_file(io.BytesIO(response.content))
-            except StoreUnavailable as e:
+            except (StoreUnavailable, ChunkNotFound) as e:
```
bmerry: Why not just ChunkStoreError, which would cover anything?

Are there any errors that might leak out directly from requests (maybe DNS errors?), or does `_standard_errors` wrap them all?

katdal/datasources.py
bmerry: No, it's in the stream-specific `s3_endpoint_url` in telstate (the on-site versions of the data are given a different stream_name to disambiguate them).

katdal/datasources.py
ludwigschwardt: Gave up on this for now (since `ChunkNotFound` is a `KeyError` and not an `OSError`, it's all muddied anyway).

katdal/datasources.py
ludwigschwardt: `ChunkStoreError`, `ChunkNotFound`, got mixed up there :-)

Not sure about other errors - I'll do a quick DNS check and respond accordingly.

katdal/datasources.py
ludwigschwardt: In that case I'll just refrain from inferring the chunk store too early. I thought it would save some time but the store opens quite quickly anyway.

katdal/datasources.py
```diff
@@ -635,19 +635,34 @@ def from_url(cls, url, chunk_store='auto', upgrade_flags=True, **kwargs):
             telstate = katsdptelstate.TelescopeState(katsdptelstate.memory.MemoryBackend())
             try:
                 telstate.load_from_file(url_parts.path)
-            except OSError as err:
-                raise DataSourceNotFound(str(err))
+            except OSError as e:
+                raise DataSourceNotFound(str(e))
```
bmerry: I'm guessing this is a left-over from your OSError foray, and can be reverted.

katdal/datasources.py
```diff
@@ -635,19 +635,34 @@ def from_url(cls, url, chunk_store='auto', upgrade_flags=True, **kwargs):
...
         elif url_parts.scheme == 'redis':
             # Redis server
             try:
                 telstate = katsdptelstate.TelescopeState(url_parts.netloc, db)
             except katsdptelstate.ConnectionError as e:
                 raise DataSourceNotFound(str(e))
+        elif url_parts.scheme in {'http', 'https'}:
+            # Treat URL prefix as an S3 object store (with auth info in kwargs)
+            store_path = os.path.dirname(os.path.dirname(url_parts.path))
```
bmerry: Will this work on Windows, where os.path uses `\` as separator but the URL path will use `/`?

One option (that also avoids the need to unparse I think) is
```python
store_url = urllib.parse.urljoin(url, '..')
```

katdal/datasources.py
ludwigschwardt: No, I was just normalising the variable name for the exception with the rest of the file.

katdal/datasources.py
ludwigschwardt: I was worrying about the exact same thing but could not find a suitable "url.path.join" in `urllib`. I therefore like your suggestion a lot - will investigate.

ludwigschwardt: Fixed as suggested.

PR #234, opened 2019-03-22
Retrieve raw sensor data from CAM katstore

ludwigschwardt:
If a sensor is not found in the `SensorCache`, look it up in the katstore central database if a store address is provided. Retrieve the historical sensor values in the time range of the data set.

I've only enabled v4 support for now as I don't know if earlier sensor data is still available with the same REST API.

Usage:

```Python
f = katdal.open('1543151779_sdp_l0.full.rdb', sensor_store='portal.mkat.karoo.kat.ac.za')
#  or: f.sensor.store = 'portal.mkat.karoo.kat.ac.za'
f.sensor.get('cbf_1_target')
```

This addresses JIRA ticket SR-1710.

### DIFF
--- a/katdal/sensordata.py
+++ b/katdal/sensordata.py
@@ -18,7 +18,7 @@
 from __future__ import print_function, division, absolute_import
 from future import standard_library
 standard_library.install_aliases()  # noqa: E402
-import future.utils
+from future.utils import raise_from, PY3
 from builtins import zip, range, object
 from past.builtins import unicode
 
@@ -29,6 +29,7 @@ import threading
 import numpy as np
 import katpoint
 import katsdptelstate
+import requests
 
 from .categorical import (ComparableArrayWrapper, infer_dtype,
                           sensor_to_categorical)
@@ -174,7 +175,7 @@ def to_str(value):
     exception that numpy structured types with string or object fields won't
     be handled.
     """
-    if future.utils.PY3:
+    if PY3:
         if isinstance(value, np.ndarray) and value.dtype.kind == 'S':
             return np.char.decode(value, 'utf-8', 'surrogateescape')
         elif isinstance(value, bytes):
@@ -403,7 +404,6 @@ class TelstateSensorData(SensorData):
         else:
             raise ValueError("Sensor %r data has no key '%s'" % (self.name, key))
 
-
 # -------------------------------------------------------------------------------------------------
 # -- Utility functions
 # -------------------------------------------------------------------------------------------------
@@ -460,6 +460,73 @@ def _safe_linear_interp(xi, yi, x):
     # Do zeroth-order interpolation beyond the range of xi
     end_weight = np.clip(end_weight, 0.0, 1.0)
     return (1.0 - end_weight) * yi[start] + end_weight * yi[end]
+
+
+_KATCP_DECODER = {'integer': int, 'float': float, 'boolean': lambda x: x == '1',
+                  'lru': str, 'timestamp': float, 'discrete': str,
+                  'address': str, 'string': str}
+
+
+def get_sensor_from_katstore(store, name, start_time, end_time):
+    """Get raw sensor data from katstore (CAM's central sensor database).
+
+    Parameters
+    ----------
+    store : string
+        Hostname / endpoint of katstore webserver speaking katstore REST API
+    name : string
+        Sensor name (the escaped version with underscores)
+    start_time, end_time : float
+        Time range for sensor records as UTC seconds since Unix epoch
+
+    Returns
+    -------
+    data : :class:`RecordSensorData` object
+        Retrieved sensor data with 'timestamp', 'value' and 'status' fields
+
+    Raises
+    ------
+    ConnectionError
+        If this cannot connect to the katstore server
+    ValueError
+        If connection succeeded but interaction with REST API failed
+    KeyError
+        If the sensor was not found in the store or it has no data
+    """
+    # First check existence of sensor and get its KATCP type
+    url = "http://{}/katstore/sensors".format(store)
+    try:
+        result = requests.get(url, params={'sensors': name})
+    except requests.exceptions.ConnectionError as exc:
+        raise_from(ConnectionError("Could not connect to sensor store '%s'" %
+                                   (store,)), exc)
+    try:
+        sensor_info = {rec[0]: rec[2] for rec in result.json()}
+    except ValueError as exc:
+        raise_from(ValueError("Could not retrieve sensor info from '%s' (%d: %s)" %
+                              (url, result.status_code, result.reason)), exc)
+    if name not in sensor_info:
+        raise KeyError("Sensor store has no sensor named '%s'" % (name,))
+    decode = _KATCP_DECODER[sensor_info[name]['type']]
+    # Now get the historical data for the sensor in the given time range
+    url = "http://{}/katstore/samples".format(store)
+    params = {'sensor': name, 'start': start_time, 'end': end_time,
+              'time_type': 's', 'limit': 1000000}
+    try:
+        result = requests.get(url, params)
+    except requests.exceptions.ConnectionError as exc:
+        raise_from(ConnectionError("Could not connect to sensor store '%s'" %
+                                   (store,)), exc)
+    try:
+        samples = [(rec[1], decode(rec[3]), rec[5])
+                   for rec in result.json() if rec[4] == name]
+    except ValueError as exc:
+        raise_from(ValueError("Could not retrieve samples from '%s' (%d: %s)" %
+                              (url, result.status_code, result.reason)), exc)
+    if not samples:
+        raise KeyError("Sensor store has no data for sensor '%s'" % (name,))
+    samples = np.rec.fromrecords(samples, names='timestamp,value,status')
+    return RecordSensorData(samples, name)
 
 
 def dummy_sensor_data(name, value=None, dtype=np.float64, timestamp=0.0):
@@ -650,7 +717,8 @@ class SensorCache(dict):
 
     """
 
-    def __init__(self, cache, timestamps, dump_period, keep=slice(None), props=None, virtual={}, aliases={}):
+    def __init__(self, cache, timestamps, dump_period, keep=slice(None),
+                 props=None, virtual={}, aliases={}, store=None):
         # Initialise cache via dict constructor
         super(SensorCache, self).__init__(cache)
         self.timestamps = timestamps
@@ -662,6 +730,7 @@ class SensorCache(dict):
         # Add sensor aliases
         for alias, original in aliases.items():
             self.add_aliases(alias, original)
+        self.store = store
         # This needs to be an RLock because instantiating a virtual sensor
         # may require further sensor lookups (hopefully without a loop, which
         # would really cause problems).
@@ -799,7 +868,15 @@ class SensorCache(dict):
                         sensor_data = create_sensor(self, name, **match.groupdict())
                         break
                 else:
-                    raise KeyError("Unknown sensor '%s' (does not match actual name or virtual template)" % (name,))
+                    if self.store:
+                        # Katstore samples typically at least every 10 seconds
+                        start_time = self.timestamps[0] - self.dump_period - 10
+                        end_time = self.timestamps[-1] + self.dump_period + 10
+                        sensor_data = get_sensor_from_katstore(
+                            self.store, name, start_time, end_time)
+                    else:
+                        raise KeyError("Unknown sensor '%s' (does not match actual name or "
+                                       "virtual template and no sensor store provided)" % (name,))
             # If this is the first time this sensor is accessed, extract its data and store it in cache, if enabled
             if isinstance(sensor_data, SensorData) and extract:
                 # Look up properties associated with this specific sensor
--- a/katdal/sensordata.py
+++ b/katdal/sensordata.py
@@ -505,8 +505,10 @@ def get_sensor_from_katstore(store, name, start_time, end_time):
             raise_from(err, exc)
         with response:
             try:
+                response.raise_for_status()
                 sensor_info = {rec[0]: rec[2] for rec in response.json()}
-            except (ValueError, IndexError, TypeError, KeyError) as exc:
+            except (ValueError, IndexError, TypeError, KeyError,
+                    requests.exceptions.RequestException) as exc:
                 err = RuntimeError("Could not retrieve sensor info from '%s' (%d: %s)" %
                                    (url, response.status_code, response.reason))
                 raise_from(err, exc)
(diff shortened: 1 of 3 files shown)


bmerry:
I'm leaving on Wednesday for two weeks leave. If it's urgent I'd suggest asking someone else to review.

katdal/sensordata.py
```diff
@@ -18,7 +18,7 @@
 from __future__ import print_function, division, absolute_import
 from future import standard_library
 standard_library.install_aliases()  # noqa: E402
```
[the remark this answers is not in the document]
ludwigschwardt: Very useful. The irritating bit is that `# noqa` is all you really need, so this actually suppresses all flake8 errors, not only the suggested E402. One day `flake8` will get it right :-)

katdal/sensordata.py
```diff
@@ -462,6 +462,73 @@ def _safe_linear_interp(xi, yi, x):
...
+    ------
+    ConnectionError
+        If this cannot connect to the katstore server
+    ValueError
+        If connection succeeded but interaction with REST API failed
+    KeyError
+        If the sensor was not found in the store or it has no data
+    """
+    # First check existence of sensor and get its KATCP type
+    url = "http://{}/katstore/sensors".format(store)
```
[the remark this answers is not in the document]
ludwigschwardt: I also wondered whether I should look at `/api/client` aka the site map. I basically short-circuited that.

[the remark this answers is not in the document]
ajoubertza: Best to look at it, but then you'd have 3 requests - and you were trying to reduce latency.  Using the sitemap would help if we moved the internal endpoints around, but not if if the API changes.

katdal/sensordata.py
```diff
@@ -462,6 +462,73 @@ def _safe_linear_interp(xi, yi, x):
...
+    try:
+        result = requests.get(url, params={'sensors': name})
```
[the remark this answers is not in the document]
ludwigschwardt: Is that a Python regular expression?

katdal/sensordata.py
```diff
@@ -462,6 +462,73 @@ def _safe_linear_interp(xi, yi, x):
...
+
+def get_sensor_from_katstore(store, name, start_time, end_time):
+    """Get raw sensor data from katstore (CAM's central sensor database).
+
+    Parameters
+    ----------
+    store : string
+        Hostname / endpoint of katstore webserver speaking katstore REST API
+    name : string
+        Sensor name (the escaped version with underscores)
```
bmerry: What escaping needs to be applied? Or do you mean normalised?

katdal/sensordata.py
```diff
@@ -462,6 +462,73 @@ def _safe_linear_interp(xi, yi, x):
...
+    Returns
+    -------
+    data : :class:`RecordSensorData` object
+        Retrieved sensor data with 'timestamp', 'value' and 'status' fields
+
+    Raises
+    ------
+    ConnectionError
+        If this cannot connect to the katstore server
+    ValueError
```
bmerry: That doesn't feel like quite the right error for this situation. Maybe RuntimeError?

katdal/sensordata.py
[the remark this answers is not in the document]
ajoubertza: No, it gets handled by Postgres.  Part of a SQL statement like ` "WHERE name ~* %s "`

katdal/sensordata.py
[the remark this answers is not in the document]
bmerry: Is it worth taking a base URL rather than just the hostname? That makes it future-proof against things like possibly changing to https, or path changes.

katdal/sensordata.py
```diff
@@ -462,6 +462,73 @@ def _safe_linear_interp(xi, yi, x):
...
+    KeyError
+        If the sensor was not found in the store or it has no data
+    """
+    # First check existence of sensor and get its KATCP type
+    url = "http://{}/katstore/sensors".format(store)
+    try:
+        result = requests.get(url, params={'sensors': name})
+    except requests.exceptions.ConnectionError as exc:
+        raise_from(ConnectionError("Could not connect to sensor store '%s'" %
+                                   (store,)), exc)
```
bmerry: You're mixing `%s` and `{}` style formatting in this function.

katdal/sensordata.py
```diff
@@ -462,6 +462,73 @@ def _safe_linear_interp(xi, yi, x):
...
+    try:
+        sensor_info = {rec[0]: rec[2] for rec in result.json()}
```
bmerry: Before calling `result.json`, maybe check that the HTTP status is successful e.g. with `result.ok` or `result.raise_for_status`.

katdal/sensordata.py
```diff
@@ -462,6 +462,73 @@ def _safe_linear_interp(xi, yi, x):
...
+    except ValueError as exc:
```
bmerry: There are probably other errors that could occur e.g. IndexError if the record doesn't have 3 entries, and TypeError if it isn't a list of lists.

katdal/sensordata.py
```diff
@@ -462,6 +462,73 @@ def _safe_linear_interp(xi, yi, x):
...
+        raise_from(ValueError("Could not retrieve sensor info from '%s' (%d: %s)" %
+                              (url, result.status_code, result.reason)), exc)
+    if name not in sensor_info:
+        raise KeyError("Sensor store has no sensor named '%s'" % (name,))
+    decode = _KATCP_DECODER[sensor_info[name]['type']]
```
bmerry: This can also raise all sorts of exceptions - should it go inside the try block? And should it fall back to string decoding if the type isn't a known type?

katdal/sensordata.py
```diff
@@ -462,6 +462,73 @@ def _safe_linear_interp(xi, yi, x):
...
+    # Now get the historical data for the sensor in the given time range
+    url = "http://{}/katstore/samples".format(store)
+    params = {'sensor': name, 'start': start_time, 'end': end_time,
+              'time_type': 's', 'limit': 1000000}
+    try:
+        result = requests.get(url, params)
```
bmerry: Since you're making multiple requests, use a requests session. That will allow the requests to use the same connection. But also be sure to close each result when you're done with it (it can be used as a context manager).

katdal/sensordata.py
```diff
@@ -462,6 +462,73 @@ def _safe_linear_interp(xi, yi, x):
...
+    except ValueError as exc:
+        raise_from(ValueError("Could not retrieve sensor info from '%s' (%d: %s)" %
+                              (url, result.status_code, result.reason)), exc)
+    if name not in sensor_info:
+        raise KeyError("Sensor store has no sensor named '%s'" % (name,))
+    decode = _KATCP_DECODER[sensor_info[name]['type']]
+    # Now get the historical data for the sensor in the given time range
+    url = "http://{}/katstore/samples".format(store)
+    params = {'sensor': name, 'start': start_time, 'end': end_time,
+              'time_type': 's', 'limit': 1000000}
```
bmerry: What happens if there are more than this arbitrary maximum number?

katdal/sensordata.py
```diff
@@ -462,6 +462,73 @@ def _safe_linear_interp(xi, yi, x):
...
+    try:
+        result = requests.get(url, params)
+    except requests.exceptions.ConnectionError as exc:
+        raise_from(ConnectionError("Could not connect to sensor store '%s'" %
+                                   (store,)), exc)
+    try:
+        samples = [(rec[1], decode(rec[3]), rec[5])
+                   for rec in result.json() if rec[4] == name]
```
bmerry: Why would the name not match?

katdal/sensordata.py
```diff
@@ -462,6 +462,73 @@ def _safe_linear_interp(xi, yi, x):
...
+    except ValueError as exc:
+        raise_from(ValueError("Could not retrieve samples from '%s' (%d: %s)" %
+                              (url, result.status_code, result.reason)), exc)
```
bmerry: Again, check the status_code before even trying to parse the response.

katdal/sensordata.py
```diff
@@ -462,6 +462,73 @@ def _safe_linear_interp(xi, yi, x):
...
+        Retrieved sensor data with 'timestamp', 'value' and 'status' fields
+
+    Raises
+    ------
+    ConnectionError
+        If this cannot connect to the katstore server
+    ValueError
+        If connection succeeded but interaction with REST API failed
+    KeyError
+        If the sensor was not found in the store or it has no data
```
bmerry: no data in the time range?

katdal/sensordata.py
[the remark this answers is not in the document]
ludwigschwardt: So I assume the anchors are `^` and `$`?

katdal/visdatav4.py
```diff
@@ -116,7 +116,7 @@ class VisibilityDataV4(DataSet):
 
     """
     def __init__(self, source, ref_ant='', time_offset=0.0, applycal='',
-                 **kwargs):
+                 sensor_store=None, **kwargs):
```
bmerry: Need documentation for this arg.

katdal/sensordata.py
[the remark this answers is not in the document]
ajoubertza: I assume so too, but best to test it.

katdal/sensordata.py
```diff
@@ -650,7 +717,8 @@ class SensorCache(dict):
 
     """
 
-    def __init__(self, cache, timestamps, dump_period, keep=slice(None), props=None, virtual={}, aliases={}):
+    def __init__(self, cache, timestamps, dump_period, keep=slice(None),
+                 props=None, virtual={}, aliases={}, store=None):
```
bmerry: New arg needs documentation.

katdal/sensordata.py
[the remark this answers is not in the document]
bmerry: You should also escape the contents as well. I guess the fun part will be telling whether `re.escape` is going to escape the right things for Postgres.

katdal/sensordata.py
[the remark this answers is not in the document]
ludwigschwardt: `http://portal.mkat.karoo.kat.ac.za/katstore`? That is not safe either. I'll rather just update it when / if katstore changes. The main point to remember is that this feature has no users...

katdal/sensordata.py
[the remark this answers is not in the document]
ludwigschwardt: I'm just doing what `katportalclient` would have done here.

katdal/sensordata.py
ludwigschwardt: The relevant KATCP function is `katcp.resource.escape_name`:

```Python
def escape_name(name):
    """Escape sensor and request names to be valid Python identifiers."""
    return name.replace('.', '_').replace('-', '_')
```

I like "normalised" better but then reverted to the KATCP nomenclature, confusing as it is with the on-the-wire escape mechanism and all.

katdal/sensordata.py
ludwigschwardt: Indeed, thanks for pointing it out. I decided at some point to keep katdal homogeneous and stick to the older format, but I'm already regretting that decision :-)

katdal/sensordata.py
ludwigschwardt: I've found that `result.json()` fails with a `ValueError` if anything went wrong on the HTTP side, so now I only check for that instead of doing two checks (I still report the status code, so no debug info is lost).

katdal/sensordata.py
ludwigschwardt: Good, I'll catch those too.

katdal/sensordata.py
ludwigschwardt: I'm pleasantly surprised... That would have been my choice, but then I thought "What would Bruce do? Check rigorously that only the expected types came through." I'll happily revert 😛

katdal/sensordata.py
ludwigschwardt: Who knows - @lmagnus? This is the sanctioned cargo cult from his notebook example at http://kat-imager.kat.ac.za:8888/notebooks/tmp/katstore_query.ipynb. The number is supposed to be arbitrary and stupendously large (you need more than 32 samples/second on an 8-hour dataset to breach this - remember we are requesting one sensor at a time).

katdal/sensordata.py
ludwigschwardt: If something went wrong with the sensor selection (you can ask for multiple sensors with a name pattern / regex).

katdal/sensordata.py
ludwigschwardt: As before (the JSON error covers all cases).

katdal/sensordata.py
ludwigschwardt: Yes, will update.

katdal/visdatav4.py
ludwigschwardt: Done.

katdal/sensordata.py
ludwigschwardt: Good catch, forgot about that.

katdal/sensordata.py
ludwigschwardt: I like that... I caught the JSON ValueError, but now I'm catching others too so this makes more sense.

katdal/sensordata.py
bmerry: Does the API guarantee to return a non-JSON payload in the case of an error?

ludwigschwardt: @ajoubertza can hopefully clarify this. That has been my experience with simple 404's and misspelt sensors (an empty JSON for the latter).

katdal/sensordata.py
ludwigschwardt: Also, in the shiny katstore64 future, this *might* alll go away and only a single request will suffice without KATCP type tomfoolery.

ludwigschwardt: Gone with a slightly simpler and more secure option.

katdal/sensordata.py
ludwigschwardt: Maybe I'll mention both.

katdal/sensordata.py
ludwigschwardt: Done.

katdal/sensordata.py
ajoubertza: No guarantees.  This interface is not for public consumption.

ludwigschwardt: I'm thinking I should even remove the `sensor_store` parameter from `VisDataV4` since I want to emphasise that this is a power-user / debug-level feature. On the other hand, overriding the sensor store during `katdal.open` would allow katdal to open a severely broken RDB file, so I'm probably keeping it there for now.

I do like the following though 😄 

```Python
f = katdal.open('blah')
f.sensor.store = 'portal.mkat.karoo.kat.ac.za'
```

That also works for H5 files 😉

katdal/sensordata.py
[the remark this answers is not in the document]
bmerry: The anchoring and escaping doesn't seem to have been implemented.

katdal/sensordata.py
bmerry: I still think it's better to explicitly check the error code. You're relying on the the error payload having a specific format (namely, anything that isn't valid JSON) that I don't think is guaranteed.

bmerry: Also if you use raise_for_status then you can catch the requests error and then use raise_from, which means the original requests exception class is available in the error cause chain.

katdal/sensordata.py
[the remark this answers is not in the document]
ludwigschwardt: I didn't implement it because I took katportalclient as the gold standard and it didn't do it. But if you insist... :-) I've now added a sanitisation step before the HTTP request which should at least avoid the worst offenses.

ludwigschwardt:
This should be ready for another stab.

PR #235, opened 2019-03-26
Allow an expiration time for new buckets to be set

sratcliffe:
This add an expiry_days parameter to the chunk store that is used
to create lifecycle policies for any new buckets the chunk store
makes.

Only allows integer days.

### DIFF
--- a/katdal/chunkstore_s3.py
+++ b/katdal/chunkstore_s3.py
@@ -57,6 +57,16 @@ from .chunkstore import (ChunkStore, StoreUnavailable, ChunkNotFound, BadChunk,
 from .sensordata import to_str
 
 
+# Lifecycle policies unfortunately use XML encoding rather than JSON
+# Following path of least resistance we simply .format() this string
+# with the number of days for the expiry (and produced a sanitised
+# ID at the same time.
+_BASE_LIFECYCLE_POLICY = """<?xml version="1.0" encoding="UTF-8"?>
+<LifecycleConfiguration><Rule>
+<ID>katdal_expiry_{0}_days</ID><Prefix></Prefix><Status>Enabled</Status>
+<Expiration><Days>{0}</Days></Expiration>
+</Rule></LifecycleConfiguration>"""
+
 _BUCKET_POLICY = {
     "Version": "2012-10-17",
     "Id": "KatdalPolicy",
@@ -267,6 +277,10 @@ class S3ChunkStore(ChunkStore):
     public_read : bool
         If set to true, new buckets will be created with a policy that allows
         everyone (including unauthenticated users) to read the data.
+    expiry_days : int
+        If set to a value greater than 0 will set a future expiry time in days
+        for any new buckets created.
+
 
     Raises
     ------
@@ -274,7 +288,7 @@ class S3ChunkStore(ChunkStore):
         If requests is not installed (it's an optional dependency otherwise)
     """
 
-    def __init__(self, session_factory, url, public_read=False):
+    def __init__(self, session_factory, url, public_read=False, expiry_days=0):
         try:
             # Quick smoke test to see if the S3 server is available, by listing
             # buckets. Depending on the server in use, this may return a 403
@@ -294,9 +308,10 @@ class S3ChunkStore(ChunkStore):
         self._session_pool = _Pool(session_factory)
         self._url = to_str(url)
         self.public_read = public_read
+        self.expiry_days = int(expiry_days)
 
     @classmethod
-    def _from_url(cls, url, timeout, token, credentials, public_read):
+    def _from_url(cls, url, timeout, token, credentials, public_read, expiry_days):
         """Construct S3 chunk store from endpoint URL (see :meth:`from_url`)."""
         if token is not None:
             parsed = urllib.parse.urlparse(url)
@@ -318,11 +333,12 @@ class S3ChunkStore(ChunkStore):
             session.mount(url, adapter)
             return session
 
-        return cls(session_factory, url, public_read)
+        return cls(session_factory, url, public_read, expiry_days)
 
     @classmethod
     def from_url(cls, url, timeout=300, extra_timeout=10,
-                 token=None, credentials=None, public_read=False, **kwargs):
+                 token=None, credentials=None, public_read=False,
+                 expiry_days=0, **kwargs):
         """Construct S3 chunk store from endpoint URL.
 
         Parameters
@@ -341,6 +357,9 @@ class S3ChunkStore(ChunkStore):
         public_read : bool
             If set to true, new buckets will be created with a policy that allows
             everyone (including unauthenticated users) to read the data.
+        expiry_days : int
+            If set to a value greater than 0 will set a future expiry time in days
+            for any new buckets created.
         kwargs : dict
             Extra keyword arguments (unused)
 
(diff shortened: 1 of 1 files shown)


katdal/chunkstore_s3.py
```diff
@@ -57,6 +57,16 @@
 from .sensordata import to_str
 
 
+# Lifecycle policies unfortunately use XML encoding rather than JSON
+# Following path of least resistance we simply .format() this string
+# with the number of days for the expiry (and produced a sanitised
+# ID at the same time.
```
bmerry: Missing `)`.

katdal/chunkstore_s3.py
```diff
@@ -267,14 +277,18 @@ class S3ChunkStore(ChunkStore):
     public_read : bool
         If set to true, new buckets will be created with a policy that allows
         everyone (including unauthenticated users) to read the data.
+    expiry_days : int
+        If set to a value greater than 0 will set a future expiry time in days
+        for any new buckets created.
```
bmerry: I prefer to use `None` for never. Zero sounds like "expire immediately", and since it's not C so we don't have to make some numbers have special meanings.

katdal/chunkstore_s3.py
```diff
@@ -446,6 +465,15 @@ def create_array(self, array_name):
             with self.request(None, 'PUT', policy_url, data=json.dumps(policy)):
                 pass
 
+        if self.expiry_days > 0:
+            lifecycle_url = urllib.parse.urljoin(url, '?lifecycle')
+            lifecycle = copy.deepcopy(_BASE_LIFECYCLE_POLICY)
+            xml_payload = lifecycle.format(self.expiry_days)
+            b64_md5 = base64.encodestring(hashlib.md5(xml_payload.encode('utf-8')).digest()).decode('utf-8')[:-1]
+            lifecycle_headers = {'Content-Type':'text/xml', 'Content-MD5': b64_md5}
```
bmerry: Missing space after first `:`.

katdal/chunkstore_s3.py
```diff
@@ -446,6 +465,15 @@ def create_array(self, array_name):
             with self.request(None, 'PUT', policy_url, data=json.dumps(policy)):
                 pass
 
+        if self.expiry_days > 0:
+            lifecycle_url = urllib.parse.urljoin(url, '?lifecycle')
+            lifecycle = copy.deepcopy(_BASE_LIFECYCLE_POLICY)
```
bmerry: This copy is unnecessary. Unlike for the bucket policy, you're not modifying anything in place.

katdal/chunkstore_s3.py
```diff
@@ -446,6 +465,15 @@ def create_array(self, array_name):
...
+            xml_payload = lifecycle.format(self.expiry_days)
+            b64_md5 = base64.encodestring(hashlib.md5(xml_payload.encode('utf-8')).digest()).decode('utf-8')[:-1]
```
bmerry: I'm not quite clear why you're using `encodestring`. If I'm understanding the docs right, it's going to add a newline (compared to using `b64encode`), which you're then stripping off again.

katdal/chunkstore_s3.py
```diff
@@ -446,6 +465,15 @@ def create_array(self, array_name):
             with self.request(None, 'PUT', policy_url, data=json.dumps(policy)):
                 pass
 
+        if self.expiry_days > 0:
+            lifecycle_url = urllib.parse.urljoin(url, '?lifecycle')
```
bmerry: You can pass `params='lifecycle'` to `request` rather than mangling the url yourself.

katdal/chunkstore_s3.py
bmerry: Any reason to be using `encodestring` rather than `b64encode`? From what I can see, the difference is that it's going to insert newlines, but I'm assuming `[:-1]` is to undo that.

bmerry:
Do we actually want to bundle the number of days into the policy name, or would it be more convenient to have a single name for if we want to remove/replace the policy without knowing the current number of days?

sratcliffe:
The policies exist specifically per bucket so there is no universal way to replace them. Including the day count just makes it a little simpler to see what is going on with the very rudimentary tools that radosgw-admin provides (you can only see the policy name, not the contents, to get that you have to use s3cmd).

katdal/chunkstore_s3.py
```diff
@@ -57,6 +57,16 @@
 from .sensordata import to_str
 
 
+# Lifecycle policies unfortunately use XML encoding rather than JSON
+# Following path of least resistance we simply .format() this string
+# with the number of days for the expiry (and produced a sanitised
+# ID at the same time.
+_BASE_LIFECYCLE_POLICY = """<?xml version="1.0" encoding="UTF-8"?>
+<LifecycleConfiguration><Rule>
+<ID>katdal_expiry_{0}_days</ID><Prefix></Prefix><Status>Enabled</Status>
```
bmerry: Looking at https://docs.aws.amazon.com/AmazonS3/latest/dev/intro-lifecycle-rules.html, you also need to wrap `<Prefix>` in a `<Filter>` tag?

katdal/chunkstore_s3.py
sratcliffe: The reason is to make it clear that 0 does nothing, as opposed to explaining that you can't actually use 0 in expiry policies, but no problem to change.

bmerry:
> Including the day count just makes it a little simpler to see what is going on with the very rudimentary tools that radosgw-admin provides

Ok that sounds like a good reason to do it that way then.

katdal/chunkstore_s3.py
sratcliffe: Fixed (silly holdover from JSON version)

katdal/chunkstore_s3.py
sratcliffe: Good catch - works well.

katdal/chunkstore_s3.py
sratcliffe: I had an issue with doing that earlier, but it may have been conflated with some other issue. Let me test further...

sratcliffe: Seems to work - now including an empty <Filter></Filter> element as this seems valid and conveys our intent more clearly.

ludwigschwardt:
I feel Simon's pain on the interplay of S3ChunkStore() with `S3ChunkStore.from_url` and `S3ChunkStore._from_url`. I have two suggestions:

- Do away with the complicated structure as far as possible. This is supported by the actual use cases (9 instances of construction via `from_url` and 0 constructions via the usual initialiser).
- If not possible, at least pass all the top-level parameters like `expiry_days` around in `kwargs`.

katdal/chunkstore_s3.py
```diff
@@ -57,6 +57,16 @@
 from .sensordata import to_str
 
 
+# Lifecycle policies unfortunately use XML encoding rather than JSON
```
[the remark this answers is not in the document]
sratcliffe: Trying to buy some review kindness from bmerry :)

bmerry:
This needs a unit test. Obviously it won't be able to wait a day to check that the objects actually vanish, but it will presumably at least validate the XML. And if minio doesn't support lifecycle policies, we need to know anyway so that we can come up with a new plan for dev machine testing.

PR #239, opened 2019-04-24
Include lifecycle policy tests

sratcliffe:
We can't test expiry directly, firstly because minio doesn't support lifecycle polices, and secondly because the minimum expiry time is 1 day.

This basically just test the infrastructure and that the schema for the policy is created correctly.

### DIFF
--- a/katdal/chunkstore.py
+++ b/katdal/chunkstore.py
@@ -37,6 +37,9 @@ class ChunkStoreError(Exception):
 class StoreUnavailable(OSError, ChunkStoreError):
     """Could not access underlying storage medium (offline, auth failed, etc)."""
 
+class NotSupported(ChunkStoreError):
+    """The underlying store does not support the requested operation
+    (e.g. minio doesn't support lifecycle policies on buckets)"""
 
 class ChunkNotFound(KeyError, ChunkStoreError):
     """The store was accessible but a chunk with the given name was not found."""
--- a/katdal/chunkstore.py
+++ b/katdal/chunkstore.py
@@ -37,9 +37,11 @@ class ChunkStoreError(Exception):
 class StoreUnavailable(OSError, ChunkStoreError):
     """Could not access underlying storage medium (offline, auth failed, etc)."""
 
+
 class NotSupported(ChunkStoreError):
     """The underlying store does not support the requested operation
     (e.g. minio doesn't support lifecycle policies on buckets)"""
+
 
 class ChunkNotFound(KeyError, ChunkStoreError):
     """The store was accessible but a chunk with the given name was not found."""
--- a/katdal/chunkstore_s3.py
+++ b/katdal/chunkstore_s3.py
@@ -24,7 +24,7 @@ from future import standard_library
 standard_library.install_aliases()  # noqa: E402
 from builtins import object
 import future.utils
-from future.utils import raise_, bytes_to_native_str
+from future.utils import raise_, bytes_to_native_str, raise_from
 
 import contextlib
 import io
@@ -53,8 +53,10 @@ except ImportError:
     botocore = None
 
 from .chunkstore import (ChunkStore, StoreUnavailable, ChunkNotFound, BadChunk,
-                         npy_header_and_body)
+                         NotSupported, npy_header_and_body)
 from .sensordata import to_str
+
+from . import schemas
 
 
 # Lifecycle policies unfortunately use XML encoding rather than JSON
@@ -180,6 +182,8 @@ def _raise_for_status(response):
     except requests.HTTPError as error:
         if response.status_code == 404:
             raise ChunkNotFound(str(error))
+        elif response.status_code == 501:
+            raise_from(NotSupported(str(error)), error)
         else:
             raise StoreUnavailable(str(error))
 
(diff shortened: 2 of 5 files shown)


katdal/schemas/__init__.py
```diff
@@ -0,0 +1,58 @@
...
+
+        Raises
+        ------
+        etree.DocumentInvalid
+            if `xml_string` does not validate against the XSD schema
+        ValueError
+            if `xml_string` cannot be parsed into a valid XML document
+        """
+        try:
+            xml_doc = etree.fromstring(bytes(bytearray(xml_string, encoding='utf-8')))
```
ludwigschwardt: Out of curiosity, why `bytes` and `bytesarray`?

katdal/schemas/__init__.py
```diff
@@ -0,0 +1,58 @@
...
+        except etree.XMLSyntaxError as e:
+            raise ValueError(e)
```
ludwigschwardt: This is a candidate for `raise_from`.

katdal/schemas/__init__.py
```diff
@@ -0,0 +1,58 @@
...
+        if not self.validator.validate(xml_doc):
+            log = self.validator.error_log
+            raise etree.DocumentInvalid(log.last_error)
```
ludwigschwardt: You are converting `etree.XMLSyntaxError` to `ValueError` but leaving `etree.DocumentInvalid` as is... This seems a bit inconsistent. Will the end user typically be expected to catch invalid document exceptions? If that's the case, I'd probably also transmogrify it for the user's convenience.

katdal/schemas/__init__.py
```diff
@@ -0,0 +1,58 @@
...
+        return True
+
+
+def validate(validator_name, string_to_validate):
+    try:
+        return validators[validator_name].validate(string_to_validate)
+    except KeyError:
+        raise KeyError("Specified validator {} doesn't map to an installed"
```
ludwigschwardt: Maybe `raise_from(..., None)` here to suppress the double `KeyError` in the chain.

test-requirements.txt
```diff
@@ -1,5 +1,6 @@
 coverage
 funcsigs
+lxml==4.3.3
```
ludwigschwardt: Do you want this in `requirements.txt` as well, to be used on site?

katdal/chunkstore.py
```diff
@@ -38,6 +38,11 @@ class StoreUnavailable(OSError, ChunkStoreError):
     """Could not access underlying storage medium (offline, auth failed, etc)."""
 
 
+class NotSupported(ChunkStoreError):
```
ludwigschwardt: This name is a bit vague... How about `UnsupportedStoreFeature`?

katdal/chunkstore_s3.py
```diff
@@ -467,6 +483,10 @@ def create_array(self, array_name):
 
         if self.expiry_days > 0:
             xml_payload = _BASE_LIFECYCLE_POLICY.format(self.expiry_days)
+            if self.validate_xml_policies:
+                if not schemas.has_lxml:
+                    raise ImportError("XML schema validation requires lxml to be installed.")
```
ludwigschwardt: It would be nice to link to the original error via `raise_from`. See e.g. the RDB reader handling in katsdptelstate.

katdal/chunkstore_s3.py
```diff
@@ -467,6 +483,10 @@ def create_array(self, array_name):
 
         if self.expiry_days > 0:
             xml_payload = _BASE_LIFECYCLE_POLICY.format(self.expiry_days)
```
ludwigschwardt: So `xml_payload` is always `_BASE_LIFECYCLE_POLICY`? When would it not validate - a bad value for `expiry_days`? It seems like a lot of effort to solve future problems.

katdal/schemas/__init__.py
sratcliffe: Python2.7 doesn't allow encoding types in bytes, but does in bytearray. Kludge to keep 2/3 compatible.

katdal/schemas/__init__.py
sratcliffe: Done

katdal/schemas/__init__.py
sratcliffe: Simply because the XMLSyntaxError is quite messy. I also wanted to easily catch the difference between an XML syntax error and an invalid document. If I didn't use DocumentInvalid, then ValueError would be the sensible standard exception but wouldn't allow this differentiation.

katdal/schemas/__init__.py
sratcliffe: Done

test-requirements.txt
sratcliffe: Nope - the import checking handles this being missing for production systems and lets you know that validation wont be used if it is missing.

katdal/chunkstore.py
sratcliffe: Fair enough - changed

katdal/chunkstore_s3.py
sratcliffe: Not sure I agree - I prefer the very unambiguous nature of just telling the user that you need lxml if you want to use validators.

katdal/chunkstore_s3.py
sratcliffe: The grand scheme is that you will sometimes want to override the base for very user specific things - like managing storage migration say. I can see us ending up with quite a number of policies as future versions of CEPH support more elaborate management via lifecycle policies.

katdal/schemas/__init__.py
bmerry: There should be something in `future.utils` to force things one way or another.

But also, why encode at all? lxml.etree.fromstring seems perfectly happy to take Unicode.

PR #240, opened 2019-05-06
Use features introduced in katsdptelstate v0.8

ludwigschwardt:
The memory backend is now the default - no need to specify it.
Use the `TelescopeState.join` helper method instead of invoking the
namespace separator itself. Catch RdbParseError to improve error
reporting. Since `keys` and `prefixes` already return native strings
there is no need to override them in `TelstateToStr`. The RDBWriter
has been reworked. Remove fakeredis from package requirements.

### DIFF
--- a/katdal/sensordata.py
+++ b/katdal/sensordata.py
@@ -296,11 +296,11 @@ class TelstateToStr(object):
         return TelstateToStr(self._telstate.root())
 
     def keys(self, filter='*'):
-        return to_str(self._telstate.keys(filter))
+        return self._telstate.keys(filter)
 
     @property
     def prefixes(self):
-        return to_str(self._telstate.prefixes)
+        return self._telstate.prefixes
 
     def __getattr__(self, key):
         # __getattr__ can be used for item access or to get a property of the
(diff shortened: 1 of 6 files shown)


katdal/sensordata.py
```diff
@@ -296,11 +296,11 @@ def root(self):
         return TelstateToStr(self._telstate.root())
 
     def keys(self, filter='*'):
-        return to_str(self._telstate.keys(filter))
+        return self._telstate.keys(filter)
```
bmerry: This whole function can be deleted, because the version in `_telstate` will work (and will be found by `__getattr__`).

katdal/sensordata.py
```diff
@@ -296,11 +296,11 @@ def root(self):
...
 
     @property
     def prefixes(self):
-        return to_str(self._telstate.prefixes)
+        return self._telstate.prefixes
```
bmerry: I think this function can also safely be deleted.

ludwigschwardt: Yip.

ludwigschwardt:
So I'm leaving this as is for now and considering a complete overhaul in the next release instead :-)

See JIRA ticket [SR-1777](https://skaafrica.atlassian.net/browse/SR-1777) for more details.

PR #241, opened 2019-05-07
Restore chunk store inference from RDB URL

ludwigschwardt:
If the RDB file is opened via archive URL, use that URL and
corresponding S3 credentials or token to access the chunk store. This
has been decreed since it is the expected behaviour 99% of the time.
On-site chunk stores will be accessed via whitelists only, thereby not
requiring credentials or tokens, which removes the concern that
prevented this shortcut. Overrides are still possible via URL parameters
(and separate RDB retrieval) for custom use cases.

This reverts the change in commit 98cbc8a and addresses JIRA ticket
[SR-1443](https://skaafrica.atlassian.net/browse/SR-1443).

### DIFF
--- a/katdal/datasources.py
+++ b/katdal/datasources.py
@@ -655,6 +655,10 @@ class TelstateDataSource(DataSource):
                     telstate.load_from_file(io.BytesIO(response.content))
             except ChunkStoreError as e:
                 raise DataSourceNotFound(str(e))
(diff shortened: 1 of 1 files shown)


ludwigschwardt:
@bmerry, last one before the changelog update PR (and it's only 4 lines 😉 )

PR #250, opened 2019-06-18
Add some guidelines on getting good performance from katdal

bmerry:


bmerry:
@nirujmohanr the documentation we already have is at https://katdal.readthedocs.io.

nirujmohanr:
I had a quick look, but not at the modules API since I haven't used those yet. A trivial point is that, when you print d and get an output of the properties of the dataset, the file size in MB or GB uses 1000 to calculate the size. It should use 1024 instead, isn't it?

mserylak:
@nirujmohanr, so you want data size in Gb (https://en.wikipedia.org/wiki/Binary_prefix#kibi) or GB?

ludwigschwardt:
As Maciej said, 1 GiB = 1024 MiB but 1 GB = 1000 MB. The storage folks don't care about factors of 1024 anymore, since they can sell you a 1000 GB drive and then give you 7% less space than you expected while not technically lying.

nirujmohanr:
Ah, i didnt know that. I had assumed that all of it was by a factor of 1024.

PR #251, opened 2019-07-05
Clean up virtual sensors

ludwigschwardt:
This is in preparation of the (u,v,w) speed-up PR.

Highlights:
- Unit tests! The `MinimalDataSet` class is quite nifty.
- Improved code-level documentation and refactoring.

### DIFF
--- a/katdal/sensordata.py
+++ b/katdal/sensordata.py
@@ -874,7 +874,9 @@ class SensorCache(dict):
                 # Otherwise, iterate through virtual sensor templates and look for a match
                 for pattern, create_sensor in self.virtual.items():
                     # Expand variable names enclosed in braces to the relevant regular expression
-                    pattern = re.sub(r'({[a-zA-Z_]\w*})', lambda m: '(?P<' + m.group(0)[1:-1] + '>[^//]+)', pattern)
+                    # (match anything but slashes, which are preferred delimiters in virtual sensor names)
+                    pattern = re.sub(r'({[a-zA-Z_]\w*})',
+                                     lambda m: '(?P<%s>[^/]+)' % (m.group(0)[1:-1],), pattern)
                     match = re.match(pattern, name)
                     if match:
                         # Call sensor creation function with extracted variables from sensor name
--- a/katdal/sensordata.py
+++ b/katdal/sensordata.py
@@ -875,7 +875,7 @@ class SensorCache(dict):
                 for pattern, create_sensor in self.virtual.items():
                     # Expand variable names enclosed in braces to the relevant regular expression
                     # (match anything but slashes, which are preferred delimiters in virtual sensor names)
-                    pattern = re.sub(r'({[a-zA-Z_]\w*})',
+                    pattern = re.sub(r'(\{[a-zA-Z_]\w*\})',
                                      lambda m: '(?P<%s>[^/]+)' % (m.group(0)[1:-1],), pattern)
                     match = re.match(pattern, name)
                     if match:
--- a/katdal/test/test_dataset.py
+++ b/katdal/test/test_dataset.py
@@ -50,6 +50,7 @@ class MinimalDataSet(DataSet):
             az, el = target.azel(timestamps, ant)
             sensors['Antennas/%s/az' % (ant.name,)] = az
             sensors['Antennas/%s/el' % (ant.name,)] = el
+        # Extract array reference position as 'array_ant' from last antenna
         array_ant_fields = ['array'] + ant.description.split(',')[1:5]
         array_ant = Antenna(','.join(array_ant_fields))
         sensors['Antennas/array/antenna'] = constant_sensor(array_ant)
@@ -109,6 +110,7 @@ class TestVirtualSensors(object):
         subarray = Subarray(self.antennas, corrprods)
         spw = SpectralWindow(centre_freq=1284e6, channel_width=0, num_chans=16,
                              sideband=1, bandwidth=856e6)
+        # Pick a time when the source is up as that seems more realistic
         self.timestamps = 1234667890.0 + 1.0 * np.arange(10)
         self.dataset = MinimalDataSet(self.target, subarray, spw, self.timestamps)
         self.array_ant = self.dataset.sensor.get('Antennas/array/antenna')[0]
@@ -117,6 +119,7 @@ class TestVirtualSensors(object):
         mjd = Timestamp(self.timestamps[0]).to_mjd()
         assert_equal(self.dataset.mjd[0], mjd)
         lst = self.array_ant.local_sidereal_time(self.timestamps)
+        # Convert LST from radians (katpoint) to hours (katdal)
         assert_array_equal(self.dataset.lst, lst * (12 / np.pi))
 
     def test_pointing(self):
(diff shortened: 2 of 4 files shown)


katdal/sensordata.py
```diff
@@ -874,7 +874,9 @@ def get(self, name, select=False, extract=True, **kwargs):
                 # Otherwise, iterate through virtual sensor templates and look for a match
                 for pattern, create_sensor in self.virtual.items():
                     # Expand variable names enclosed in braces to the relevant regular expression
-                    pattern = re.sub(r'({[a-zA-Z_]\w*})', lambda m: '(?P<' + m.group(0)[1:-1] + '>[^//]+)', pattern)
+                    # (match anything but slashes, which are preferred delimiters in virtual sensor names)
+                    pattern = re.sub(r'({[a-zA-Z_]\w*})',
```
bmerry: Maybe escape the curly braces (from regex). Python tends to be getting stricter about escaping over time, so at some stage it may start complaining that the value inside the braces is not a valid repetition count.

katdal/test/test_sensordata.py
```diff
@@ -71,3 +71,29 @@ def test_numpy_object(self):
...
+        # If your parameter values contain underscores, don't use it as delimiter
+        params = {'ant': 'm000', 'param1': 'one', 'param2': 'two_three'}
+        with assert_raises(AssertionError):
+            self.cache.get(template.format(**params))
+        template = 'Antennas/{ant}/{param1}/{param2}'
+        # The updated template has not yet been added to the cache
+        with assert_raises(KeyError):
+            self.cache.get(template.format(**params))
+        self.cache.virtual[template] = _check_sensor
+        self.cache.get(template.format(**params))
```
bmerry: Looks good so far, but I can suggest two more tests:
1. Have `_check_sensor` return something, and check that this becomes the value of the virtual sensor.
2. Retrieve the sensor twice, and check that `_check_sensor` is only called the first time i.e. that the cache actually caches results.

Possibly mock would help with this (it basically gives you a callable that keeps a record of the calls made to it), but it might not be worth the extra complexity.

katdal/test/test_dataset.py
```diff
@@ -38,3 +94,48 @@ def test_selection_to_list():
...
+                                 '-8.264 -207.29 8.5965'),
+                         Antenna('m063, -30:42:39.8, 21:26:38.0, 1086.6, 13.5, '
+                                 '-3419.5845 -1840.48 16.3825')]
+        corrprods = [('m000h', 'm000h'), ('m000v', 'm000v'),
+                     ('m063h', 'm063h'), ('m063v', 'm063v'),
+                     ('m000h', 'm063h'), ('m000v', 'm063v')]
+        subarray = Subarray(self.antennas, corrprods)
+        spw = SpectralWindow(centre_freq=1284e6, channel_width=0, num_chans=16,
+                             sideband=1, bandwidth=856e6)
+        self.timestamps = 1234667890.0 + 1.0 * np.arange(10)
```
bmerry: Did you meant to use 1234567890.0? Or do you just have something against 5's? :-)

katdal/test/test_dataset.py
```diff
@@ -38,3 +94,48 @@ def test_selection_to_list():
...
+        self.dataset = MinimalDataSet(self.target, subarray, spw, self.timestamps)
+        self.array_ant = self.dataset.sensor.get('Antennas/array/antenna')[0]
+
+    def test_timestamps(self):
+        mjd = Timestamp(self.timestamps[0]).to_mjd()
+        assert_equal(self.dataset.mjd[0], mjd)
+        lst = self.array_ant.local_sidereal_time(self.timestamps)
+        assert_array_equal(self.dataset.lst, lst * (12 / np.pi))
```
bmerry: What does 12/pi represent?

katdal/test/test_dataset.py
```diff
@@ -38,3 +94,48 @@ def test_selection_to_list():
...
+
+    def test_pointing(self):
+        az, el = self.target.azel(self.timestamps, self.antennas[1])
+        assert_array_equal(self.dataset.az[:, 1], rad2deg(az))
+        assert_array_equal(self.dataset.el[:, 1], rad2deg(el))
+        ra, dec = self.target.radec(self.timestamps, self.antennas[0])
+        assert_array_almost_equal(self.dataset.ra[:, 0], rad2deg(ra), decimal=5)
+        assert_array_almost_equal(self.dataset.dec[:, 0], rad2deg(dec), decimal=5)
+        angle = self.target.parallactic_angle(self.timestamps, self.antennas[0])
+        # TODO: Check why this is so poor...
```
bmerry: Have you checked? That seems pretty worrying, given that it's presumably using the same code to compute it.

katdal/test/test_dataset.py
```diff
@@ -38,3 +94,48 @@ def test_selection_to_list():
...
+        x, y = self.target.sphere_to_plane(az, el, self.timestamps, self.antennas[1])
+        assert_array_equal(self.dataset.target_x[:, 1], rad2deg(x))
+        assert_array_equal(self.dataset.target_y[:, 1], rad2deg(y))
+
+    def test_uvw(self):
+        u, v, w = self.target.uvw(self.antennas[0], self.timestamps,
+                                  self.antennas[1])
+        assert_array_equal(self.dataset.u[:, 4], u)
+        assert_array_equal(self.dataset.v[:, 4], v)
+        assert_array_equal(self.dataset.w[:, 5], w)
```
bmerry: Why 4, 4, 5?

katdal/test/test_dataset.py
```diff
@@ -19,8 +19,64 @@
...
+        self.subarrays = [subarray]
+        self.spectral_windows = [spectral_window]
+        sensors = {}
+        for ant in subarray.ants:
+            sensors['Antennas/%s/antenna' % (ant.name,)] = constant_sensor(ant)
+            az, el = target.azel(timestamps, ant)
+            sensors['Antennas/%s/az' % (ant.name,)] = az
+            sensors['Antennas/%s/el' % (ant.name,)] = el
+        array_ant_fields = ['array'] + ant.description.split(',')[1:5]
+        array_ant = Antenna(','.join(array_ant_fields))
```
bmerry: This bit of magic could use some comments (and maybe it's time for a utility function in katpoint).

katdal/sensordata.py
ludwigschwardt: Good point, will do.

katdal/test/test_dataset.py
ludwigschwardt: I had the 5 there but then the source was below the horizon and I fiddled with the timestamp until it rose 😄 I figured that would be a more representative data set.

katdal/test/test_dataset.py
ludwigschwardt: This turns radians into hours. Will make it clearer.

katdal/test/test_dataset.py
ludwigschwardt: I have checked and could not find anything obvious, hence the TODO... More eyeballs appreciated!

katdal/test/test_dataset.py
ludwigschwardt: Since corrprod 4 is `('m000h', 'm063h')` and corrprod 5 is `('m000v', 'm063v')`, they only differ in polarisation and should have identical `(u, v, w)`. This way I check both. Adding comment.

katdal/test/test_dataset.py
ludwigschwardt: Agreed.

katdal/sensordata.py
ludwigschwardt: Done.

katdal/test/test_sensordata.py
ludwigschwardt: Done.

bmerry:
I'm happy with the latest changes. Up to you if you want to just merge and file a bug to look into the parallactic angle thing later, or dig into it first.

ludwigschwardt:
I filed a bug at https://skaafrica.atlassian.net/browse/SR-1882.

PR #252, opened 2019-07-18
Speed up (u, v, w) calculations

ludwigschwardt:
This brings the `katdal.DataSet` (u, v, w) calculations in line with the rest of the ecosystem (`mvftoms`, `katsdpcal`, `katsdpimager`, ... thanks @bmerry!).

The speed-up is on the order of the number of baselines, measuring about 500x on a dataset with 60 antennas (1542913317). The accuracy was still to be verified in JIRA SR-1883, but since this has been in use since at least April 2018 in MSes produced by katdal and verified by Tom on PKS 1934-638, we can lay that part to rest.

One caveat is that this will break (u, v, w) on older MVF formats before v4 that do not yet define an 'array' antenna. We'll need to remedy that still.

This addresses JIRA ticket SR-1509 (and SR-1640).

### DIFF
--- a/katdal/dataset.py
+++ b/katdal/dataset.py
@@ -238,8 +238,8 @@ def _calc_uvw_basis(cache, name, ant):
     for segm, target in targets.segments():
         basis = target.uvw_basis(cache.timestamps[segm], antenna)
         u[segm], v[segm], w[segm] = basis.transpose(0, 2, 1)
-    new_sensors = {ant_group + 'u_basis': u, ant_group + 'v_basis': v,
-                   ant_group + 'w_basis': w}
+    new_sensors = {ant_group + 'basis_u': u, ant_group + 'basis_v': v,
+                   ant_group + 'basis_w': w}
     cache.update(new_sensors)
     return new_sensors[name]
 
@@ -249,7 +249,7 @@ def _calc_uvw_per_ant(cache, name, ant):
     array_ant_group = 'Antennas/array/'
     array_antenna = cache.get(array_ant_group + 'antenna')[0]
     antenna = cache.get('Antennas/%s/antenna' % (ant,))[0]
-    basis = cache.get('%s%s_basis' % (array_ant_group, name[-1]))
+    basis = cache.get('%sbasis_%s' % (array_ant_group, name[-1]))
     # Obtain baseline vector from array reference to specified antenna
     baseline_m = array_antenna.baseline_toward(antenna)
     coord = np.tensordot(basis, baseline_m, ([1], [0]))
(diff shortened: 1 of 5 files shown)


katdal/test/test_dataset.py
```diff
@@ -139,10 +139,13 @@ def test_pointing(self):
     def test_uvw(self):
         u, v, w = self.target.uvw(self.antennas[0], self.timestamps,
                                   self.antennas[1])
-        assert_array_equal(self.dataset.u[:, 4], u)
-        assert_array_equal(self.dataset.v[:, 4], v)
-        assert_array_equal(self.dataset.w[:, 4], w)
+        # The answers now differ because the target direction is calculated once
+        # from the array reference instead of from one of the dishes in each
+        # correlation product pair (the whole reason for the speed-up!)
```
bmerry: You could still compute the reference value as the difference between two UVWs measured to the array centre.

katdal/dataset.py
```diff
@@ -227,29 +227,43 @@ def _calc_target_coords(cache, name, ant, projection, coordsys):
...
     targets = cache.get('Observation/target')
     for segm, target in targets.segments():
-        u[segm], v[segm], w[segm] = target.uvw(antennaA, cache.timestamps[segm],
-                                               antennaB)
-    cache['Corrprods/%s/%s/u' % (antA, antB)] = u
-    cache['Corrprods/%s/%s/v' % (antA, antB)] = v
-    cache['Corrprods/%s/%s/w' % (antA, antB)] = w
-    return u if name.endswith('u') else v if name.endswith('v') else w
+        basis = target.uvw_basis(cache.timestamps[segm], antenna)
+        u[segm], v[segm], w[segm] = basis.transpose(0, 2, 1)
```
bmerry: Maybe worth indicating which axes you're swapping in a comment.

Also, is it necessary to permute here, rather than feeding a different axis to tensordot below?

katdal/dataset.py
```diff
@@ -1025,9 +1039,11 @@ def sensor_data(ant_name):
 
     def _sensor_per_corrprod(self, base_name):
         """Extract a single sensor per corrprod and safely stack the results."""
```
bmerry: Maybe rename this, since the implementation is now very specific to values where the corrprod result is the difference of per-antenna results.

katdal/test/test_dataset.py
ludwigschwardt: Good point.

katdal/dataset.py
ludwigschwardt: I wondered about that too.

katdal/test/test_dataset.py
ludwigschwardt: Now fixed.

katdal/dataset.py
ludwigschwardt: Renamed.

katdal/dataset.py
ludwigschwardt: Added comments to explain the reshaping. It is necessary to permute, since a sensor expressed as an array must have time as the first dimension. On the other hand, I've now realised that the cargo-copied tensordot is overkill - a simple `dot` will do since `basis` is a T x 3 array.

ludwigschwardt:
Should be good to check again.

katdal/h5datav3.py
```diff
@@ -422,6 +422,10 @@ def register_telstate_sensor(name, obj):
         # Store antenna objects in sensor cache too, for use in virtual sensor calculations
         for ant in ants:
             self.sensor['Antennas/%s/antenna' % (ant.name,)] = CategoricalData([ant], [0, num_dumps])
+        # Extract array reference from first antenna (first 5 fields of description)
+        array_ant_fields = ['array'] + ants[0].description.split(',')[1:5]
+        array_ant = katpoint.Antenna(','.join(array_ant_fields))
```
bmerry: This is getting copy-pasted a lot. I'll approve for now, but I'd really like to see a helper function in katpoint for this.

ludwigschwardt: Indeed. I've made a note of this in SR-1619.

PR #253, opened 2019-07-19
Make L2: preparations

ludwigschwardt:
This updates "make L1" to be closer to what we need for L2:

- Remove hard-coded `cal` stream names. Make it part of the calibration product name instead.
- Remove `has_cal_product` which will be hard to support for virtual cal product sensors.
- Rename the virtual correction sensors to improve matches to funny stream names.
- Add virtual product sensors to standardise stream names to something like `l1` and `l2`.
- Minor doc updates and refactors.

So a "cal product" is now an amalgam of a cal stream (`l1`) and a product type (`G`), with a name like `l1.G`. The streams may be actual or virtual. The `add_applycal_sensors` function operates on a single cal stream at a time.

(EDIT: renamed `L1` to `l1`)

### DIFF
--- a/katdal/applycal.py
+++ b/katdal/applycal.py
@@ -16,17 +16,11 @@
 
 """Utilities for applying calibration solutions to visibilities and weights."""
 from __future__ import print_function, division, absolute_import
-from builtins import range, zip
 
 import logging
-import itertools
-import operator
 
 import numpy as np
 import dask.array as da
-import dask.base
-import dask.utils
-import toolz
 import numba
 
 from .categorical import CategoricalData, ComparableArrayWrapper
@@ -397,7 +391,8 @@ def calc_correction(chunks, cache, corrprods, cal_products):
             products[product].append(data)
     params = CorrectionParams(inputs, input1_index, input2_index, products)
     name = 'corrections[{}]'.format(','.join(cal_products))
-    return da.map_blocks(_correction_block, chunks=chunks, dtype=np.complex64, params=params)
+    return da.map_blocks(_correction_block, dtype=np.complex64, chunks=chunks,
+                         name=name, params=params)
 
 
 @numba.jit(nopython=True, nogil=True)
--- a/katdal/applycal.py
+++ b/katdal/applycal.py
@@ -109,7 +109,7 @@ def get_cal_product(cache, attrs, product_type, cal_stream='cal'):
     product_type : string
         Calibration product type (e.g. "G")
     cal_stream : string, optional
-        Name of calibration stream (e.g. "L1")
+        Name of calibration stream (e.g. "l1")
     """
     sensor_name = 'Calibration/Products/{}/{}'.format(cal_stream, product_type)
     try:
@@ -246,7 +246,7 @@ def add_applycal_sensors(cache, attrs, data_freqs, cal_stream='cal',
     data_freqs : array of float, shape (*F*,)
         Centre frequency of each frequency channel of visibilities, in Hz
     cal_stream : string, optional
-        Name of (possibly virtual) calibration stream (e.g. "L1")
+        Name of (possibly virtual) calibration stream (e.g. "l1")
     cal_substreams : sequence of string, optional
         Names of actual underlying calibration streams (e.g. ["cal"]),
         defaults to [`cal_stream`] itself
--- a/katdal/applycal.py
+++ b/katdal/applycal.py
@@ -220,6 +220,15 @@ def add_applycal_sensors(cache, attrs, data_freqs):
     function picks the appropriate correction calculator based on the cal
     product name, which also uses auxiliary info like the channel frequencies,
     `data_freqs`.
+
+    Parameters
+    ----------
+    cache : :class:`~katdal.sensordata.SensorCache` object
+        Sensor cache serving cal product sensors and receiving correction sensors
+    attrs : dict-like
+        Calibration product attributes (e.g. a "cal" telstate view)
+    data_freqs : array of float, shape (*F*,)
+        Centre frequency of each frequency channel of visibilities, in Hz
     """
     cal_ants = attrs.get('cal_antlist', [])
     cal_pols = attrs.get('cal_pol_ordering', [])
@@ -280,20 +289,20 @@ class CorrectionParams(object):
 
     Parameters
     ----------
-    products : dict
+    inputs : list of str
+        Names of inputs, in the same order as the input axis of products
+    input1_index, input2_index : array of int
+        Indices into `inputs` of first and second items of correlation product
+    corrections : dict
         A dictionary (indexed by cal product name) of lists (indexed
         by input) of sequences (indexed by dump) of numpy arrays, with
         corrections to apply.
-    inputs : list of str
-        Names of inputs, in the same order as the input axis of products
-    input1_index, input2_index : ndarray
-        Indices into `inputs` of first and second items of correlation product
     """
-    def __init__(self, inputs, input1_index, input2_index, products):
+    def __init__(self, inputs, input1_index, input2_index, corrections):
         self.inputs = inputs
         self.input1_index = input1_index
         self.input2_index = input2_index
-        self.products = products
+        self.corrections = corrections
 
 
 def calc_correction_per_corrprod(dump, channels, params):
@@ -311,7 +320,7 @@ def calc_correction_per_corrprod(dump, channels, params):
     channels : slice
         Channel indices (applicable to full data set, i.e. absolute)
     params : :class:`CorrectionParams`
-        Data for obtaining corrections to merge
+        Corrections per input, together with correlation product indices
 
     Returns
     -------
@@ -325,13 +334,13 @@ def calc_correction_per_corrprod(dump, channels, params):
     """
     n_channels = channels.stop - channels.start
     g_per_input = np.ones((len(params.inputs), n_channels), dtype='complex64')
-    for product in params.products.values():
-        for n in range(len(params.inputs)):
-            sensor = product[n]
+    for product_corrections in params.corrections.values():
+        for i in range(len(params.inputs)):
+            sensor = product_corrections[i]
             g_product = sensor[dump]
             if np.shape(g_product) != ():
                 g_product = g_product[channels]
-            g_per_input[n] *= g_product
+            g_per_input[i] *= g_product
     # Transpose to (channel, input) order, and ensure C ordering
     g_per_input = np.ascontiguousarray(g_per_input.T)
     g_per_cp = np.empty((n_channels, len(params.input1_index)), dtype='complex64')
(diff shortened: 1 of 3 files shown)


katdal/visdatav4.py
```diff
@@ -82,6 +82,8 @@ def _add_sensor_alias(cache, new_name, old_name):
 VIRTUAL_SENSORS.update({'Antennas/{ant}/az': _calc_azel,
                         'Antennas/{ant}/el': _calc_azel})
 
+DEFAULT_CAL_PRODUCTS = ('L1.K', 'L1.B', 'L1.G')
```
bmerry: Should we have some consistency in the capitalisation of L1? In the stream names it is sdp_l1 (lower-case).

katdal/applycal.py
```diff
@@ -92,42 +86,50 @@ def complex_interp(x, xi, yi, left=None, right=None):
...
-        parts = int(attrs[key + '_parts'])
-    except KeyError:
-        return key in cache
-    else:
-        return any(key + str(part) in cache for part in range(parts))
+def _parse_cal_product(cal_product):
+    """Split `cal_product` into `cal_stream` and `product_type` parts."""
+    fields = cal_product.rsplit('.', 1)
+    product_type = fields.pop()
+    cal_stream = fields.pop() if fields else 'cal'
```
bmerry: Should the default be L1, since the default list is L1.K, L1.B, L1.G?

katdal/applycal.py
```diff
@@ -216,57 +219,88 @@ def calc_gain_correction(sensor, index):
...
+        Sensor cache serving cal product sensors and receiving correction sensors
+    attrs : dict-like
+        Calibration stream attributes (e.g. a "cal" telstate view)
+    data_freqs : array of float, shape (*F*,)
+        Centre frequency of each frequency channel of visibilities, in Hz
+    cal_stream : string, optional
+        Name of (possibly virtual) calibration stream (e.g. "L1")
+    cal_substreams : sequence of string, optional
+        Names of actual underlying calibration streams (e.g. ["cal"]),
+        defaults to [`cal_stream`] itself
```
bmerry: It looks like it currently only uses the first element. Is this one of those places that you expect to extend for L2?

katdal/applycal.py
```diff
@@ -216,57 +219,88 @@ def calc_gain_correction(sensor, index):
...
     except KeyError:
-        logger.warning('Missing cal spectral attributes, disabling applycal')
+        logger.warning("Disabling cal stream '%s' due to missing "
+                       "spectral attributes", cal_stream)
         return
 
-    def calc_correction_per_input(cache, name, inp, product):
+    def indirect_cal_product(cache, name, product_type):
+        # XXX The first underscore below is actually a telstate separator...
+        return cache.get('{}_product_{}'.format(cal_substreams[0], product_type))
```
bmerry: If you want to avoid the separator, could you instead use `attrs.get('product_{}'.format(product_type))`? That would bypass the cache of the raw attribute, but it presumably still be cached in the virtual sensor? And for now it would also remove the need to pass cal_substreams at all, although I assume you have other plans for it for L2.

Or are you trying to avoid conflating the use of `attrs` for metadata with access to the solutions?

katdal/applycal.py
```diff
@@ -394,10 +451,15 @@ def calc_correction(chunks, cache, corrprods, cal_products):
...
-    params = CorrectionParams(inputs, input1_index, input2_index, products)
-    name = 'corrections[{}]'.format(','.join(cal_products))
-    return da.map_blocks(_correction_block, chunks=chunks, dtype=np.complex64, params=params)
+            corrections_per_product.append(data)
+        else:
+            corrections[cal_product] = corrections_per_product
+    if not corrections:
+        return None
+    params = CorrectionParams(inputs, input1_index, input2_index, corrections)
+    name = 'corrections[{}]'.format(','.join(corrections.keys()))
```
bmerry: Maybe worth sorting keys to give reproducible ordering (particularly since 2.7, 3.5 and 3.6+ will all do different things)?

katdal/test/test_applycal.py
```diff
@@ -66,11 +65,11 @@
 GAIN_EVENTS = list(range(0, N_DUMPS, 10))
 BAD_GAIN_ANT = 3
 BAD_GAIN_DUMPS = [20, 40]
-CAL_PRODUCTS = ['K', 'B', 'G']
+CAL_PRODUCTS = ['cal.K', 'cal.B', 'cal.G']
```
bmerry: This looks different to what's actually used by visdatav4.

katdal/test/test_applycal.py
```diff
@@ -345,40 +331,42 @@ def test_add_sensors_does_nothing_if_no_ants_pols_or_spw(self):
...
-            self.cache.get('Calibration/unknown_correction_K')
+            self.cache.get('Calibration/Corrections/cal/K/unknown')
         with assert_raises(KeyError):
-            self.cache.get(known_input + '_correction_unknown')
+            self.cache.get('Calibration/Corrections/cal/unknown/' + known_input)
         with assert_raises(KeyError):
-            self.cache.get(known_input + '_correction_K_unknown')
+            self.cache.get('Calibration/Corrections/cal/K_unknown/' + known_input)
+        with assert_raises(KeyError):
+            self.cache.get('Calibration/Corrections/unknown/K/' + known_input)
```
bmerry: Should this test be extended to test the Calibration/Product namespace too?

katdal/visdatav4.py
ludwigschwardt: A valid question... Every other place I could find has L1 capitalised (including the docs), and since we have the glorious standard of all-caps product types I thought this way the cal product would have a homogenous look at last 😛But the underlying streams tend to be (mostly) lower-case. I'll study the colour chart for the shed some more.

katdal/applycal.py
ludwigschwardt: I initially aimed that default at the unit tests to avoid fiddling with them too much. I could remove it here and enforce it on the `visdatav4` level instead.

My thinking is that `visdatav4` is the place to analyse `sdp_archived_streams`, find `sdp.cal` streams and construct "L1" and "L2" combinations, while `applycal` is slightly lower-level and only knows about the stream internals.

katdal/applycal.py
ludwigschwardt: Yes.

katdal/applycal.py
ludwigschwardt: The problem is that `attrs['product_G']` is a telstate mutable on which you need to do `.get_range`, while `cache['cal_product_G']` is a `SensorData` wrapper that can be extracted to a `CategoricalData`, which is what you really need in the applycal machinery. And `cache` does not have views (yet), which leads to this code. I wanted to avoid sensor extraction inside `applycal`.  (And yes, L2 will need the substreams.)

katdal/test/test_applycal.py
ludwigschwardt: Sure, but it's a unit test so I figured I won't add indirection yet.

katdal/test/test_applycal.py
ludwigschwardt: Good idea.

katdal/applycal.py
ludwigschwardt: The name is only for display / debugging purposes, but this sounds reasonable.

katdal/applycal.py
```diff
@@ -86,17 +86,6 @@ def complex_interp(x, xi, yi, left=None, right=None):
     return y.astype(yi.dtype)
 
 
-def has_cal_product(cache, attrs, product):
```
ludwigschwardt: I have a question on this...

I removed this because I could not check `Calibration/Products` virtual sensors. I still have the opportunity to scan through the keys of the underlying stream(s) and find all products up front though, which could be useful for debugging / diagnostics. This would be simpler if `telstate.keys()` were view-aware instead of listing the global keys (rather like `telstate._ipython_key_completions_` 😄). What was the reasoning behind this again?

bmerry: I think we just decided that there was no obvious right answer for what "view-aware" would mean. In particular, in a telstate with views, some keys could be inaccessible because they're shadowed, while others may be reachable via multiple names.

But I think `key in telstate` should be view-aware, which should be all you need, right?

ludwigschwardt: I was thinking of iterating over the keys and finding all the ones that start with `product_`, but I would have done this on an exclusive "cal" view so there is no issue of shadowing or duplicates.

bmerry: With an exclusive view I guess it would be well-defined.

I would hesitate to change the behaviour of `keys` at this point (it'll be very difficult to find all the places it gets called), but maybe it would make sense to introduce another function that has behaviour similar to the IPython completer that could be used for this.

katdal/visdatav4.py
ludwigschwardt: The consensus is `l1` so there you go :smile:

katdal/applycal.py
bmerry: In that case, maybe the higher level should also supply the default. I'm guessing that at the moment, passing `applycal=['K']` to katdal.open is going to give an error about cal.K not being found?

katdal/applycal.py
```diff
@@ -92,42 +86,50 @@ def complex_interp(x, xi, yi, left=None, right=None):
...
+    """Split `cal_product` into `cal_stream` and `product_type` parts."""
+    fields = cal_product.rsplit('.', 1)
+    product_type = fields.pop()
+    cal_stream = fields.pop() if fields else 'cal'
+    return cal_stream, product_type
 
 
-def get_cal_product(cache, attrs, product):
-    """Extract calibration solution `product` from `cache` as a sensor.
+def get_cal_product(cache, attrs, product_type, cal_stream='cal'):
```
bmerry: Perhaps `cal_stream` shouldn't have a default if the default is wrong for every case except unit tests.

katdal/applycal.py
bmerry: It's not just for debugging purposes. It forms part of the dask key, which dask uses to decide when two things are the same and don't need to be recomputed. I don't know enough about dask.distributed to know if this is actually a concern in practice, but I think it's at least possible for two processes in a dask cluster to get different orders and hence do unnecessary recomputation.

katdal/test/test_applycal.py
bmerry: So do you have a separate test for indirection somewhere?

katdal/applycal.py
ludwigschwardt: Good point, forgot about that!

katdal/applycal.py
ludwigschwardt: Agreed. The previous comment is also True about `_parse_cal_product`. I'll rework the defaults.

katdal/test/test_applycal.py
ludwigschwardt: I'll add something when I rework the default stream.

ludwigschwardt:
Ready to check again.

ludwigschwardt:
Do you mean having different product contents and not simply another name? I think that will get tested once L2 proper arrives.

bmerry:
> Do you mean having different product contents and not simply another name? I think that will get tested once L2 proper arrives.

I meant just having different names. But I'm happy to wait until L2 arrives and get it tested then.

bmerry:
Sorry, forgot to explicitly approve. Done now.

PR #255, opened 2019-07-30
Make L2: the Prestige

ludwigschwardt:
Hook up the L2 stream to applycal:

- Synthesise the indirect L2 products (GPHASE and GAMP_PHASE) by merging the raw sensor data of the underlying cal substreams. Do this by interleaving the timestamps of each sensor to form a combined raw `RecordSensorData` object.

- Register sensors for the standard "l1" and "l2" cal streams, if found. The "l1" stream is the first "sdp.cal" stream found in the list of archived streams (or "cal"), and the "l2" stream is derived from the first "sdp.continuum_image" stream in the list.

- Parse the string supplied to the `applycal` parameter of `katdal.open` and turn it into a list of fully qualified cal stream names. Both product types and cal streams may be specified independently. Also add a "default" grouping (which should eventually become the default value for the parameter) that applies L1 and phase-only selfcal for L2, if found. It's a ValueError if the cal product can't be normalised. And allow missing products to be skipped if any product in the applycal string is incomplete or a grouping.

### DIFF
--- a/katdal/visdatav4.py
+++ b/katdal/visdatav4.py
@@ -34,7 +34,7 @@ from .categorical import CategoricalData
 from .lazy_indexer import DaskLazyIndexer
 from .applycal import (add_applycal_sensors, calc_correction,
                        apply_vis_correction, apply_weights_correction,
-                       apply_flags_correction)
+                       apply_flags_correction, CAL_PRODUCT_TYPES)
 from .flags import NAMES as FLAG_NAMES, DESCRIPTIONS as FLAG_DESCRIPTIONS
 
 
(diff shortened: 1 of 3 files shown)


katdal/visdatav4.py
```diff
@@ -78,11 +78,36 @@ def _add_sensor_alias(cache, new_name, old_name):
         pass
 
 
+def _normalise_cal_products(products, cal_streams):
+    """"""
+    requested_cal_products = _selection_to_list(products, all=cal_streams,
+                                                default=DEFAULT_CAL_PRODUCTS)
+    skip_missing_products = products in ('all', 'default') or any(
+        ['.' not in product for product in requested_cal_products])
```
bmerry: The square brackets aren't needed: `any` works fine on generator expressions.

katdal/visdatav4.py
```diff
@@ -399,6 +418,37 @@ def _before(date):
         # on selection in the process
         self.select(spw=0, subarray=0, ants=obs_ants)
 
+    def _register_standard_cal_streams(self):
+        freqs = self.spectral_windows[0].channel_freqs
+        attrs = self.source.metadata.attrs
+        archived_streams = attrs.get('sdp_archived_streams', [])
+        # The default L1 cal stream, useful for older files
+        l1_stream = 'cal'
```
bmerry: What will happen with files that don't have any cal stream (e.g. because there are <4 antennas, or Mattieu has turned it off for holography)?

katdal/visdatav4.py
ludwigschwardt: Oops, yes.

katdal/visdatav4.py
ludwigschwardt: A 'cal' view will be established on telstate but it will be empty, prompting `add_applycal_sensors` to skip the virtual sensors and return `None`.

I improved it so that the `l1` stream won't be registered as an entry in `cal_freqs` either, which will help point out the missing stream when parsing user input.

ludwigschwardt:
The doc update is a good idea, but involves more than just adding GPHASE and GAMP_PHASE. I would argue that the description is already too detailed for the average user. I would want to describe the applycal machinery instead of explaining what `1560498744_continuum_image_PKS_0408-65_selfcal_product_GAMP_PHASE` means... 🙈 And probably move the detailed RDB layout to an advanced section.

bmerry:
The MVFv4 documentation was never intended for the general user. It's documentation of the internals of the file format and probably only useful for SDP developers and maybe a few power users. End-user howto documentation is certainly a good idea too.

ludwigschwardt:
Good, that makes sense then! I've opened a ticket (SR-1945) to document the rest.

PR #249, opened 2019-06-10
Added options -S or --s3_endpoint_url, and  -T or --token for mvftoms.py script

gigjozsa:
Allows the user to enter the corresponding kwarg s3_endpoint_url of the open method as command line argument 
--s3_endpoint_url theparameter
with the alias
-S theparameter
and 
it allows the user to enter the corresponding kwarg s3_endpoint_url of the open method as command line argument 
-- token theparameter
with the alias
-T theparameter
This is a requirement if someone wants to use mvftoms with newer data.

PR #258, opened 2019-08-08
Various fixes and improvements to MVFtoMS

ludwigschwardt:
This addresses a slew of recent tickets:

  - [SR-1828](https://skaafrica.atlassian.net/browse/SR-1828): let `mvftoms` open RDB URLs, including access tokens. It turns out that this is already possible, thereby removing the need for #249... I only needed to fix the construction of a default output MS name. One thing to note is that the MS now goes into the working directory as always promised, instead of next to the RDB file.

  - [SR-1811](https://skaafrica.atlassian.net/browse/SR-1811): Fix the optional REST_FREQUENCY column in optional SOURCE table. It is not optional for UVFITS export in CASA... Clean up the whole SOURCE sub-table. @sjperkins should have a good look at this.

  - [SR-1948](https://skaafrica.atlassian.net/browse/SR-1948): Add `--applycal` option to pass through to `katdal.open`. This allows MS data to be calibrated up front.

  - [SR-1654](https://skaafrica.atlassian.net/browse/SR-1654): Fix various help strings to reflect the MVFv4 reality (@bmerry was the original reporter).

  - Strip out casapy support for good measure.

### DIFF
--- a/katdal/ms_extra.py
+++ b/katdal/ms_extra.py
@@ -94,7 +94,7 @@ def define_hypercolumn(desc):
                                          for k, v in desc.items() if v['dataManagerType'] == 'TiledShapeStMan'])
 
 
-# Map Measurement Set string types to numpy types
+# Map MeasurementSet string types to numpy types
 MS_TO_NP_TYPE_MAP = {
     'INT': np.int32,
     'FLOAT': np.float32,
@@ -107,7 +107,7 @@ MS_TO_NP_TYPE_MAP = {
 
 def kat_ms_desc_and_dminfo(nbl, nchan, ncorr, model_data=False):
     """
-    Creates Table Description and Data Manager Information objecs that
+    Creates Table Description and Data Manager Information objects that
     describe a MeasurementSet suitable for holding MeerKAT data.
 
     Creates additional DATA, IMAGING_WEIGHT and possibly
@@ -117,7 +117,7 @@ def kat_ms_desc_and_dminfo(nbl, nchan, ncorr, model_data=False):
 
     :param nbl: Number of baselines.
     :param nchan: Number of channels.
-    :param ncorr: Number of correlations.
+    :param ncorr: Number of correlation products.
     :param model_data: Boolean indicated whether MODEL_DATA and CORRECTED_DATA
                         should be added to the Measurement Set.
     :return: Returns a tuple containing a table description describing
--- a/katdal/ms_extra.py
+++ b/katdal/ms_extra.py
@@ -981,18 +981,21 @@ def write_rows(t, row_dict, verbose=True):
     if verbose:
         print("  added %d rows" % (num_rows,))
     for col_name, col_data in row_dict.items():
-        if col_name in t.colnames():
-            if col_data.dtype.kind == 'U':
-                col_data = np.char.encode(col_data, encoding='utf-8')
-            try:
-                t.putcol(col_name, col_data, startrow)
-                if verbose:
-                    print("  wrote column '%s' with shape %s" % (col_name, col_data.shape))
-            except RuntimeError as err:
-                print("  error writing column '%s' with shape %s (%s)" % (col_name, col_data.shape, err))
-        else:
+        if col_name not in t.colnames():
             if verbose:
                 print("  column '%s' not in table" % (col_name,))
+            continue
+        if col_data.dtype.kind == 'U':
+            col_data = np.char.encode(col_data, encoding='utf-8')
+        try:
+            t.putcol(col_name, col_data, startrow)
+        except RuntimeError as err:
+            print("  error writing column '%s' with shape %s (%s)" %
+                  (col_name, col_data.shape, err))
+        else:
+            if verbose:
+                print("  wrote column '%s' with shape %s" %
+                      (col_name, col_data.shape))
 
 
 def write_dict(ms_dict, ms_name, verbose=True):
--- a/katdal/ms_extra.py
+++ b/katdal/ms_extra.py
@@ -762,29 +762,54 @@ def populate_spectral_window_dict(center_frequencies, channel_bandwidths, ref_fr
     return spectral_window_dict
 
 
-def populate_source_dict(phase_centers, time_origins, center_frequencies, field_names=None):
+def populate_source_dict(phase_centers, time_origins, field_names=None):
     """Construct a dictionary containing the columns of the SOURCE subtable.
 
     The SOURCE subtable describes time-variable source information, that may
     be associated with a given FIELD_ID. It appears to be optional, but for
-    completeness it is included here (with no time varying terms).
+    completeness it is included here (with no time varying terms). Some RARG
+    tasks and CASA's exportuvfits do require it, though.
+
+    Parameters
+    ----------
+    phase_centers : array of float, shape (M, 2)
+        Direction of *M* phase centers as (ra, dec) coordinates in radians
+    time_origins : array of float, shape (M,)
+        Time origins where the *M* phase centers are correct, as Modified Julian
+        Dates in seconds
+    field_names : array of string, shape (M,), optional
+        Names of fields/pointings (typically some source names)
+
+    Returns
+    -------
+    source_dict : dict
+        Dictionary containing columns of SOURCE subtable
+
     """
-    num_channels = len(center_frequencies)
     phase_centers = np.atleast_2d(np.asarray(phase_centers, np.float64))
     num_fields = len(phase_centers)
     if field_names is None:
         field_names = ['Source%d' % (field,) for field in range(num_fields)]
     source_dict = {}
+    # Source identifier as specified in the FIELD sub-table (integer)
     source_dict['SOURCE_ID'] = np.arange(num_fields, dtype=np.int32)
+    # Source proper motion in radians per second (double, 1-dim, shape=(2,))
     source_dict['PROPER_MOTION'] = np.zeros((num_fields, 2), dtype=np.float32)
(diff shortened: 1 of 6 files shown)


katdal/ms_extra.py
```diff
@@ -1026,9 +1015,11 @@ def write_dict(ms_dict, ms_name, verbose=True):
...
             if verbose:
                 print("Table %s:" % (sub_table_name,))
-            # Open table using whichever casacore library was found
-            t = open_table(ms_name, ack=verbose) if sub_table_name == 'MAIN' else \
-                open_table(os.path.join(ms_name, sub_table_name))
+            # Open main table or sub-table
+            if sub_table_name == 'MAIN':
+                t = open_table(ms_name, ack=verbose)
+            else:
+                t = open_table(os.path.join(ms_name, sub_table_name))
```
sjperkins: The convention is to separate table and sub-table with ::, "WSRT.MS::ANTENNA" for e.g.

```suggestion
                t = open_table("::"join((ms_name, sub_table_name)))
```

katdal/ms_extra.py
```diff
@@ -32,34 +28,39 @@
...
+if not pyc_ver >= req_ver:
+    raise ImportError("python-casacore %s is required, but the current version is %s. "
+                      "Note that python-casacore %s requires at least casacore 2.3.0."
+                      % (req_ver, pyc_ver, req_ver))
+
+
+def open_table(filename, readonly=False, ack=False, **kwargs):
+    """Open casacore Table."""
+    t = tables.table(filename, readonly=readonly, ack=ack, **kwargs)
+    return t if type(t) == tables.table else None
```
sjperkins: Could you explain why the type check is required? Why would the constructor return anything except a tables.table?

scripts/mvftoms.py
```diff
@@ -241,47 +252,57 @@ def _cp_index(a1, a2, pol):
...
+            if dataset_filename.endswith('.full.rdb'):
+                dataset_basename = dataset_filename[:-9]
+            else:
+                dataset_basename = os.path.splitext(dataset_filename)[0]
+            # Add frequency to name to disambiguate multiple spectral windows
+            if len(dataset.spectral_windows) > 1:
+                dataset_basename += '_{:d}MHz'.format(int(freq_MHz))
+            # Add ".et_al" as reminder that we concatenated multiple datasets
+            ms_name = '{}{}.ms'.format(dataset_basename,
+                                       "" if len(args) == 1 else ".et_al")
```
sjperkins: This filename generation logic looks like it could be in a function. The script itself has become quite large so it might be nice to start the process of moving this kind of stuff into individual modules.

scripts/mvftoms.py
```diff
@@ -241,47 +252,57 @@ def _cp_index(a1, a2, pol):
...
-            # create MS in current working directory
-            ms_name = '%s_%d.%s%s.ms' % (
-                os.path.splitext(args[0])[0], freq_MHz,
-                "" if len(args) == 1 else "et_al.", pol_for_name)
+            url_parts = urllib.parse.urlparse(args[0], scheme='file')
+            # Create MS in current working directory (strip off directories)
+            dataset_filename = os.path.basename(url_parts.path)
+            # Get rid of the ".full" bit on RDB files (it's the same dataset)
+            if dataset_filename.endswith('.full.rdb'):
+                dataset_basename = dataset_filename[:-9]
```
sjperkins:
```python
    full_rdb_ext = '.full.rdb'
    if dataset_filename.endswith(ext):
                dataset_basename = dataset_filename[:-len(ext)]
```

katdal/ms_extra.py
```diff
@@ -1006,7 +995,7 @@ def write_rows(t, row_dict, verbose=True):
             if col_data.dtype.kind == 'U':
                 col_data = np.char.encode(col_data, encoding='utf-8')
             try:
-                t.putcol(col_name, col_data.T if casacore_binding == 'casapy' else col_data, startrow)
+                t.putcol(col_name, col_data, startrow)
                 if verbose:
```
sjperkins: I would suggest using the else clause in the try-except-else-finally structure.

```python
try:
   t.putcol(...)
except RuntimeError as e:
   print(...)
else:
    if verbose:
       print(...)
```

katdal/ms_extra.py
ludwigschwardt: Thanks for the tip, done. Would you do the same inside `tables.default_ms_subtable()`?

sjperkins: > Thanks for the tip, done. Would you do the same inside `tables.default_ms_subtable()`?

No, ``default_ms_subtable("SOURCE")`` will simply create a ``SOURCE`` directory in the MS. Hence the need to manually add the keyword linking the main table to it.

katdal/ms_extra.py
ludwigschwardt: 😅 
I think that is a wart left over from supporting casapy. In CASA-land you have the `tb` tool, and `tb.open` returns True on successfully opening the table. I then returned either the `tb` tool itself as a proxy for the `table` object, or `None` otherwise. But as you correctly point out, this does not make sense in casacore-land. In retrospect I should have done
```Python
try:
    return tables.table(...)
except RunTimeError:
    return None
```
Now that casapy is off the table, rather let the exception be raised on error.

I've now reworked both `open_table` and `open_main` into one function with no `None`s in sight.

scripts/mvftoms.py
ludwigschwardt: The relevant individual module is called `ms_extra` and it already has more lines than the script 😛 

There is still a lot of dead code in both `ms_extra` and `mvftoms` (like the cal table generation), so I'm not too worried about the line count. I'll still move it to a function though - that sounds good.

scripts/mvftoms.py
ludwigschwardt: Done.

katdal/ms_extra.py
ludwigschwardt: Done (plus restore the Happy Path).

scripts/mvftoms.py
```diff
@@ -241,47 +271,48 @@ def _cp_index(a1, a2, pol):
...
-        # from the first dataset.
+        # If no output MS directory name supplied, infer it from dataset(s)
         if options.output_ms is None:
-            # create MS in current working directory
-            ms_name = '%s_%d.%s%s.ms' % (
-                os.path.splitext(args[0])[0], freq_MHz,
-                "" if len(args) == 1 else "et_al.", pol_for_name)
+            if len(dataset.spectral_windows) > 1:
+                # Use frequency label to disambiguate multiple spectral windows
+                ms_name = default_ms_name(args, freq_MHz)
```
bmerry: What if there are two spectral windows within 1MHz of each other?

scripts/mvftoms.py
```diff
@@ -50,6 +51,25 @@
 SLOTS = 4    # Controls overlap between loading and writing
 
 
+def default_ms_name(args, freq_MHz=None):
+    """Infer default MS name from argument list and optional frequency label."""
+    # Use the first dataset in the list to generate the base part of the MS name
+    url_parts = urllib.parse.urlparse(args[0], scheme='file')
+    # Create MS in current working directory (strip off directories)
+    dataset_filename = os.path.basename(url_parts.path)
```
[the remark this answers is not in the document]
ludwigschwardt: Good point, will fix.

scripts/mvftoms.py
```diff
@@ -241,47 +271,48 @@ def _cp_index(a1, a2, pol):
 
     # Open dataset
     open_args = args[0] if len(args) == 1 else args
-    # katdal can handle a list of files, which get virtually concatenated internally
-    dataset = katdal.open(open_args, ref_ant=options.ref_ant)
+    # katdal can handle a list of datasets, which get virtually concatenated internally
+    dataset = katdal.open(open_args, ref_ant=options.ref_ant, applycal=options.applycal)
+    applycal_products = ', '.join(getattr(dataset, '_applycal_products', []))
```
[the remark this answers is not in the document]
ludwigschwardt: I feel the same way but did not want to lump it into this PR. On the other hand, it's a rather small change, so on second thought, why not?

scripts/mvftoms.py
ludwigschwardt: Sure. This is a unlikely event (windows within 1 MHz) on top of an unlikely event (a dataset even having more than one spectral window). I'll add a few digits then. Kilohertz should be safe :-)

scripts/mvftoms.py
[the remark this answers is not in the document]
ludwigschwardt: What should I do if I'm on Windows and have `args[0] = r'\1234\3456.rdb'`? I could use `os.path` if the scheme is `file` instead, but would someone ever do `args[0] = 'file://1234/3456.rdb'` - is that supposed to be the same thing?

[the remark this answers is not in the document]
bmerry: Hmm, good question. I guess the issue is that we're relying on the behaviour of urlparse of turning a POSIX path into a `file` URL, but it doesn't do backslash-to-slash convention on Windows-style paths (at least not when run on Linux - I don't have a Windows machine to test). Maybe it's actually safest to leave as is for now. As far as I recall, forward-slash is actually a valid separator on Windows (just not the canonical one) so `os.path.basename` may actually do the right thing.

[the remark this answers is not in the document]
ludwigschwardt: Yes, it seems to do the right thing when looking at the output of the `ntpath` module.

PR #261, opened 2019-08-20
Update documentation for sdp.cal keys

ludwigschwardt:
Spell out the shapes and dtypes of all calibration products.

Correct the description of `refant` which is soon the antenna name and not an antenna object description.

### DIFF
--- a/doc/mvf_v4.rst
+++ b/doc/mvf_v4.rst
@@ -225,15 +225,15 @@ Calibration solutions
 
 Streams of type ``sdp.cal`` have the following keys.
 
-``antlist`` (list of string, length ``n_ants``)
+``antlist`` (list of string, length n_ants)
     List of antenna names. Arrays of calibration solutions use this
     order along the antenna axis.
 
-``pol_ordering`` (list of string, length ``n_pols``)
+``pol_ordering`` (list of string, length n_pols)
     List of polarisations (from ``v`` and ``h``). Arrays of calibration
     solutions use this order along the polarisation axis.
 
-``bls_ordering`` (list of string pairs, length ``n_bls``)
+``bls_ordering`` (list of string pairs, length n_bls)
     Same meaning as for ``sdp.vis`` streams, but describes the internal
     ordering used within the calibration pipeline and not of much use to
     users.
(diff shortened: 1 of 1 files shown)


doc/mvf_v4.rst
```diff
@@ -225,36 +225,42 @@ Calibration solutions
...
 Streams of type ``sdp.cal`` have the following keys.
 
-``antlist`` (list of string)
+``antlist`` (list of string, length ``n_ants``)
     List of antenna names. Arrays of calibration solutions use this
     order along the antenna axis.
 
-``pol_ordering`` (list of string)
+``pol_ordering`` (list of string, length ``n_pols``)
     List of polarisations (from ``v`` and ``h``). Arrays of calibration
```
spassmoor: Just curious, are the polarisations hard-coded to be h & v?  Can they be  R and L or more than 2  eg. IQUV?

bmerry: I suppose in theory if we ever build a circular-polarisation dish they could be R&L, but that's not defined for now and I don't expect software to handle such values.

These are for inputs, not correlation products, so I don't think more than two or Stokes parameters makes any sense.

doc/mvf_v4.rst
```diff
@@ -225,36 +225,42 @@ Calibration solutions
...
     solutions use this order along the polarisation axis.
 
-``bls_ordering`` (2D array)
+``bls_ordering`` (list of string pairs, length ``n_bls``)
     Same meaning as for ``sdp.vis`` streams, but describes the internal
     ordering used within the calibration pipeline and not of much use to
     users.
```
bmerry: Interesting, I see sdp.vis bls_ordering is a numpy array while cal's bls_ordering is a list of lists of strings. Is that worth drawing attention to?

doc/mvf_v4.rst
```diff
@@ -225,36 +225,42 @@ Calibration solutions
...
 
-``product_G`` (2D array) — sensor
-    Gain solutions (derived e.g. on a phase calibrator), indexed by antenna
-    and polarisation. The complex values in the array apply to the entire band.
+``refant`` (string)
+    Name of the selected reference antenna (which will also appear in
+    ``antlist``). The reference antenna is only chosen when first needed in
+    a capture block, so this key may be absent if there was no calibration yet.
+
+``product_G`` (array of complex64, shape (``n_pols``, ``n_ants``)) — sensor
```
bmerry: Maybe we should just say array of complex. While we currently choose to use a particular precision, I don't think it's part of the interface contract.

doc/mvf_v4.rst
```diff
@@ -225,36 +225,42 @@ Calibration solutions
...
 ``param_*``
     Parameters used to configure the calibration.
 
-``product_G`` (2D array) — sensor
-    Gain solutions (derived e.g. on a phase calibrator), indexed by antenna
-    and polarisation. The complex values in the array apply to the entire band.
+``refant`` (string)
+    Name of the selected reference antenna (which will also appear in
+    ``antlist``). The reference antenna is only chosen when first needed in
+    a capture block, so this key may be absent if there was no calibration yet.
```
bmerry: Need to document that older files will have a description string instead of just a name here.

ludwigschwardt: Done.

doc/mvf_v4.rst
ludwigschwardt: Done.

doc/mvf_v4.rst
ludwigschwardt: Just one of the quirks of our namespace :-)

Should I change it to a "sequence of string pairs" to highlight the intended interface?

bmerry: Maybe. Or else just say that it's "either a list of string pairs or a 2D array" to make it clear that the type is uncertain.

ludwigschwardt: Done (for both instances of `bls_ordering`).

PR #265, opened 2019-10-01
Support katstore64 API

ludwigschwardt:
Since CAM v24 release the katstore API has changed to katstore64. This does away with the need to query the sensor KATCP type, since the HTTP request now returns a JSON string with the appropriate value type. Having one HTTP request instead of two speeds things up, on top of the improved performance of katstore64 itself.

This addresses JIRA ticket [SR-1962](https://skaafrica.atlassian.net/browse/SR-1962).


ludwigschwardt:
This seems to be working so I'm merging.

PR #272, opened 2019-11-20
SPR1-181: initial refactor

ludwigschwardt:
I'm splitting the chunk store rework into at least four PRs to lighten your load...

1. Initial refactor to get rid of icky layers of code [THIS ONE]
2. Token checking and authorisation error reporting (done, waiting in the wings)
3. Timeout and retries rework (current activity)
4. Collect request response statistics for debugging (future work)

Some of these bits (like da08290) are reworked in the future PRs, but not by a lot.

@sratcliffe, you'll like this one because it addresses your concerns in #235 😄

### DIFF
--- a/katdal/chunkstore_s3.py
+++ b/katdal/chunkstore_s3.py
@@ -142,6 +142,22 @@ class _AWSAuth(requests.auth.AuthBase):
         return r
 
 
+def _auth_factory(url, token=None, credentials=None):
+    """Turn either JWT token or AWS credentials into a requests auth handler."""
+    if token is not None and credentials is not None:
+        raise ValueError('Cannot specify both token and credentials')
+    if token is not None:
+        parsed = urllib.parse.urlparse(url)
+        # The exception for 127.0.0.1 lets the unit test work
+        if parsed.scheme != 'https' and parsed.hostname != '127.0.0.1':
+            raise StoreUnavailable('Auth token may only be used with https')
+        return _BearerAuth(token)
+    elif credentials is not None:
+        if not botocore:
+            raise StoreUnavailable('Passing credentials requires botocore to be installed')
+        return _AWSAuth(credentials)
+
+
 class _CacheSettingsSession(requests.Session):
     """Session that caches the result of proxy lookup.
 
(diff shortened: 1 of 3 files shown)


katdal/chunkstore_s3.py
```diff
@@ -118,23 +112,12 @@ def read_array(fp):
...
-        super(_TimeoutHTTPAdapter, self).__init__(*args, **kwargs)
-
-    def send(self, request, stream=False, timeout=None, *args, **kwargs):
-        if timeout is None:
-            timeout = self._default_timeout
-        return super(_TimeoutHTTPAdapter, self).send(request, stream, timeout, *args, **kwargs)
-
-
 class _BearerAuth(requests.auth.AuthBase):
+    """Add bearer token to authorisation request header."""
```
bmerry: Blank line after class docstring.

katdal/chunkstore_s3.py
```diff
@@ -143,6 +126,7 @@ def __call__(self, r):
 
 
 class _AWSAuth(requests.auth.AuthBase):
+    """Add AWS access + secret credentials to authorisation request header."""
```
bmerry: Blank line after class docstring.

katdal/chunkstore_s3.py
```diff
@@ -156,24 +140,68 @@ def __call__(self, r):
...
+    if token is not None:
+        parsed = urllib.parse.urlparse(url)
+        # The exception for 127.0.0.1 lets the unit test work
+        if parsed.scheme != 'https' and parsed.hostname != '127.0.0.1':
+            raise StoreUnavailable('Auth token may only be used with https')
+        return _BearerAuth(token)
+    elif credentials is not None:
+        if not botocore:
+            raise StoreUnavailable('Passing credentials requires botocore to be installed')
+        return _AWSAuth(credentials)
```
bmerry: Add an explicit `return None` when neither type of auth is provided: PEP 8 says

> Be consistent in return statements. Either all return statements in a function should return an expression, or none of them should. If any return statement returns an expression, any return statements where no value is returned should explicitly state this as return None, and an explicit return statement should be present at the end of the function (if reachable).

katdal/chunkstore_s3.py
```diff
@@ -270,139 +282,60 @@ class S3ChunkStore(ChunkStore):
...
-        A callable called with no arguments that returns an instance of
-        :class:`requests.Session`. The returned session is only used by
-        one thread at a time.
     url : str
-        Base URL for the S3 service. It can be specified as either bytes or
-        unicode, and is converted to the native string type with UTF-8.
+        Endpoint of S3 service, e.g. 'http://127.0.0.1:9000'. It can be
+        specified as either bytes or unicode, and is converted to the native
+        string type with UTF-8.
+    timeout : int or float, optional
```
bmerry: If you're documenting this as optional, then all the subsequent parameters should probably be documented as optional too.

katdal/chunkstore_s3.py
ludwigschwardt: Fixed.

katdal/chunkstore_s3.py
ludwigschwardt: Fixed.

katdal/chunkstore_s3.py
ludwigschwardt: Fixed.

katdal/chunkstore_s3.py

ludwigschwardt:
Merging @sratcliffe - comments still welcome :-)

PR #273, opened 2019-11-21
SPR1-181: Check tokens

ludwigschwardt:
Yes, it overlaps a bit with [PyJWT](https://github.com/jpadilla/pyjwt)...

My initial attempts only decoded JWT payloads, but then I had to add the encoders for the sake of unit tests 🤦‍♂  At least I'm not doing any cryptography (and I learned a lot in the process).

There is now a new type of `StoreUnavailable` called `AuthorisationFailed` to help the user. The token is checked for expiration, allowed prefixes and the correct length.

Remap HTTP response statuses:
- 401 -> AuthorisationFailed
- 403 + 404 on top-level directory / bucket level -> StoreUnavailable
- 403 + 404 below bucket level -> ChunkNotFound

On an aside, the strict token prefix check now prevents root directory access, even though the proxy server allows anyone to list the root dir as a special case. This breaks the smoke test, but that's OK since the smoke test has outlived its usefulness anyway and is slated for imminent  removal. The special case is troublesome since it always succeeds, even with a bad or missing token.


katdal/test/test_chunkstore_s3.py
```diff
@@ -361,4 +372,7 @@ def from_url(cls, url, authenticate=True, **kwargs):
             cls.proxy_url = 'http://{}:{}'.format(proxy_host, proxy_port)
         elif url != cls.httpd.target:
             raise RuntimeError('Cannot use multiple target URLs with http proxy')
-        return S3ChunkStore(cls.proxy_url, timeout=10, token='mysecret', **kwargs)
+        # For now this token authorises all prefixes (with dud signature of expected length)
+        token = encode_jwt({'alg': 'ES256', 'typ': 'JWT'},
+                           {'prefix': ['']}, 64 * b'*')
+        return S3ChunkStore(cls.proxy_url, timeout=10, token=token, **kwargs)
```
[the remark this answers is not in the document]
ludwigschwardt: `InvalidToken` is now a subclass of `AuthorisationFailed` to address this.

ludwigschwardt:
Rewrote the token routines based on PyJWT: -166 lines 😄

ludwigschwardt:
Now also removed the smoke test so that this PR can be merged.

requirements.txt
```diff
@@ -18,6 +18,7 @@ netifaces
 numba
 numpy
 pyephem
+pyjwt==1.7.1
```
bmerry: Could you add this to katsdpdockerbase - we're going to need to use it in all packages that use katdal.

katdal/chunkstore_s3.py
```diff
@@ -112,16 +112,94 @@ def read_array(fp):
...
         self._token = token
 
     def __call__(self, r):
+        # Check if token authorises URL even before hitting server for better reporting
+        path = urllib.parse.urlparse(r.url).path.lstrip('/')
+        valid_prefixes = self._claims['prefix']
+        if not any(path.startswith(prefix) for prefix in valid_prefixes):
+            allowed = ', '.join(prefix + '*' for prefix in valid_prefixes)
+            raise InvalidToken(self._token, 'Token does not grant permission for '
+                                            '{}, only for {}'.format(path, allowed))
```
bmerry: In the Jenkins coverage report it seems that this check is never tested.

katdal/chunkstore_s3.py
```diff
@@ -112,16 +112,94 @@ def read_array(fp):
...
+
+
+def decode_jwt(token):
+    """Decode JSON Web Token (JWT) string and extract claims.
+
+    The MeerKAT archive uses JWT bearer tokens for authorisation. Each token is
+    a JSON Web Signature (JWS) string with a payload of claims. This function
+    extracts the claims as a dict, while also doing basic checks on the token
+    (mostly to catch copy-n-paste errors). The signature is decoded but not
+    validated, since that would require the server secrets.
```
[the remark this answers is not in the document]
ludwigschwardt: You're right - I forgot that the verification part is based on public knowledge (only making a signature is restricted). I agree with your summary.

katdal/test/test_chunkstore_s3.py
```diff
@@ -125,6 +128,43 @@ def testShort(self):
...
+    token_bytes = jwt.encode(payload, '', algorithm='none', headers=header)
+    return bytes_to_native_str(token_bytes) + signature
+
+
+class TestTokenUtils(object):
+    """Test token utility and validation functions."""
+
+    def test_jwt_broken_token(self):
+        header = {'alg': 'ES256', 'typ': 'JWT'}
+        payload = {'exp': 9234567890, 'iss': 'kat', 'prefix': ['123']}
```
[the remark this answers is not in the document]
ludwigschwardt: I would have thought that 2262 is far enough in the future, or are you concerned that we'll still be using katdal then? 😛

requirements.txt
ludwigschwardt: Can I first add it to katsdpdockerbase, and then come back and remove the pin here? Does the order matter?

katdal/test/test_chunkstore_s3.py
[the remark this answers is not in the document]
bmerry: I just miscounted the digits.

requirements.txt
bmerry: If you remove it here before adding it to katsdpdockerbase, the build of this will fail. So I'd recommend adding it to katsdpdockerbase first.

ludwigschwardt: Done.

katdal/chunkstore_s3.py
ludwigschwardt: This is now also tested by replacing the allow-all token in the tests with a bucket-specific one.

PR #275, opened 2020-01-16
Make mark_complete work on whole bucket again

bmerry:
PR #274 broke `mark_complete` when run on a whole bucket, because it
calls `create_array` which now crashes if the "array_name" doesn't have
a separator.

This keeps the behaviour for `create_array`, but adds a backdoor
`_create_bucket` so that `mark_complete` can once again put a completion
marker at the bucket level.

### DIFF
--- a/katdal/chunkstore_s3.py
+++ b/katdal/chunkstore_s3.py
@@ -613,6 +613,9 @@ class S3ChunkStore(ChunkStore):
         """See the docstring of :meth:`ChunkStore.create_array`."""
         # Array name is formatted as bucket/array but we only need to create bucket
         bucket, _ = self.split(array_name, 1)
+        self._create_bucket(bucket)
+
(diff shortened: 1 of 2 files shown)


bmerry:
@ludwigschwardt I'm going to merge just to unblock our deployment. If you'd prefer a different approach feel free to make comments nevertheless.

katdal/chunkstore_s3.py
```diff
@@ -699,7 +702,8 @@ def _read_xml(response):
 
     def mark_complete(self, array_name):
         """See the docstring of :meth:`ChunkStore.mark_complete`."""
-        self.create_array(array_name)
+        bucket = self.split(array_name, 1)[0]
```
ludwigschwardt: Thanks! You could have just used this line in `create_array`, but I like the explicit `_create_bucket`.

PR #276, opened 2020-01-16
Increase S3 connection timeouts in unit tests

ludwigschwardt:
Since we are connecting to a local HTTP proxy server running on a
second Python thread in the same process, things can get a bit jerky.
Increase the connect timeout substantially to 5 seconds. Also redo the
unresponsive server test and give it its own (short) timeout without
retries to speed it up (no need to wait for 10 seconds).

This addresses JIRA ticket SPR1-364.

### DIFF
--- a/katdal/test/test_chunkstore_s3.py
+++ b/katdal/test/test_chunkstore_s3.py
@@ -65,11 +65,11 @@ from katdal.test.test_chunkstore import ChunkStoreTestBase
 
 BUCKET = 'katdal-unittest'
 # Pick quick but different timeouts and retries for unit tests:
-#  - The effective connect timeout is 0.2 (initial) + 0.2 (1 retry) = 0.4 seconds
+#  - The effective connect timeout is 5.0 (initial) + 5.0 (1 retry) = 10 seconds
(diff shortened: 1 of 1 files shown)


katdal/test/test_chunkstore_s3.py
```diff
@@ -301,14 +301,14 @@ def test_public_read(self):
         y = reader.get_chunk('public/x', slices, x.dtype)
         np.testing.assert_array_equal(x, y)
 
-    @timed(0.4 + 0.2)
+    @timed(0.1 + 0.2)
```
bmerry: The original bug was that you were making timeouts unnecessarily tight for no good reason, and now you're doing it again here...

ludwigschwardt: But this is the timeout for the timeout... 😛I want to check that the correct timeout is being used.

bmerry: Hmm, I see. There is no way to guarantee that your process won't get scheduled out for hundreds of milliseconds - a better test would be to using mocking to check what timeout is set on the socket. But we need to get this fixed ASAP, so carry on.

PR #277, opened 2020-01-16
Improve error message for expired tokens

ludwigschwardt:
The current error message is the default one provided by PyJWT which
reads "Signature has expired" - but what is a "signature"? To help the
end user, call it a token instead and include the expiry date/time for
handy reference. Also mention that the correct course of action is to
obtain a new one (thanks Ruby!).

This addresses JIRA ticket SR-1971.

### DIFF
--- a/katdal/chunkstore_s3.py
+++ b/katdal/chunkstore_s3.py
@@ -33,6 +33,7 @@ import base64
 import warnings
 import copy
 import json
+import time
 
 import defusedxml.ElementTree
 import defusedxml.cElementTree
@@ -244,6 +245,11 @@ def decode_jwt(token):
         # This covers e.g. bad characters in the signature or non-JSON-dict payload
         raise_from(InvalidToken(token, "Could not decode token - maybe it's truncated "
                                        "or corrupted? ({})".format(err)), err)
+    except jwt.exceptions.ExpiredSignatureError as err:
+        claims = jwt.decode(token, verify=False)
+        exp_time = time.strftime('%d-%b-%Y %H:%M:%S', time.gmtime(claims['exp']))
+        raise_from(InvalidToken(token, 'Token has expired on {} UTC, please '
+                                'obtain a new one'.format(exp_time)), err)
     except jwt.exceptions.InvalidTokenError as err:
         raise_from(InvalidToken(token, str(err)), err)
     return claims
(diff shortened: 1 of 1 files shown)


katdal/chunkstore_s3.py
```diff
@@ -244,6 +245,11 @@ def decode_jwt(token):
         # This covers e.g. bad characters in the signature or non-JSON-dict payload
         raise_from(InvalidToken(token, "Could not decode token - maybe it's truncated "
                                        "or corrupted? ({})".format(err)), err)
+    except jwt.exceptions.ExpiredSignatureError as err:
+        claims = jwt.decode(token, verify=False)
+        exp_time = time.strftime('%d-%b-%Y %H:%M:%S', time.gmtime(claims['exp']))
+        raise_from(InvalidToken(token, 'Token has expired on {} UTC, please '
```
bmerry:
```suggestion
        raise_from(InvalidToken(token, 'Token expired at {} UTC, please '
```

katdal/chunkstore_s3.py
```diff
@@ -244,6 +245,11 @@ def decode_jwt(token):
         # This covers e.g. bad characters in the signature or non-JSON-dict payload
         raise_from(InvalidToken(token, "Could not decode token - maybe it's truncated "
                                        "or corrupted? ({})".format(err)), err)
+    except jwt.exceptions.ExpiredSignatureError as err:
+        claims = jwt.decode(token, verify=False)
```
bmerry: I haven't looked through the sequence of operations in pyjwt; is there any risk that this will also raise an exception e.g. if the signature has expired AND there is something else wrong, and turning off verification exposes the second problem. I'm guessing it should be okay.

katdal/chunkstore_s3.py
ludwigschwardt: Thanks, I like.

katdal/chunkstore_s3.py
ludwigschwardt: I did check that. At the JWT level `verify=False` lets `pyjwt` load the JWS segments (while not verifying the signature) and then extract the JSON payload without any further validation of claims.

My only concern is whether the older `verify` parameter is deprecated in favour of `options`. It seems to be discouraged at the JWS level but still accepted at the JWT level. And it is very convenient for this use case.

PR #278, opened 2020-01-17
 Add flux calibration to applycal

ludwigschwardt:
This applies a flux scale to those gains of the "G" calibration product that are associated with gain calibrator targets with a valid flux density in the `gaincal_fluxes` dict. More specifically, the gains are multiplied by the square root of the flux, which turns the default flux model value of 1.0 into the appropriate flux for the gain calibrator. The fluxes are obtained from the `measured_flux` telstate key produced by the cal pipeline but it can also be overridden via `katdal.open` or disabled by specifying an empty dict.

This is equivalent to the final step of the AIPS GETJY and CASA fluxscale tasks.

This addresses JIRA tickets SPR1-67 and SPAZA-27.

### DIFF
--- a/katdal/applycal.py
+++ b/katdal/applycal.py
@@ -246,7 +246,39 @@ def calc_gain_correction(sensor, index, targets=None):
     return np.reciprocal(smooth_gains)
 
 
-def add_applycal_sensors(cache, attrs, data_freqs, cal_stream, cal_substreams=None):
+def calibrate_flux(sensor, targets, gaincal_fluxes):
+    """Apply flux scale to calibrator gains (aka flux calibration).
+
+    Given the gain calibration solution `sensor`, this identifies the target
+    associated with each set of solutions by looking up the gain events in the
+    `targets` sensor, and then scales the gains by the square root of the
+    relevant flux if a valid match is found in the `gaincal_fluxes` dict. This
+    is equivalent to the final step of the AIPS GETJY and CASA fluxscale tasks.
+    """
+    # If no calibration info is available, do nothing
+    if not gaincal_fluxes:
+        return sensor
+    calibrated_gains = []
+    for segment, gains in sensor.segments():
+        # Ignore "invalid gain" placeholder (typically the initial value)
+        if gains is INVALID_GAIN:
+            calibrated_gains.append(ComparableArrayWrapper(gains))
+            continue
+        # Find the target at the time of the gain solution (i.e. gain calibrator)
+        target = targets[segment.start]
+        for name in [target.name] + target.aliases:
+            flux = gaincal_fluxes.get(name, np.nan)
+            # Scale the gains if a valid flux density was found for this target
+            if flux > 0.0:
+                calibrated_gains.append(ComparableArrayWrapper(gains * np.sqrt(flux)))
+                break
+        else:
+            calibrated_gains.append(ComparableArrayWrapper(gains))
+    return CategoricalData(calibrated_gains, sensor.events)
+
+
+def add_applycal_sensors(cache, attrs, data_freqs, cal_stream, cal_substreams=None,
+                         gaincal_fluxes=None):
     """Register virtual sensors for one calibration stream.
 
     This operates on a single calibration stream called `cal_stream` (possibly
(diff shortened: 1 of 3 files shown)


katdal/applycal.py
```diff
@@ -302,6 +337,8 @@ def add_applycal_sensors(cache, attrs, data_freqs, cal_stream, cal_substreams=No
                        "spectral attributes", cal_stream)
         return
     targets = cache.get('Observation/target')
+    if gaincal_fluxes is None:
+        gaincal_fluxes = attrs.get('measured_flux', {})
```
mauch: You might want to merge the user provided dict with the cal dict- so that someone can change the flux of just one in a long list of calibrators without deleting all of the others.

katdal/applycal.py
```diff
@@ -342,6 +379,7 @@ def calc_correction_per_input(cache, name, inp, product_type):
             correction_sensor = calc_bandpass_correction(product_sensor, index,
                                                          data_freqs, cal_freqs)
         elif product_type == 'G':
+            product_sensor = calibrate_flux(product_sensor, targets, gaincal_fluxes)
```
mauch: Is this the optimal place to scale the gains?

katdal/test/test_applycal.py
```diff
@@ -456,11 +472,38 @@ def test_indirect_cal_product(self):
...
+        self.sensor = create_categorical_sensor(GAIN_EVENTS, gains, INVALID_GAIN)
+        calibrated_gains = create_product(partial(create_gain, fluxes=True))
+        self.calibrated_sensor = create_categorical_sensor(
+            GAIN_EVENTS, calibrated_gains, INVALID_GAIN)
+
+    def test_basic(self):
+        calibrated_sensor = calibrate_flux(self.sensor, TARGET_SENSOR, FLUXES)
+        assert_categorical_data_equal(calibrated_sensor, self.calibrated_sensor)
+
+    def test_missing_fluxes(self):
```
mauch: Perhaps some further tests here testing user supplied gains. E.g. what if there is a 'measured_flux' target that isn't in the gain calibrator list?

katdal/visdatav4.py
```diff
@@ -451,7 +456,8 @@ def _register_standard_cal_streams(self):
             for prefix in reversed(attrs.prefixes):
                 l2_attrs = l2_attrs.view(prefix + l2_streams[0])
             l2_freqs = add_applycal_sensors(self.sensor, l2_attrs, freqs,
-                                            cal_stream='l2', cal_substreams=l2_streams)
+                                            cal_stream='l2', cal_substreams=l2_streams,
```
mauch: Adding `gaincal_fluxes` to l2 gains is a worrying as you should not be flux scaling the self-calibrated gains.

katdal/visdatav4.py
```diff
@@ -142,14 +142,18 @@ class VisibilityDataV4(DataSet):
         while the keyword 'all' means all available products will be applied.
         *NB* In future the default will probably change to 'all'.
         *NB* This is still very much an experimental feature...
+    gaincal_fluxes : dict mapping string to float, optional
```
mauch: How about `flux_gaincals`? Which maybe implies that its a single flux per gaincal that the user can provide.

But I'm happy with this either way:)

mauch:
One other comment- I see you haven't included reporting the measured flux and its error of the gaincals in this PR. Is that for a future one?

katdal/applycal.py
ludwigschwardt: Ideally one could move it earlier, to when the G product is extracted from telstate, like a custom virtual sensor for `'Calibration/Products/l1/G'` instead of the usual `indirect_cal_product` function, but I consider it one hoop too many unless the performance benefit is clear.

katdal/test/test_applycal.py
ludwigschwardt: Easy to add.

katdal/visdatav4.py
ludwigschwardt: Removed this for clarity (would only have happened if l2 had `G` and not `GPHASE`).

ludwigschwardt:
The reporting is for a future PR, yes, since it probably needs some pipeline support (and I should probably report applycal in general).

katdal/visdatav4.py
ludwigschwardt: Or `gaincal_flux`? Looks more like `measured_flux` and the word `gaincal` should already make you think "which gaincal"?

katdal/applycal.py
ludwigschwardt: A side effect of this is that `gaincal_fluxes=None` now disables flux calibration, while the default becomes `gaincal_fluxes={}` instead.

PR #279, opened 2020-01-22
Fix flux calibration

ludwigschwardt:
- Look for `measured_flux` in the right place by adding relative telstate views to access `cb_cal` namespace
- Divide gains by flux instead of multiplying it

There is also a slight rework of `_register_standard_cal_streams` to use the new relative views.


ludwigschwardt:
@bmerry, you are welcome to have a look at this and #278 now that you are back. No obligation though :-)

PR #280, opened 2020-01-31
Fix mvf_rechunk script

bmerry:
Closes SPR1-402.


scripts/mvf_rechunk.py
```diff
@@ -75,7 +74,7 @@ def get_chunk_store(source, telstate, array):
 
 
 def comma_list(value):
-    return ','.split(value)
```
ludwigschwardt: :-)

PR #282, opened 2020-02-13
Mark Python 2 support as deprecated.

bmerry:


bmerry:
@ludwigschwardt @sratcliffe what do you think of putting in some form of deprecation warning for people using katdal on Python 2, to give users some warning before we pull the rug from under them?

PR #285, opened 2020-02-19
Avoid using has_array in ChunkStoreVisFlagsWeights

bmerry:
Instead, just try to load things, and if any of them fail, apply flags.
This has a few advantages:

1. chunkstore_s3 is non-deterministic, because it will treat a 403 or
50x error as a missing chunk even if the bucket listing included the
chunk. This meant that even when loading vis, flags, weights jointly,
the vis/weights could be missing and the flags wouldn't reflect this (or
if the flags got a 50x, it would return zero flags). With this change,
joint loading should always return consistent results.

2. Getting the list of all chunks could add a lot of time to
`katdal.open`, which now disappears.

A synthetic benchmark (with no applycal) shows that loading drops from
about 3.3GB/s to about 3GB/s, but that's still faster than the 25Gb/s
network can deliver the data.

This is part 1, which just removes the calls to has_array. Part 2 will
remove all the has_array interfaces and the dependence on defusedxml.

### DIFF
--- a/katdal/chunkstore.py
+++ b/katdal/chunkstore.py
@@ -237,13 +237,13 @@ class ChunkStore(object):
         """
         raise NotImplementedError
 
-    def get_chunk_or_default(self, array_name, slices, dtype, fill_value=0):
+    def get_chunk_or_default(self, array_name, slices, dtype, default_value=0):
         """Get chunk from the store but return default value if it is missing."""
         try:
             return self.get_chunk(array_name, slices, dtype)
         except ChunkNotFound:
             chunk_name, shape = self.chunk_metadata(array_name, slices)
-            return np.full(shape, fill_value, dtype)
+            return np.full(shape, default_value, dtype)
 
     def get_chunk_or_none(self, array_name, slices, dtype):
         """Get chunk from the store but return ``None`` if it is missing."""
@@ -495,7 +495,7 @@ class ChunkStore(object):
             get_func = self.get_chunk
         else:
             get_func = self.get_chunk_or_default
-            kwargs['fill_value'] = errors
+            kwargs['default_value'] = errors
         getter = functools.partial(get_func, **kwargs)
         if offset:
             getter = _add_offset_to_slices(getter, offset)
--- a/katdal/chunkstore.py
+++ b/katdal/chunkstore.py
@@ -237,20 +237,17 @@ class ChunkStore(object):
         """
         raise NotImplementedError
 
-    def get_chunk_or_default(self, array_name, slices, dtype, fill_value=0):
-        """Get chunk from the store but return default value if it is missing."""
+    def get_chunk_or_default(self, array_name, slices, dtype, default_value=0):
+        """Get chunk from the store but return default value if it is missing.
+
+        If a default value is returned, it is guaranteed to have all-zero
+        strides. This can be used to distinguish it from a real chunk.
+        """
         try:
             return self.get_chunk(array_name, slices, dtype)
         except ChunkNotFound:
             chunk_name, shape = self.chunk_metadata(array_name, slices)
-            return np.full(shape, fill_value, dtype)
-
-    def get_chunk_or_none(self, array_name, slices, dtype):
-        """Get chunk from the store but return ``None`` if it is missing."""
-        try:
-            return self.get_chunk(array_name, slices, dtype)
-        except ChunkNotFound:
-            return None
+            return np.broadcast_to(np.full((), default_value, dtype), shape)
 
     def create_array(self, array_name):
         """Create a new array if it does not already exist.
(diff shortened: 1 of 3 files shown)


katdal/datasources.py
```diff
@@ -228,34 +240,78 @@ def __init__(self, store, chunk_info, corrprods):
...
+            dsk = {
+                (new_name,) + index: (
+                    _default_zero,
+                    (array.name,) + index,
+                    shape,
+                    array.dtype
+                ) for index, shape in zip(indices, itertools.product(*array.chunks))
+            }
+            dsk = HighLevelGraph.from_collections(new_name, dsk, dependencies=[array])
+            darray[array_name] = da.Array(dsk, new_name,
```
ludwigschwardt: So the "filled" dask arrays replace the ones in `darray` but the new `flags` array doesn't?

bmerry:
> So this solves the 50x problem too?

What 50x problem?

bmerry:
> So this solves the 50x problem too?

Oh right, that 50x problem (I parsed it as "50 times" which made no sense). Yes, it should fix it (as long as one reads jointly). Since the benchmark was occasionally hitting 2.5GB/s over the 25Gb/s network I'm assuming that I did things right to ensure that it doesn't read visibility chunks twice.

ludwigschwardt:
Excellent!

katdal/chunkstore.py
```diff
@@ -237,13 +237,20 @@ def get_chunk(self, array_name, slices, dtype):
         """
         raise NotImplementedError
 
-    def get_chunk_or_zeros(self, array_name, slices, dtype):
-        """Get chunk from the store but return zeros if it is missing."""
+    def get_chunk_or_default(self, array_name, slices, dtype, fill_value=0):
```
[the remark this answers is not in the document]
bmerry: I don't see `default_value` anywhere in katdal. Or are you referring to the `default` in the name of the method?

katdal/datasources.py
bmerry: Good question. It doesn't actually matter because it doesn't get used again, but I've updated the code to update `darray['flags']` with the fixed-up flags.

katdal/test/test_chunkstore.py
```diff
@@ -249,15 +252,15 @@ def test_dask_array_put_parts_get_whole(self):
...
             array_name, dask_array, offset = self.make_dask_array('big_y2')
             pull = self.store.get_dask_array(array_name, dask_array.chunks,
-                                             dask_array.dtype, offset)
+                                             dask_array.dtype, offset, errors=17)
             array_retrieved = pull.compute()
             assert_equal(array_retrieved.shape, dask_array.shape)
             assert_equal(array_retrieved.dtype, dask_array.dtype)
-            assert_array_equal(array_retrieved[np.s_[3:8, 30:60, 0:2]], 0,
+            assert_array_equal(array_retrieved[np.s_[3:8, 30:60, 0:2]], 17,
                                "Missing chunk in {} not replaced by zeros"
```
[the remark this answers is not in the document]
bmerry: Done.

katdal/chunkstore.py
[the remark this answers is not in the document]
ludwigschwardt: Yes, the `default` in the method name, in the method docstring, in `get_dask_array`'s docstring, in the `_default_zero` function name, etc.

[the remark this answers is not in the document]
bmerry: Fair enough, done. And I've made a cunning change that removes the need for `get_chunk_or_none` and the subsequent remapping of the None value.

bmerry:
It looks like the clever scheme in my last commit was a little too clever: it works great with dask's threaded scheduler, but if one uses a different scheduler (e.g. the multiprocess scheduler) then the arrays get pickled and unpickled, which doesn't preserve the stride property I was depending on. So I'm going to roll back that last change.

katdal/datasources.py
```diff
@@ -291,27 +285,6 @@ def __init__(self, store, chunk_info, corrprods):
                          dtype=darray['flags'].dtype)
         darray['flags'] = flags
```
ludwigschwardt: I guess this is not necessary anymore now that the "filled-" dask arrays are gone.

katdal/chunkstore.py
```diff
@@ -237,20 +237,17 @@ def get_chunk(self, array_name, slices, dtype):
...
-        """Get chunk from the store but return ``None`` if it is missing."""
+        If a default value is returned, it is guaranteed to have all-zero
+        strides. This can be used to distinguish it from a real chunk.
+        """
         try:
             return self.get_chunk(array_name, slices, dtype)
         except ChunkNotFound:
-            return None
+            chunk_name, shape = self.chunk_metadata(array_name, slices)
+            return np.broadcast_to(np.full((), default_value, dtype), shape)
```
ludwigschwardt: Very cunning indeed!

katdal/datasources.py
ludwigschwardt: Scratch this...

katdal/chunkstore.py
ludwigschwardt: :-(

ludwigschwardt:
Sorry, only saw the revert after my comments... All good to go then!

PR #287, opened 2020-02-21
Only apply culling optimisation when selecting a small slice

bmerry:
dask_getitem is used for both first- and second-stage indexing in
DaskLazyIndexer. When first-stage indexing keeps all (no `select`) or
most (e.g. `scans=track`) of the data, the culling is not particularly
useful, very slow, and might be harmful as it throws away the
highlevelgraph structure. So only apply it when the number of blocks is
reduced by at least 50%.

Also ensure that the replacement graph is a HighLevelGraph - dask
doesn't currently do it automatically, unlike the da.Array constructor.

### DIFF
--- a/katdal/lazy_indexer.py
+++ b/katdal/lazy_indexer.py
@@ -25,6 +25,7 @@ from functools import reduce, partial
 
 import numpy as np
 import dask.array as da
+import dask.highlevelgraph
 import dask.optimization
 
 # TODO support advanced integer indexing with non-strictly increasing indices (i.e. out-of-order and duplicates)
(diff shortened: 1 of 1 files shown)


bmerry:
This can knock over a minute off the time to start reading a large dataset (it doesn't come off `katdal.open`, but rather than first access to the data because we construct the datasets lazily).

bmerry:
> Great! Is 50% roughly optimal?

It's a complete thumb-suck. But I figure most cases will either have a fraction close to 0 (e.g. spectral imaging, or selecting a single track) or close to 1 (e.g. scans='track').

PR #290, opened 2020-02-24
Optimise set(targets) -> targets.unique_values

bmerry:
CategoricalData already has a convenient way to get all the unique
values :-)


bmerry:
Done.

PR #292, opened 2020-03-09
Refactor SensorData

bmerry:
SensorData is now explicitly a lazy abstract interface to access sensor values: one calls `get` to retrieve an instance of SensorValues which holds data in memory (feel free to suggest better names). This started as an approach to avoid caching in TelstateSensorData, but turned out to be a long string to pull on. It does clean up some of the complexities around dtype (where SensorData might or might not have a known dtype, depending on whether the lazy loading has occurred yet). Now, SensorData has no dtype, while SensorValues always has a known dtype.

There may be some minor changes in behaviour because previously RecordSensorData was used as a way to take some already-in-memory timestamps and values and put them back into the SensorData interface, but RecordSensorData also had the side effect of calling `to_str` on the values (which means that, for example, `remove_duplicates_and_invalid_values` could alter the types of the data). That side effect is gone, so it's necessary to be more careful to call `to_str` where appropriate.

There are still some issues in concatenated data sets, but I don't think anything that wasn't already broken before.


katdal/concatdata.py
```diff
@@ -217,20 +217,16 @@ def common_dtype(sensor_data_sequence):
...
         The promoted dtype of the sequence, or None if dtype is unknown
 
     """
-    # The sd object will always have a dtype attribute, unless we are extracting
-    # a categorical sensor with selection resulting in a list of sensor values.
-    # For that reason only, use the infer_dtype function just to be safe.
-    dtypes = [infer_dtype(sd) for sd in sensor_data_sequence]
+    dtypes = [sd.dtype for sd in sensor_data_sequence]
     # Ignore all unavailable dtypes for now
     dtypes = [dt for dt in dtypes if dt is not None]
```
ludwigschwardt: Is it still possible to have a dtype of `None`?

katdal/concatdata.py
```diff
@@ -354,34 +334,24 @@ def _set_keep(self, keep=None):
...
                 cache._set_keep(keep[self._segments[n]:self._segments[n + 1]])
 
-    def _get(self, name, select, extract, **kwargs):
+    def _get(self, name, **kwargs):
         """Extract sensor data from multiple caches (see :meth:`get` for docs).
 
-        This extracts a sequence of sensor data objects, one from each cache,
+        This extracts a sequence of sensor data objects, one from each cache.
+        For caches which do not contain the sensor it returns `None`.
         in the process filling in any missing data if the relevant dtype
```
ludwigschwardt: There is some dangling text here.

katdal/concatdata.py
```diff
@@ -312,24 +310,6 @@ def __init__(self, caches, keep=None):
...
-                    # name due to aliasing, and concatenation cares about it.
-                    # Fully extracted ndarrays don't have .name, though...
-                    raw_name = getattr(sensor_data, 'name', name)
-                    dtype = sensor_data.dtype
-                    if dtype is None:
-                        # Instantiate base class as placeholder if type unknown
-                        cache[name] = SensorData(raw_name, dtype)
-                    else:
-                        cache[name] = dummy_sensor_data(raw_name, dtype=dtype)
-        # Pad out virtual sensors with default functions (nans)
```
ludwigschwardt: This virtual sensor padding seems to be gone now (probably for the best). It looks like the last virtual sensor lambda will win the day (if they differ between the underlying datasets).

katdal/test/test_concatdata.py
```diff
@@ -0,0 +1,129 @@
...
+
+    def test_no_extract_missing(self):
+        data = self.cache.get('float_missing', extract=False)
+        values = data.get()
+        np.testing.assert_array_equal(values.timestamp, [995.0])
+        np.testing.assert_array_equal(values.value, [3.0])
+
+    def test_missing_sensor(self):
+        with assert_raises(KeyError):
+            self.cache['sir_not_appearing_in_this_cache']
```
ludwigschwardt: 😄

katdal/test/test_concatdata.py
```diff
@@ -0,0 +1,129 @@
...
+    def test_partially_extract(self):
+        self.cache1['foo']
+        data = self.cache.get('foo', extract=False)
+        np.testing.assert_array_equal(data, self.cache.get('foo', extract=True))
+
+    def test_setitem_categorical(self):
+        data = CategoricalData(['x', 'y', 'x'], [0, 2, 4, 8])
+        self.cache['dog'] = data
+        ans = self.cache.get('dog')
+        assert_equal(data.unique_values, ans.unique_values)
```
ludwigschwardt: Why not `assert_equal(data, ans)`?

katdal/concatdata.py
bmerry: Fixed.

katdal/sensordata.py
```diff
@@ -36,31 +36,52 @@
...
-    from scikits.fitting import PiecewisePolynomial1DFit
-except ImportError:
-    PiecewisePolynomial1DFit = None
 
 # -------------------------------------------------------------------------------------------------
 # -- CLASS :  SensorData
 # -------------------------------------------------------------------------------------------------
 
 
+class SensorValues(object):
```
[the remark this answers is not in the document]
bmerry: I've renamed SensorData → SensorGetter, SensorValues → SensorData, dummy_sensor_data → dummy_sensor_getter.

katdal/sensordata.py
```diff
@@ -36,31 +36,52 @@
...
+    """Raw (uninterpolated) sensor values.
+
+    This is a simple tuple that holds timestamps, values, and optionally
```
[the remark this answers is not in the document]
bmerry: Fair enough, changed to "struct".

katdal/sensordata.py
```diff
@@ -353,114 +360,35 @@ class TelstateSensorData(SensorData):
...
-        with self._lock:
-            if not self._times:
-                value_times = self._telstate.get_range(self.name, st=0)
-                self._values = [v for v, t in value_times]
-                self.dtype = infer_dtype(self._values)
-                if self.dtype == np.object:
-                    self._values = [ComparableArrayWrapper(v) for v in self._values]
-                self._times = [t for v, t in value_times]
+    def get(self):
+        values, times = zip(*self._telstate.get_range(self.name, st=0))
```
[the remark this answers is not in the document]
bmerry: zip is really transpose :-)

katdal/sensordata.py
```diff
@@ -582,23 +508,20 @@ def remove_duplicates_and_invalid_values(sensor):
 
     Parameters
     ----------
-    sensor : :class:`SensorData` object, length *N*
+    sensor : :class:`SensorValues` object, length *N*
         Raw sensor dataset, which acts like a record array with fields
```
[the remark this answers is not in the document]
bmerry: Shortened (interested readers can read the docstring for SensorData).

katdal/sensordata.py
```diff
@@ -582,23 +508,20 @@ def remove_duplicates_and_invalid_values(sensor):
...
         'timestamp', 'value' and optionally 'status'
 
     Returns
     -------
-    clean_sensor : :class:`RecordSensorData` object, length *M*
+    clean_sensor : :class:`SensorValues` object, length *M*
         Sensor data with duplicate timestamps and invalid values removed
         (*M* <= *N*), and only 'timestamp' and 'value' fields left
```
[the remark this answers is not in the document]
bmerry: Done.

katdal/sensordata.py
```diff
@@ -582,23 +508,20 @@ def remove_duplicates_and_invalid_values(sensor):
...
 
     """
-    x = np.atleast_1d(sensor['timestamp'])
-    y = np.atleast_1d(sensor['value'])
-    try:
-        z = np.atleast_1d(sensor['status'])
-    except ValueError:
-        z = None
+    x = sensor.timestamp
```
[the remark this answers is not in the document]
bmerry: The new SensorData (the new version) should always be constructed with 1D arrays, and I think it's the responsibility of the code calling the constructor to provide such.

katdal/sensordata.py
```diff
@@ -778,6 +699,58 @@ def _set_keep(self, keep=None):
...
+        # (float data are non-categorical, by default)
+        categ = props.get('categorical', not np.issubdtype(sensor_data.value.dtype, np.floating))
+        props['categorical'] = categ
+        if categ:
+            sensor_data = sensor_to_categorical(sensor_data.timestamp, sensor_data.value,
+                                                timestamps, dump_period, **props)
+        else:
+            # Interpolate numerical data onto data timestamps
+            sensor_timestamps = sensor_data.timestamp
+            # Warn if sensor data will be extrapolated to start or end
```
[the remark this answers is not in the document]
bmerry: Since we don't have a better way yet, how about I just remove the part " - extrapolation may lead to ridiculous values"?

katdal/test/test_concatdata.py
bmerry: That gives an array of booleans (one per dump).

katdal/sensordata.py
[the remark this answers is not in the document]
ludwigschwardt: Good point.

katdal/sensordata.py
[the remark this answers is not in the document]
ludwigschwardt: Great!

katdal/test/test_concatdata.py
ludwigschwardt: Ah, yes.

katdal/applycal.py
```diff
@@ -327,7 +327,7 @@ def indirect_cal_product_raw(cache, name, product_type):
             ordered = timestamps.argsort()
             timestamps = timestamps[ordered]
             values = values[ordered]
-            return SimpleSensorData(indirect_cal_product_name(name, product_type),
+            return SimpleSensorGetter(indirect_cal_product_name(name, product_type),
                                     timestamps, values)
```
ludwigschwardt: It looks like the indentation is misaligned here.

katdal/applycal.py
```diff
@@ -379,7 +379,7 @@ def indirect_cal_product(cache, name, product_type):
         if not timestamps:
             raise KeyError("No cal product '{}' parts found (expected {})"
                            .format(name, n_parts))
-        return SimpleSensorData(indirect_cal_product_name(name, product_type),
+        return SimpleSensorGetter(indirect_cal_product_name(name, product_type),
                                 np.array(timestamps), np.array(values))
```
ludwigschwardt: It looks like the indentation is misaligned here.

katdal/sensordata.py
```diff
@@ -73,7 +73,7 @@ def __bool__(self):
         return len(self.timestamp) > 0
 
 
-class SensorData(object):
+class SensorGetter(object):
     """Raw (uninterpolated) sensor data placeholder.
 
     This is basically a placeholder for uninterpolated sensor data resembling
```
ludwigschwardt: This sentence now sounds out of place. Maybe remove it and change the next sentence to:

```
This is an abstract lazy interface that provides a :class:`SensorData` object on request but does not store values itself.
```

katdal/sensordata.py
```diff
@@ -111,22 +111,22 @@ def __repr__(self):
                (self.__class__.__name__, self.name, id(self))
 
 
-class SimpleSensorData(SensorData):
+class SimpleSensorGetter(SensorGetter):
     """Raw sensor data held in memory.
```
ludwigschwardt: While the abstract `SensorData` warns against caching sensor values, that's exactly what `SimpleSensorGetter` does...
And it should 😄

katdal/sensordata.py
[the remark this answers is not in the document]
ludwigschwardt: OK, happy with removing the part.

katdal/sensordata.py
bmerry: I wouldn't call it "caching", since that implies a copy of data that can be retrieved from elsewhere. It holds the master copy in memory.

katdal/sensordata.py
bmerry: Good point. I've taken that and welded it in to the rest of the docstring. See what you think now.

katdal/applycal.py
bmerry: Thanks, fixed (and if you had a `.flake8` then Jenkins would have shouted at me - it looks like katdal is pretty close to being flake8-clean so shouldn't be much work).

katdal/applycal.py
bmerry: Fixed.

katdal/sensordata.py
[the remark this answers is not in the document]
bmerry: Done.

bmerry:
Thanks, that was useful feedback.

katdal/sensordata.py
ludwigschwardt: Sure, I thought about that too, but the `SensorGetter` docstring also says "... but does not store values itself." :-)
I don't think there is anything to change here. I added the comment to the PR mostly for my future self...

PR #295, opened 2020-03-13
Release v0.15

ludwigschwardt:
This updates the changelog, which will get merged back into master. The final commit will be tagged as v0.15 and released.

### DIFF
--- a/NEWS.rst
+++ b/NEWS.rst
@@ -4,9 +4,11 @@ History
 0.15 (2020-03-13)
 -----------------
 * Improve S3 chunk store: check tokens, improve timeouts and retries (#272 - #277)
+* Retry truncated reads and 50x errors due to S3 server overload (#274)
 * Apply flux calibration if available (#278, #279)
 * Improve mvf_rechunk and mvf_read_benchmark scripts (#280, #281, #284)
 * Fix selection by target description (#271)
+* Mark Python 2 support as deprecated (#282)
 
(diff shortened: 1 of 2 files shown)


NEWS.rst
```diff
@@ -1,6 +1,13 @@
 History
 =======
 
+0.15 (2020-03-13)
+-----------------
+* Improve S3 chunk store: check tokens, improve timeouts and retries (#272 - #277)
+* Apply flux calibration if available (#278, #279)
+* Improve mvf_rechunk and mvf_read_benchmark scripts (#280, #281, #284)
+* Fix selection by target description (#271)
```
bmerry: Maybe add a note that Python 2 support is deprecated?

NEWS.rst
```diff
@@ -1,6 +1,13 @@
 History
 =======
 
+0.15 (2020-03-13)
+-----------------
+* Improve S3 chunk store: check tokens, improve timeouts and retries (#272 - #277)
```
bmerry: So you want to say anything specific about truncated reads?

NEWS.rst
ludwigschwardt: Ah yes, missed that since I only looked at PR merge commits...

NEWS.rst
ludwigschwardt: Good idea.

PR #297, opened 2020-04-09
Allow DecodeError to propagate in v3

bmerry:
A likely cause of such errors is pickled encoding, which triggers an
error unless the user has set the appropriate environment variable to
allow it (unsafe).

This may cause some files that have invalid encoding (e.g. from the
early days where telescope state just used `repr`) to fail, but that's
probably better than just ignoring data in the file and using a default
instead. If users complain we can then fix katdal to handle such files.

The approach of treating decoding errors the same as missing data seems
to date back to 645a5e2, where it was applied just to the band (and
would at least lead to a warning that the receiver band couldn't be
determined).

Closes SPR1-446.

### DIFF
--- a/katdal/h5datav3.py
+++ b/katdal/h5datav3.py
@@ -609,7 +609,7 @@ class H5DataV3(DataSet):
             # needs to stay binary
             value = self.file['TelescopeState'].attrs[key]
             return telstate_decode(value, no_decode)
-        except (KeyError, katsdptelstate.DecodeError):
+        except KeyError:
             # In some cases the value is placed in a sensor instead. Return
             # the most recent value.
             try:


bmerry:
@ludwigschwardt let me know what you think. I'm not 100% convinced that this won't create new problems, but any new problems it does throw up will be old bugs that have been silently ignored.

ludwigschwardt:
Do you have an example where this tripped you up?

bmerry:
> Do you have an example of where this tripped you up?

Yes, see SPR1-446.

ludwigschwardt:
How about forcing the user to enable pickle support in v2 / v3? There should be no msgpack in those files...

bmerry:
> How about forcing the user to enable pickle support in v2 / v3? There should be no msgpack in those files...

Interesting idea, although if they're old enough (certainly v2, probably also early v3) I don't think there will be any pickles either, in which case there shouldn't be any need to force anything.

But even if we do that, shouldn't decoding errors be thrown rather than silently treated as if the attribute wasn't present?

ludwigschwardt:
What about the old repr-based telstates in v3? Would they still work?

bmerry:
> What about the old repr-based telstates in v3? Would they still work? I'll see if I can test it on an example...

Thanks, let me know. I couldn't see where that was handled, but if they raise an error with this change then it means they were silently broken (telstate values being ignored) before. Maybe we just didn't notice because the fallback heuristics to guess missing values was good enough.

ludwigschwardt:
I had a look at this again... The new changes break old repr-based files like 1454045147.h5 (anything before 2016-11-30).

I agree with the basic premise of the PR but would like to see old files working as far as possible. You've changed it from a cryptic crash on some files fixable by enabling pickles to a cryptic crash on the rest of the files that cannot be fixed without changing the code.

I would add a warning that pickles need to be enabled to solve potential decode errors in v3 files. And then I would consider making TelescopeModel the default above TelescopeState in `_get_l0_attr`, because this is the fallback for repr-ed attrs; i.e. change the order of
```Python
            self._raise_if_attr_found(self.stream_name, key)
            if sdp_group is not None:
                self._raise_if_attr_found('l0', key, h5_group=sdp_group)
```
The repr-ed sensors go in via `katdal.sensordata._h5_telstate_unpack`, which gloriously ignores decode errors :-)

ludwigschwardt:
This is getting too tricky to fix for now. If users are interested in old files, they should always enable pickles and then this won't be an issue.

PR #298, opened 2020-04-09
Cope with rx_serial being a string

bmerry:
It was being turned into a string via %d, but in some old files (e.g.
1462454843.h5) it ended up encoded as a string already, causing this to
crash.


bmerry:
> Looks safe. Where do you find those old files these days for testing?

You gave me a collection of old files that you use to test.

PR #299, opened 2020-04-09
Explicitly pass file mode to h5py.File

bmerry:
For dummy_dataset (the real file already does it). This avoids a
deprecation warning from h5py that the default mode will change to 'r'.


bmerry:
> Why the sudden interest in v2 / v3? :-)

It's low-hanging fruit - bugs that are small but didn't seem important enough to fix at the time. So I'm paying off some technical debt, because even with COVID-19 I don't think we can get the central bank to write it off :-)

PR #300, opened 2020-04-09
Use MutableMapping for SensorCache

bmerry:
This provides more predictable/controllable behaviour compared to
inheriting directly from dict. I also made some fixes to
ConcatenatedSensorCache regarding iteration and length (previously it
implemented iterkeys, which didn't work at all in Python 3 and also made
the assumption that the underlying caches had exactly the same keys).

Querying a sensor cache can cause new keys to appear in the underlying
dict (as virtuals are expanded), so it was necessary to make some
choices). I decided that __contains__ will only consider instantiated
keys, not virtuals, making it consistent with `keys()` and `len` (and
the previous implementation of SensorCache), but inconsistent with
`__getitem__`.

One user-visible change in SensorCache is that `items()` and `.values()`
will now return extracted data, because MutableMapping implements them
in terms of `__getitem__`.

Closes SPR1-445.

### DIFF
--- a/katdal/concatdata.py
+++ b/katdal/concatdata.py
@@ -451,7 +451,7 @@ class ConcatenatedSensorCache(SensorCache):
             except KeyError:
                 pass
         if not found:
-            raise KeyError
+            raise KeyError(name)
 
     def __contains__(self, name):
         return any(name in cache for cache in self.caches)
--- a/katdal/concatdata.py
+++ b/katdal/concatdata.py
@@ -301,11 +301,9 @@ class ConcatenatedSensorCache(SensorCache):
 
     def __init__(self, caches, keep=None):
         self.caches = caches
-        # Collect all actual and virtual sensors in caches as well as properties
-        # The main point is to discover the name and dtype of each known sensor
-        actual, virtual, self.props = {}, {}, {}
+        # Collect all virtual sensors in caches as well as properties.
+        virtual, self.props = {}, {}
         for cache in caches:
-            actual.update(cache.items())
             virtual.update(cache.virtual)
             self.props.update(cache.props)
         self.virtual = virtual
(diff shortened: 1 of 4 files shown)


bmerry:
Looks like `future` didn't handle the rename with `collections.abc`, so I'll need to fix that before merging.

katdal/sensordata.py
```diff
@@ -887,3 +892,26 @@ def get_with_fallback(self, sensor_type, names):
...
+    def __iter__(self):
+        return iter(self._raw)
+
+    def __len__(self):
+        return len(self._raw)
+
+    def __contains__(self, key):
+        # __contains__ is implemented by MutableMapping via __getitem__, but
+        # that does unnecessary extraction. This approach is cheaper but only
+        # reflects keys that have been explicitly created or cached.
```
ludwigschwardt, line 915: This should be OK because we are already ignoring all the miriad sensors available via katstore...

katdal/concatdata.py
```diff
@@ -444,10 +442,31 @@ def __setitem__(self, name, data):
...
+    def __delitem__(self, name):
+        found = False
+        for cache in self.caches:
+            try:
+                del cache[name]
+                found = True
+            except KeyError:
+                pass
+        if not found:
+            raise KeyError
```
ludwigschwardt, line 454: Do you need to pass `name` to `KeyError` to match the error message from `dict`?

katdal/concatdata.py
```diff
@@ -444,10 +442,31 @@ def __setitem__(self, name, data):
...
+
+    def __contains__(self, name):
+        return any(name in cache for cache in self.caches)
+
+    def __len__(self):
+        return sum(1 for _ in self)
```
ludwigschwardt, line 460: Nice trick to remember for iterators.

katdal/test/test_sensordata.py
```diff
@@ -75,8 +75,57 @@ def test_numpy_object(self):
...
+
+    def test_extract_float(self):
+        data = self.cache.get('foo', extract=True)
+        np.testing.assert_array_equal(data, [3.0, 3.0, 3.0, 3.0, 3.0, 4.0, 5.0, 6.0, 6.0, 6.0])
+
+    def test_extract_categorical(self):
+        data = self.cache.get('cat', extract=True)
+        H = 'hello'
+        W = 'world'
+        np.testing.assert_array_equal(data[:], [H, H, H, H, H, H, W, W, W, W])
```
ludwigschwardt, line 100: Very nice tests! Inspirational :-D

katdal/concatdata.py
bmerry: I can't claim credit - I found it somewhere on the internet.

katdal/concatdata.py
bmerry: Good idea - done.

PR #301, opened 2020-04-13
Fix calculation of cbf_dump_period

bmerry:
It had a hard-coded assumption that the instrument is called i0, which
is no longer true.

Closes SPR-435.

### DIFF
--- a/katdal/visdatav4.py
+++ b/katdal/visdatav4.py
@@ -181,8 +181,12 @@ class VisibilityDataV4(DataSet):
         self.version = '4.0'
         self.dump_period = attrs['int_time']
         # The CBF dump period is not in the lite RDB version
-        self.cbf_dump_period = attrs.get(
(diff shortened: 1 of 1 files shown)


katdal/visdatav4.py
```diff
@@ -181,8 +181,12 @@ def _before(date):
         self.version = '4.0'
         self.dump_period = attrs['int_time']
         # The CBF dump period is not in the lite RDB version
-        self.cbf_dump_period = attrs.get(
-            'i0_baseline_correlation_products_int_time', None)
+        try:
+            src_stream = attrs['src_streams'][0]
```
ludwigschwardt, line 185: Is the `wide_b_c_p` always stream 0? And what about narrowband instruments?

bmerry: L0 streams always have exactly one element in `src_streams` - there is no difference between wide and narrow.

PR #302, opened 2020-04-17
Increase katstore search window by 600 seconds

ludwigschwardt:
Katstore samples sensors at least once every 10 minutes (not seconds!), so we have to go that far back in our search to ensure that we get values for discrete sensors that change very infrequently (like the band setting). If not, short 5-minute observations like delaycals can miss a sensor value completely.

### DIFF
--- a/katdal/sensordata.py
+++ b/katdal/sensordata.py
@@ -840,9 +840,10 @@ class SensorCache(dict):
                         break
                 else:
                     if self.store:
-                        # Katstore samples typically at least every 10 seconds
-                        start_time = self.timestamps[0] - self.dump_period - 10
-                        end_time = self.timestamps[-1] + self.dump_period + 10
+                        # Katstore samples sensors at least once every 10 minutes
+                        # Go that far back to support sporadic discrete sensors
+                        start_time = self.timestamps[0] - self.dump_period - 600
+                        end_time = self.timestamps[-1] + self.dump_period + 60
                         sensor_data = get_sensor_from_katstore(
                             self.store, name, start_time, end_time)
                     else:


mgouws:
I'm wondering if it would be more sensible to change the window to -300 / +300 instead of -600 / +10?

Delay cals are often run immediately after a subarray is created. In the worst case, if a sensor update arrives 10 minutes after the subarray was created, then the search window might still miss it.

mgouws:
In testing I no longer find any sensor key errors as before. 

However, I am seeing one new issue where eq sensors for some receptors (e.g. 'cbf_1_wide_antenna_channelised_voltage_m000v_eq') return an empty string for the entire observation, without any warnings given.

Several warnings for no usable data found, e.g.
WARNING:katdal.sensordata:No usable data found for sensor 'm000_dig_u_band_adc_overrange' - replaced with dummy data (False)

Also a few warnings that might suggest update rate is longer than 10 minutes on some sensors:
WARNING:katdal.sensordata:Last data point for sensor 'm005_rsc_rxu_lna_v_drain_current' arrives 858.921 seconds before end of data set.

ludwigschwardt:
No, you want the 600 seconds in front, otherwise a later value will be
extrapolated backwards and this might be incorrect for a discrete sensor if
its value actually changed at that time (no way of knowing...).

I’ll have a look at the latest issues uncovered. The CBF sensors probably
have their own set of issues given their torturous birth...

mgouws:
The eq sensor issue only occurs on these short delay cals, so I'm assuming it's an unexpected response to sensor data not being available. For an example, see eq sensors for m002v, m011h, el al. in observation 20200415-0008.

I'm happy to approve the merge. Shall we continue in spite of above issue? It's fairly minor in my view.

ludwigschwardt:
There's some dodgy stuff with the eq sensors. You're one of the few consumers of this sensor: thanks for helping to debug it :-)

For example, I checked 1586961195 cbf_1_wide_antenna_channelised_voltage_m000v_eq and it has an empty string value, followed by the proper array of values at an **earlier** timestamp. And then one or two duplicates of the latters (same value and timestamp).

It's best to check in with CBF to understand the sensor better.

PR #304, opened 2020-05-05
Support indexed telstate keys

bmerry:
Note that this will break on Python 2 because the last version of
katsdptelstate to support Python 2 doesn't support indexed keys.


bmerry:
Draft PR just to show what changes are needed - won't merge as is because it'll break Python 2 support.

setup.py
```diff
@@ -48,14 +48,12 @@
...
           'Programming Language :: Python',
-          'Programming Language :: Python :: 2',
-          'Programming Language :: Python :: 2.7',
           'Programming Language :: Python :: 3',
           'Topic :: Software Development :: Libraries :: Python Modules',
           'Topic :: Scientific/Engineering :: Astronomy'],
       platforms=['OS Independent'],
       keywords='meerkat ska',
-      python_requires='>=2.7,!=3.0,!=3.1,!=3.2',
+      python_requires='>=3.5',
```
ludwigschwardt, line 56: This feels like a separate PR... Or maybe not, given the age of Python 2 :-D

bmerry: The alternative is to add a bunch of code to test whether the installed version of katsdptelstate supports indexed keys so that we can continue to support the old Python 2 version of katsdptelstate.

bmerry:
@ludwigschwardt which direction do you want to take this? Are you happy to drop Python 2 support at this point, or do you want to add a legacy codepath for katsdptelstate's that don't support indexed keys?

bmerry:
@ludwigschwardt I've taken off the draft tags since you sounded ready to drop Python 2 support.

PR #308, opened 2020-06-30
Expose excision aka raw weights

ludwigschwardt:
The signal-to-noise-ratio weights in katdal typically have the value

    w_ij = K_ij / v_ii / v_jj,

i.e. the inverse of the variance of the visibility `v_ij`. To understand excision it is useful to extract `K_ij`, the number of voltage samples (aka accumulations) that make up the specified SDP integration. This is called the "unscaled" weight, before division by the relevant autocorrelations.

The modern MeerKAT system only stores K, leaving it up to katdal to complete the weights by dividing K by the power terms. Older MVFv4 files stored the scaled weights instead - for them we will undo the scaling (i.e. multiply by autocorrelations) to obtain the unscaled weights.

We then add an excision property that provides a float32 value between 0.0 and 1.0 per visibility that indicates the fraction of CBF visibilities that were excised by the SDP ingest pipeline when forming the SDP L0 visibility. It is affected by dataset selection like vis, flags and weights, and also calculated lazily. The excision also fixes common weight compression artefacts.

The first two commits are pure refactoring that split the `datasources` module into two equal parts. The last two commits add the juicy bits.


bmerry:
I'll take a look next week, unless there is urgency to get it done sooner.

ludwigschwardt:
Sure, you're welcome. Just so you know, the core changes in the last two commits only affect ~160 lines, not the ~1000 lines implied at the top (most are file reorganisations).

katdal/datasources.py
```diff
@@ -380,11 +380,10 @@ def __init__(self, telstate, capture_block_id, stream_name,
         if chunk_store is None:
             data = None
         else:
-            if telstate.get('need_weights_power_scale', False):
-                corrprods = telstate['bls_ordering']
-            else:
-                corrprods = None
```
bmerry, line 652: I think it's fine, but just to check: is it possible that bls_ordering might be missing in the old observations that lacked need_weights_power_scale?

katdal/vis_flags_weights.py
```diff
@@ -39,25 +39,32 @@
 class VisFlagsWeights(object):
     """Correlator data in the form of visibilities, flags and weights.
 
+    This container stores the triplet of visibilities, flags and weights
+    as dask arrays to provide lazy / deferred access to the data.
```
bmerry, line 43: Except it actually stores a quadruple.

katdal/vis_flags_weights.py
```diff
@@ -39,25 +39,32 @@
...
+        the number of voltage samples that constitutes each visibility (optional)
     name : string, optional
         Identifier that describes the origin of the data (backend-specific)
 
     """
-    def __init__(self, vis, flags, weights, name='custom'):
+    def __init__(self, vis, flags, weights, unscaled_weights=None, name='custom'):
         if not (vis.shape == flags.shape == weights.shape):
             raise ValueError("Shapes of vis %s, flags %s and weights %s differ"
                              % (vis.shape, flags.shape, weights.shape))
```
bmerry, line 63: Should also check that `unscaled_weights` has the right shape if it is not None.

katdal/vis_flags_weights.py
```diff
@@ -146,7 +153,7 @@ def corrprod_to_autocorr(corrprods):
 
 
 @numba.jit(nopython=True, nogil=True)
-def weight_power_scale(vis, weights, auto_indices, index1, index2, out=None):
+def weight_power_scale(vis, weights, auto_indices, index1, index2, divide=True, out=None):
```
bmerry, line 156: Maybe put `divide` at the end, so that it won't break code that passes `out` positionally (it looks like katsdpcal passes it by keyword so it is probably fine).

katdal/vis_flags_weights.py
```diff
@@ -161,6 +168,8 @@ def weight_power_scale(vis, weights, auto_indices, index1, index2, out=None):
         Chunk of weight data, with the same shape as `vis`.
     auto_indices, index1, index2 : np.ndarray
         Arrays returned by :func:`corrprod_to_autocorr`
+    divide : bool, optional
+        True if weights are divided by autocorrelations; otherwise they are multiplied
```
bmerry, line 172: It's not totally clear whether this refers to the input weights or the output weights. Maybe "should be divided"?

katdal/vis_flags_weights.py
```diff
@@ -192,17 +216,21 @@ class ChunkStoreVisFlagsWeights(VisFlagsWeights):
...
-        Correlation products. If given, the weights for baseline (inp1, inp2)
-        will be divided by the square root of the product of the corresponding
-        autocorrelations vis[inp1,inp1] and vis[inp2,inp2].
+    corrprods : sequence of 2-tuples of input labels, optional
+        Correlation products. If given, the raw weights for baseline (inp1, inp2)
+        will be divided by the product of the corresponding autocorrelations
+        vis[inp1,inp1] and vis[inp2,inp2], and the raw weights will also
+        be exposed as `unscaled_weights`.
+    raw_weights_are_unscaled : bool, optional
+        False if raw weights are already scaled by autocorrelations, in which
```
bmerry, line 225: Describing the meaning when a negated variable name (unscaled) is false is a double-negative and confusing.

katdal/vis_flags_weights.py
```diff
@@ -279,18 +307,17 @@ def __init__(self, store, chunk_info, corrprods):
...
-        VisFlagsWeights.__init__(self, vis, flags, weights, self.vis_prefix)
+            if raw_weights_are_unscaled:
+                weights = _scale_weights(vis, raw_weights, corrprods, divide=True)
+                unscaled_weights = raw_weights
+            else:
+                weights = raw_weights
+                unscaled_weights = _scale_weights(vis, raw_weights, corrprods, divide=False)
+        else:
+            # Don't bother with unscaled weights (it's optional)
+            weights = raw_weights
```
bmerry, line 321: Should it be an error if `raw_weights_are_unscaled` is true here?

katdal/visdatav4.py
```diff
@@ -185,8 +185,12 @@ def _before(date):
             src_stream = attrs['src_streams'][0]
             # XXX: should use telstate.join if attrs is a telstate
             self.cbf_dump_period = attrs[src_stream + '_int_time']
+            cbf_n_accs = attrs[src_stream + '_n_accs']
         except (KeyError, IndexError):
-            self.cbf_dump_period = None
+            self.cbf_dump_period = self.accumulations_per_dump = None
+        else:
+            cbf_dumps_per_sdp_dump = int(round(self.dump_period / self.cbf_dump_period))
```
bmerry, line 192:
```suggestion
            cbf_dumps_per_sdp_dump = round(self.dump_period / self.cbf_dump_period)
```

I discovered recently that in Python 3, `round` already returns an `int`.

katdal/visdatav4.py
```diff
@@ -555,6 +559,20 @@ def _set_keep(self, time_keep=None, freq_keep=None, corrprod_keep=None,
                 flag_transforms.append(lambda flags: da.bitwise_and(select, flags))
             flag_transforms.append(lambda flags: flags.view(np.bool_))
             self._flags = DaskLazyIndexer(self._corrected.flags, stage1, flag_transforms)
+            unscaled_weights = self._corrected.unscaled_weights
+            if unscaled_weights is None or self.accumulations_per_dump is None:
+                self._excision = None
+            else:
+                # The maximum / expected number of CBF dumps per SDP dump
+                cbf_dumps_per_sdp_dump = int(round(self.dump_period / self.cbf_dump_period))
```
bmerry, line 567:
```suggestion
                cbf_dumps_per_sdp_dump = round(self.dump_period / self.cbf_dump_period)
```

katdal/visdatav4.py
```diff
@@ -555,6 +559,20 @@ def _set_keep(self, time_keep=None, freq_keep=None, corrprod_keep=None,
...
+                accs_per_sdp_dump = np.float32(self.accumulations_per_dump)
+                accs_per_cbf_dump = accs_per_sdp_dump / np.float32(cbf_dumps_per_sdp_dump)
+                # Each unscaled weight represents the actual number of accumulations per SDP dump.
+                # Correct most of the weight compression artefacts by forcing each weight to be
+                # an integer multiple of CBF n_accs, and then convert it to an excision fraction.
+                excision_transforms = [lambda w: da.round(w / accs_per_cbf_dump) * accs_per_cbf_dump,
+                                       lambda w: (accs_per_sdp_dump - w) / accs_per_sdp_dump]
```
bmerry, line 574: I'd rather have a single named function here, so that `DaskLazyIndexer.__str__` will print a meaningful name for the transformation.

katdal/datasources.py
ludwigschwardt: I just found the oldest RDB in the archive: 1524571621 captured on 2018-04-24 (holography, of course 😆). That has `bls_ordering` so I think it's OK. For the record, `need_weights_power_scale` appeared on site on 2018-10-17.

katdal/vis_flags_weights.py
ludwigschwardt: 😄 

I was quoting [CASA](https://casacore.github.io/casacore-notes/264.pdf), which calls it the (DATA, WEIGHT, FLAG) triplet. For me, unscaled weights is a derived product of vis and weights, so not really a fourth member.

katdal/vis_flags_weights.py
ludwigschwardt: I have hopefully now improved the docstring.

katdal/vis_flags_weights.py
ludwigschwardt: I was going to leave it as is, using the lack of users and NumPy's API as excuse, but I now see that e.g. `np.multiply` does not have it as the final argument either... So I've moved it to the end.

katdal/vis_flags_weights.py
ludwigschwardt: Done.

ludwigschwardt: Done. Also added a test for this.

katdal/vis_flags_weights.py
ludwigschwardt: Point taken. I've completely reworked this - now called `stored_weights_are_scaled`. Nice and positive.

katdal/vis_flags_weights.py
ludwigschwardt: I like that - will add a ValueError.

katdal/visdatav4.py
ludwigschwardt: Thanks, fixed!

katdal/visdatav4.py
ludwigschwardt: Ditto.

katdal/visdatav4.py
ludwigschwardt: Cool idea - I should do the same for flags when I add the raw flags machinery. I even had a look at `d._excision.dataset.__dask_graph__().layers.keys()` out of curiosity but I guess that's too low-level.

katdal/visdatav4.py
```diff
@@ -564,14 +564,19 @@ def _set_keep(self, time_keep=None, freq_keep=None, corrprod_keep=None,
...
                 # an integer multiple of CBF n_accs, and then convert it to an excision fraction.
-                excision_transforms = [lambda w: da.round(w / accs_per_cbf_dump) * accs_per_cbf_dump,
-                                       lambda w: (accs_per_sdp_dump - w) / accs_per_sdp_dump]
+                def integer_cbf_dumps(w):
+                    return da.round(w / accs_per_cbf_dump) * accs_per_cbf_dump
+
+                def excision_fraction(w):
+                    return (accs_per_sdp_dump - w) / accs_per_sdp_dump
+
+                excision_transforms = [integer_cbf_dumps, excision_fraction]
```
bmerry, line 579: I'd been thinking this would be a single transformation function. What are the performance implications of doing the transformation in two steps?

ludwigschwardt: I'll have to benchmark it... I was careful to keep everything float32 as the first priority. If I look at the dask graph I see 5 operations (truediv, round, mul, subtract, truediv) so the functions are unpacked anyway. Combining the functions might get rid of one mul operation, but maybe not more.

bmerry: I guess it won't really affect the dask graph whether it is one function or two.

PR #310, opened 2020-07-27
Split out separate utility classes for Minio

bmerry:
This will make it practical for other packages to reuse the code.

It is deliberately designed not to use the SkipTest exception, which is
specific to nosetests, so that it can easily be integrated into other
test frameworks (pytest).

The version test has been removed and left to documentation: newer
versions of minio have a different way of querying the version, and it
didn't seem worth trying to support both flavours.

### DIFF
--- a/katdal/test/s3_utils.py
+++ b/katdal/test/s3_utils.py
@@ -1,0 +1,204 @@
+################################################################################
+# Copyright (c) 2017-2020, National Research Foundation (Square Kilometre Array)
+#
+# Licensed under the BSD 3-Clause License (the "License"); you may not use
+# this file except in compliance with the License. You may obtain a copy
+# of the License at
+#
+#   https://opensource.org/licenses/BSD-3-Clause
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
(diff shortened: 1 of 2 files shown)


bmerry:
@ludwigschwardt can you test this on a Mac? I've changed the way the socket is allocated to eliminate a race condition, and I don't know if it will work on Mac (previously we closed the socket before starting Minio, and it could in theory have been reassigned before minio starts).

Also, the test code is full of f-strings. Let me know if you still want to support Python 3.5.

katdal/test/s3_utils.py
```diff
@@ -0,0 +1,200 @@
...
+        env['MINIO_BROWSER'] = 'off'
+        env['MINIO_ACCESS_KEY'] = self.user.access_key
+        env['MINIO_SECRET_KEY'] = self.user.secret_key
+        try:
+            self._process = subprocess.Popen(
+                [
+                    'minio', 'server', '--quiet',
+                    '--address', f'{self.host}:{self.port}',
+                    '-C', str(self.path / 'config'),
+                    str(self.path / 'data'),
```
ludwigschwardt, line 116: Ah, now I know why `pathlib` exists :-)

PR #307, opened 2020-06-18
New 'time_offset' sensor property

ludwigschwardt:
This allows adjustment of the timestamps of any sensor, by adding a custom per-sensor time offset.

PR #317, opened 2021-01-15
Release v0.16 redux

ludwigschwardt:
I forgot to add the 0.16 release notes to master (it was done in the python2 branch).

Cherry-pick these in place instead of merging in order to keep the commit graph nice. It was already a belated release, so I prefer that the python2 branch remain unmerged to help find the spot on master that is the "actual" v0.16.

### DIFF
--- a/NEWS.rst
+++ b/NEWS.rst
@@ -9,7 +9,7 @@ History
 * Increase katstore search window by 600 seconds to find infrequent updates (#302)
 * Refactor SensorData to become a lazy abstract interface without caching (#292)
 * Refactor SensorCache to use MutableMapping (#300)
-* Fix rx_serial sensor use and file mode warning in HDF5v3 files (#298, #299)
+* Fix rx_serial sensor use and file mode warning in MVFv3 files (#298, #299)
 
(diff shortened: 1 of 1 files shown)


bmerry:
If you're cherry-picking release notes rather than merging, are you sure that actual changes described are all in master?

ludwigschwardt:
Yes, the `python2` branch splits from `master` in May 2020 and adds four commits - two cherry-picked from #307 and these two to make the release. So 0.16 is basically May 2020 with a bit of June thrown in, even though the release was cut in August.

If I merge `python2` into `master`, it would be easy to think that 0.16 contains everything up to Jan 2021, depending on how intently one stares at the merge commit and the version tag close by (I know I'd miss it 😄 ).

ludwigschwardt:
(For the record, I was wondering if it was time to make katdal Python 3 only, only to discover it has already happened :joy: Such is my memory of the lockdown period - katsdpmodels suffered the same fate :-))

In the somewhat near future I will strip out `future` and bring in `pytest` then.

bmerry (approved):

PR #318, opened 2021-01-19
Fix kwargs bug introduced by #316

ludwigschwardt:
Kwargs now flow to `TelstateDataSource.__init__` too, but if the CBID and/or stream name are specified in kwargs, they clash with the explicit versions passed to the constructor. The solution is to clear them out of kwargs after `view_l0_capture_stream` picked them up.

I've postponed the unit test improvements to a later PR since that part has grown quite a bit.


bmerry (approved):
Thanks, I can confirm that this fixes the katsdpimager tests.

PR #321, opened 2021-01-28
Remove Python 2 relics

ludwigschwardt:
Strip out `future` and all its ramifications, along with `subprocess32` and `mock`. Remove Python 2 workarounds.

This is probably best checked commit by commit. The first commit is big but reasonably straightforward.

### DIFF
--- a/katdal/chunkstore_s3.py
+++ b/katdal/chunkstore_s3.py
@@ -16,13 +16,6 @@
 
 """A store of chunks (i.e. N-dimensional arrays) based on the Amazon S3 API."""
 
-from __future__ import print_function, division, absolute_import
-from future import standard_library
-standard_library.install_aliases()  # noqa: E402
-from builtins import object
-import future.utils
-from future.utils import bytes_to_native_str, raise_from
-
 import contextlib
 import threading
 import urllib.parse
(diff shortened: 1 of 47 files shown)


katdal/chunkstore_s3.py
```diff
@@ -154,12 +154,10 @@ def _read_chunk(response):
     """Efficiently read NumPy array in NPY format from content of HTTP response."""
     data = response.raw
     # Workaround for https://github.com/urllib3/urllib3/issues/1540
-    # On Python 2, http.client.HTTPResponse doesn't implement readinto.
     # We also can't use the workaround if the content is encoded (e.g.
     # gzip compressed) because that's decoded in urllib3, not httplib.
     if ('Content-encoding' not in response.headers
-            and hasattr(data, '_fp')
-            and hasattr(data._fp, 'readinto')):
+            and hasattr(data, '_fp')):
```
bmerry, line 160: I think it might be worth leaving this, just in case there are conditions under which HTTPResponse ._fp doesn't provide a readinto (since it is a private member there is no guarantee of anything).

bmerry (changes requested):
Looks good! pyupgrade would probably also remove some other Python 2-isms (such as inheriting explicitly from `object`), but it might be best to put that in a separate PR to keep the iteration size small.

katdal/chunkstore_s3.py
ludwigschwardt: I wondered about that too... There is also `support_chunked_reads()` which is a public function which implicitly checks for `_fp`, but the same issues arise. I'll put it back.

ludwigschwardt:
Thanks for the tip about `pyupgrade` - never used that before but sure will now :-)

bmerry:
> I have another PR waiting in the wings that makes the jump to Python 3.6 (and PyJWT 2). I'll add it to that.

That may be a good time to run it - it has a `--py36-plus` flag that will also convert a lot of `.format` calls into f-strings.

ludwigschwardt:
> That may be a good time to run it - it has a `--py36-plus` flag that will also convert a lot of `.format` calls into f-strings.

I was going to do that by hand... Now I'm excited! :-)

bmerry (approved):

PR #322, opened 2021-01-29
Switch to PyJWT 2.0.1 (and Python 3.6)

ludwigschwardt:
The latest PyJWT has some breaking API changes. Two are relevant:

- The `jwt.encode` function now returns `str` and not `bytes`.
- The `jwt.decode` function won't verify any claims without verifying
  the signature first.

The second one is a bit of a pity. In PyJWT 1 there was a `verify`
parameter to `jwt.decode` that defaulted to True and enabled claim
verification. This was separate from the `verify_signature` option
(only serving as its default), so we could verify claims without
verifying the signature (which would be impossible in katdal without
the server secrets).

In PyJWT 2 there is no `verify` parameter and `verify_signature` serves
both purposes, so we are forced to do our own expiration time check.

Since PyJWT 2 requires Python 3.6+, the same now goes for katdal.


ludwigschwardt:
The deluge of Python 3.6 upgrades are next :-D

katdal/chunkstore_s3.py
```diff
@@ -226,21 +227,20 @@ def decode_jwt(token):
...
-        exp_time = time.strftime('%d-%b-%Y %H:%M:%S', time.gmtime(claims['exp']))
-        raise InvalidToken(token, 'Token expired at {} UTC, please '
-                                  'obtain a new one'.format(exp_time)) from err
     except jwt.exceptions.InvalidTokenError as err:
         raise InvalidToken(token, str(err)) from err
+    # Check if token has expired (PyJWT>=2 won't do this without signature verification)
+    if time.time() > claims.get('exp', np.inf):
+        exp_time = time.strftime('%d-%b-%Y %H:%M:%S', time.gmtime(claims['exp']))
+        raise InvalidToken(token, 'Token expired at {} UTC, please '
+                                  'obtain a new one'.format(exp_time))
```
bmerry, line 243: Maybe catch TypeError and OverflowError here, so that you'll get `InvalidToken` if `exp` is either not a numeric type or is out of the range of `time_t`.

requirements.txt
```diff
@@ -16,7 +16,7 @@ netifaces
 numba
 numpy
 pyephem
-pyjwt
+pyjwt==2.0.1
```
bmerry, line 19: Ideally we should instead update the pin in katsdpdockerbase.

bmerry (changes requested):
Looks good, some suggestions noted.

This will need a katsdpdockerbase update to bump pyjwt. Chris also has some katsdpauth requirements files which pin pyjwt==1.6.4; check with him to see if they interact with katdal at all.

ludwigschwardt:
> This will need a katsdpdockerbase update to bump pyjwt. Chris also has some katsdpauth requirements files which pin pyjwt==1.6.4; check with him to see if they interact with katdal at all.

I don't see any overlap between katdal and katsdpauth. Chris will be moving to pyjwt==2.0.1 in the next few months (already using it in dev code).

requirements.txt
ludwigschwardt: OK, should I do that first or as a separate step?

bmerry: I'd say prepare the PR for it, and merge both at the same time.

katdal/chunkstore_s3.py
ludwigschwardt: Done. Cast `exp` to int like PyJWT does, and also checked overflow.

bmerry (approved):

PR #324, opened 2021-02-01
Allow subset to be selected at katdal.open time

bmerry:
This avoids the cost of building up a massive graph only to select most of it away, and also saves memory since that large graph won't be kept around.

I'm marking this as draft for now because I'm not sure how best to handle pre-MVFv4 (see below). Nevertheless, please review to see if you're happy with the overall approach.

The problem is that an application that wants to use this feature won't necessarily know which version file it is opening, and hence the argument might or might not be ignored. It can check the file version after the fact, but then it would still need to hard-code knowledge of which versions support the feature, and that may change over time.

There are a few possibilities:
1. Just support it in all versions. I haven't looked at LazyIndexer to see how practical this is.
2. Store a flag in each class to indicate whether the feature is supported, which the caller can check.
3. Raise an exception if the caller tries to use the feature and it is not supported; the caller can either check it and fall back, or let it propagate. This has the advantage that callers that don't consider the possibly will get an exception rather than accessing the wrong data, but the disadvantage that falling back now requires re-opening the dataset which could be expensive (although pre-v4 datasets are generally cheap to open).

I also haven't considered concatenated data sets at all.

### DIFF
--- a/doc/tuning.rst
+++ b/doc/tuning.rst
@@ -96,6 +96,12 @@ You might be able to get a little more performance by using
 selections) rather than using :meth:`~.DataSet.select` with
 ``scans='track'``.
 
+When using MVF v4 one can also pass an `index` parameter to :meth:`katdal.open`
+which contains an index expression to slice a subset of the data (time and
+frequency). It is more limited than :meth:`DataSet.select` (it can only select
+contiguous ranges), but if a script is only interested in working on a subset
+of data, this method can be more efficient and uses less memory.
+
 Network versus local disk
 -------------------------
 When loading data from the network, latency is typically higher, and so
--- a/doc/tuning.rst
+++ b/doc/tuning.rst
@@ -96,11 +96,12 @@ You might be able to get a little more performance by using
 selections) rather than using :meth:`~.DataSet.select` with
 ``scans='track'``.
 
-When using MVF v4 one can also pass an `index` parameter to :meth:`katdal.open`
-which contains an index expression to slice a subset of the data (time and
-frequency). It is more limited than :meth:`DataSet.select` (it can only select
-contiguous ranges), but if a script is only interested in working on a subset
-of data, this method can be more efficient and uses less memory.
+When using MVF v4 one can also pass a `select` parameter to :meth:`katdal.open`
+which allows slicing a subset of the data (time and frequency). It is more
+limited than :meth:`DataSet.select` (it can only select contiguous ranges, and
+can only specify the selection in terms of channels and dumps), but if a script
+is only interested in working on a subset of data, this method can be more
+efficient and uses less memory.
 
 Network versus local disk
 -------------------------
--- a/doc/tuning.rst
+++ b/doc/tuning.rst
@@ -96,7 +96,7 @@ You might be able to get a little more performance by using
 selections) rather than using :meth:`~.DataSet.select` with
 ``scans='track'``.
 
-When using MVF v4 one can also pass a `select` parameter to :meth:`katdal.open`
+When using MVF v4 one can also pass a `preselect` parameter to :meth:`katdal.open`
 which allows slicing a subset of the data (time and frequency). It is more
 limited than :meth:`DataSet.select` (it can only select contiguous ranges, and
 can only specify the selection in terms of channels and dumps), but if a script
--- a/katdal/__init__.py
+++ b/katdal/__init__.py
@@ -113,6 +113,9 @@ def open(filename, ref_ant='', time_offset=0.0, **kwargs):
             avoiding the slow loading of real timestamps at the cost of
             slightly inaccurate label borders
 
+        See the documentation of :class:`VisibilityDataV4` for the keywords
+        it accepts.
+
     Returns
     -------
     data : :class:`DataSet` object
--- a/katdal/__init__.py
+++ b/katdal/__init__.py
@@ -121,7 +121,13 @@ def open(filename, ref_ant='', time_offset=0.0, **kwargs):
         Object providing :class:`DataSet` interface to file(s)
 
     """
-    filenames = [filename] if isinstance(filename, str) else filename
+    if isinstance(filename, str):
+        filenames = [filename]
+    else:
+        unexpected = set(kwargs.get('preselect', {})) - {'channels'}
+        if unexpected:
+            raise IndexError(f'Unsupported preselect key(s) for ConcatenatedDataSet: {unexpected}')
+        filenames = filename
     datasets = []
     for f in filenames:
         # V4 RDB file or live telstate with optional URL-style query string
@@ -130,6 +136,8 @@ def open(filename, ref_ant='', time_offset=0.0, **kwargs):
             dataset = VisibilityDataV4(open_data_source(f, **kwargs),
                                        ref_ant, time_offset, **kwargs)
         else:
+            if 'preselect' in kwargs:
+                raise TypeError('preselect is not supported for this format')
             dataset = _file_action('__call__', f, ref_ant, time_offset, **kwargs)
         datasets.append(dataset)
     return datasets[0] if isinstance(filename, str) else ConcatenatedDataSet(datasets)
--- a/katdal/chunkstore.py
+++ b/katdal/chunkstore.py
@@ -439,7 +439,7 @@ class ChunkStore(object):
             prefix = 'Chunk {!r}: '.format(chunk_name) if chunk_name else ''
             raise StandardisedError(prefix + str(e)) from e
 
-    def get_dask_array(self, array_name, chunks, dtype, offset=(), errors=0):
+    def get_dask_array(self, array_name, chunks, dtype, offset=(), index=(), errors=0):
         """Get dask array from the store.
 
         Handling of missing chunks is determined by the `errors` argument.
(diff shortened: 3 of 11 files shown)


doc/tuning.rst
```diff
@@ -96,6 +96,12 @@ You might be able to get a little more performance by using
 selections) rather than using :meth:`~.DataSet.select` with
 ``scans='track'``.
 
+When using MVF v4 one can also pass an `index` parameter to :meth:`katdal.open`
```
bmerry, line 99: I'm not sure if `index` is the best name, but it emphasises that it takes an index expression. Another candidate would be `select`, since it provides functionality similar to `DataSet.select`, but with different syntax.

ludwigschwardt: Yes, I share your sentiments -- the word `index` probably elicits no response in most users 😄  And `select` suggests that it will have the same syntax. I also considered `preselect`.

One option would be to go the actual `select` route, but only support `channels=slice()` and `dumps=slice()`. Then there is scope to expand this in future. And it also provides a simple way to handle older formats: simply slap it on top of the initial selection... All formats end their initialisation with a `select()` step (usually to set the spectral window, subarray and antenna list).

ludwigschwardt: Oh wait, the dataset dump/channel origin shifts so it is not a pure `select`. Mmm, then the `select` format might also not be that useful.

bmerry: Yes, was just about to point that out.

I am starting to like `select=dict(channels=slice(...), dumps=slice(...))`. It's familiar, it's explicit about which axis is which, and it's theoretically extensible. I was worried that it would confuse users into thinking it was equivalent to a `select` after opening, but since you made that mistake with the current name, maybe it's doomed to be confusing no matter what we call it.

bmerry: I've pushed a change to use `select`. Let me know what you think. At the moment I make the switch over to `index` between Datasource and ChunkStoreVisFlagsWeights, but I'm wondering if it would make more sense to move it to the interface between ChunkStoreVisFlagsWeights and ChunkStore (because ChunkStoreVisFlagsWeights is for specific sorts of data where it makes sense to talk about the "channels" axis, whereas ChunkStore is agnostic).

Also I'd still like your thoughts on how to handle pre-v4.

bmerry:
I had a look at how easy this would be to add to h5datav3. It would probably be possible by modifying `_vislike_indexer` to pad out the boolean arrays used to make the stage1 selection. But there would be some subtlety, particularly related to the workarounds for duplicated final dumps, and I'd be worried about introducing bugs. I think it would be better not to try supporting it until a mythical time when we convert the v1-3 support to use dask instead of LazyIndexer.

Currently I'm leaning towards option 3. @ludwigschwardt, what's your opinion?

For ConcatenatedDataSet, the kwargs are applied to the individual datasets. So it would work fine for selecting channels, but would behave strangely for dumps.

katdal/test/test_chunkstore.py
```diff
@@ -246,10 +261,59 @@ def test_dask_array_put_parts_get_whole(self):
...
+        self.put_dask_array('big_y2', np.s_[0:3, 30:60, 0:2])
+        array_name, dask_array, offset = self.make_dask_array('big_y2')
+        indices = [
+            (),
+            np.s_[:, :],
+            np.s_[0:8, 0:60, 0:2],
+            np.s_[..., 0:1],
+            np.s_[5:6, 29:31],
+            np.s_[5:5, 31:31]
+        ]   # TODO: use pytbest.mark.parametrize when converted to pytest
```
ludwigschwardt, line 299: typo (pytbest)

katdal/datasources.py
```diff
@@ -346,17 +346,34 @@ class TelstateDataSource(DataSource):
...
-                 upgrade_flags=True, van_vleck='off', **kwargs):
+                 upgrade_flags=True, van_vleck='off', select=None, **kwargs):
+        if select is None:
+            select = {}
+        unexpected = set(select.keys()) - {'channels', 'dumps'}
+        if unexpected:
+            raise IndexError("'select' can only specify 'channels' and 'dumps'")
+        for key, idx in select.items():
+            if not isinstance(idx, slice) or idx.step not in {None, 1}:
+                raise IndexError(f'{key} must be a slices with unit step')
```
ludwigschwardt, line 375: slices -> slice

katdal/chunkstore.py
```diff
@@ -456,19 +470,62 @@ def get_dask_array(self, array_name, chunks, dtype, offset=(), errors=0):
...
+            index = tuple(index)
+            offset = tuple(offset)
+
+        if any(offset):
             getter = _add_offset_to_slices(getter, offset)
         # Use dask utility function that forms the core of da.from_array
         dask_graph = da.core.getem(array_name, chunks, getter)
-        return da.Array(dask_graph, array_name, chunks, dtype)
+        array = da.Array(dask_graph, array_name, chunks, dtype)
+        return array[index]
```
ludwigschwardt, line 528: So `index` will be trivial if the selection is aligned with the original chunks?

doc/tuning.rst
ludwigschwardt: You could still consider `preselect`. For me it suggests `select` notation but the `pre` part warns that it is not exactly the same.

I always considered `select` a top-level feature at the `VisDataV4` level, and would probably have switched to `index` in `TelstateDataSource` already :-D But I like the current scheme and I'm comfortable with it moving down to `VisFlagsWeights` too.

ludwigschwardt (changes requested):
Looking good so far!

katdal/chunkstore.py
```diff
@@ -422,11 +437,10 @@ def _standard_errors(self, chunk_name=None):
             prefix = f'Chunk {chunk_name!r}: ' if chunk_name else ''
             raise StandardisedError(prefix + str(e)) from e
 
-    def get_dask_array(self, array_name, chunks, dtype, offset=(), errors=0):
+    def get_dask_array(self, array_name, chunks, dtype, offset=(), index=(), errors=0):
         """Get dask array from the store.
 
-        Any missing chunks are replaced with zeros, suppressing any
-        :exc:`ChunkNotFound` errors.
+        Handling of missing chunks is determined by the `errors` argument.
```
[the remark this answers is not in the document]
bmerry: Yeah, I think it's a left-over from long ago before we added the `errors` argument.

ludwigschwardt:
For pre-v4, I'd also go for option 3. I'm guilty of adding all sorts of kwargs to v4 that will be merrily ignored by v3 (applycal, van_vleck, upgrade_flags). I guess there is a stronger case to pay attention to preselection support.

katdal/chunkstore.py
bmerry: I'm not entirely sure what you mean by this. But `Array.__getitem__` has a check for a no-op index and just returns the original array without creating new layers.

bmerry:
> You could still consider preselect

Ok, that should be easy enough to rename. Will do when I fix up the other minor stuff.

katdal/chunkstore.py
ludwigschwardt: When I initially looked at this, it gave the impression that you are generating a bigger graph and then just selecting a part like before. This is before I worked my way through the rechunking code :-)

I then wondered why you would still need `index` but eventually realised that selecting e.g. channel 1 to 13 will require fractional chunks and therefore further indexing. But with a chunk per dump and a range of selected channels commensurate with the chunking, this `index` will basically be `[:, :]`. That's the part I'm asking about :-)

bmerry: Right, I'm with you now. Yes, this is here in case for the case of fractional chunks.

bmerry:
> I'm guilty of adding all sorts of kwargs to v4 that will be merrily ignored by v3 (applycal, van_vleck, upgrade_flags).

So arguably this is another instance of the general problem that passing an argument that the dataset class doesn't support is silently ignored. I've got an idea for dealing with that - I'll see how it pans out.

katdal/vis_flags_weights.py
```diff
@@ -276,6 +277,10 @@ class ChunkStoreVisFlagsWeights(VisFlagsWeights):
         Should be True if `corrprods` is `None`.
     van_vleck : {'off', 'autocorr'}, optional
         Type of Van Vleck (quantisation) correction to perform
+    index : tuple of slice, optional
+        Slice expression to apply to each array before combining them. At the
+        moment this can only have two elements (no slicing of baselines),
+        because ``weights_channel`` only has time and frequency dimensions.
```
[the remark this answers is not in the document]
bmerry: There would probably also be issues if, say, you sliced away the auto-correlations, since they'd then be missing when you tried to use them to compute weights.

ludwigschwardt:
> I've got an idea for dealing with that - I'll see how it pans out.

I'm all ears! I love the freedom of `**kwargs` but would also like to figure out a way to document them and/or verify them.

katdal/vis_flags_weights.py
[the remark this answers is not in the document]
ludwigschwardt: And the correlation products have a funny ordering, so a unit-step slice is not that useful (except, as you say, for distinguishing autocorrs and crosscorrs).

bmerry:
> I've got an idea for dealing with that - I'll see how it pans out.

My idea was to create a custom subclass of `collections.abc.Mapping` that tracks which elements have been accessed. But sadly it won't work because it won't survive `**kwargs` unpacking. We would need to change the various functions to take the arguments as a single mapping instead of via kwargs.

katdal/datasources.py
bmerry: Done

katdal/test/test_chunkstore.py
bmerry: Done

katdal/chunkstore.py
```diff
@@ -456,19 +470,62 @@ def get_dask_array(self, array_name, chunks, dtype, offset=(), errors=0):
...
+            chunks = [list(c) for c in chunks]   # Make mutable
+            shape = [sum(c) for c in chunks]
+            index = list(da.slicing.normalize_index(index, shape))
+            if not all(isinstance(idx, slice) and idx.step is None
+                       for idx in index):
+                raise IndexError('Only slices with unit step are valid indices in get_dask_array')
+            if not offset:
+                offset = [0] * len(shape)
+            else:
+                offset = list(offset)
```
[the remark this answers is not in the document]
bmerry: Done

ludwigschwardt:
> My idea was to create a custom subclass of `collections.abc.Mapping`

Yes, tried to work through that too and came to the same conclusion.

I wonder if function attributes could help, or some form of hierarchical registration of kwargs-aware functions, where each one could indicate leftover kwargs. If this state is global, it will have other issues too, though.

katdal/__init__.py
```diff
@@ -121,7 +121,13 @@ def open(filename, ref_ant='', time_offset=0.0, **kwargs):
         Object providing :class:`DataSet` interface to file(s)
 
     """
-    filenames = [filename] if isinstance(filename, str) else filename
+    if isinstance(filename, str):
+        filenames = [filename]
+    else:
+        unexpected = set(kwargs.get('preselect', {})) - {'channels'}
```
ludwigschwardt, line 127: The `dumps` could be made to work if you used the same rechunking scheme on a dataset level (i.e. adjust the slice per dataset and maybe get rid of some component datasets entirely).

Makes me think I should simply concatenate datasets on the dask level for starters 😄  The sensors will be the only challenge left then, I hope...

doc/tuning.rst
```diff
@@ -96,7 +96,7 @@ You might be able to get a little more performance by using
 selections) rather than using :meth:`~.DataSet.select` with
 ``scans='track'``.
 
-When using MVF v4 one can also pass a `select` parameter to :meth:`katdal.open`
+When using MVF v4 one can also pass a `preselect` parameter to :meth:`katdal.open`
```
ludwigschwardt, line 99: The commit message seems the wrong way around :-)

ludwigschwardt (approved):

katdal/__init__.py
bmerry: I don't see a critical use case for it right now, so even if it's possible, I don't want to hold up this PR to do it (this is blocking 5 PRs on katsdpimager that I'd like to get deployed on Wednesday).

I think there would also be a chicken-and-egg problem where you need to open the component datasets to know how many dumps they contain to know how to split up the slice across datasets before opening them.

bmerry:
> I wonder if function attributes could help, or some form of hierarchical registration of kwargs-aware functions, where each one could indicate leftover kwargs. If this state is global, it will have other issues too, though.

threading.local or contextvars could maybe help, but it would be ugly.

doc/tuning.rst
bmerry: You're right. Oh well.

katdal/__init__.py
ludwigschwardt: True!

doc/tuning.rst
ludwigschwardt: I wasn't sure if `->` could also mean "replaces" 😄

ludwigschwardt:
Before I forget: I wondered whether Dask had some tool in its shed that could do the slice + chunks merging that you did. I had a look at the `dask.array.rechunk` module and something along the lines of `intersect_chunks`, but nothing popped out.

PR #327, opened 2021-02-08
fix llvmlite version for python 2

adriaanph:
llvmlite 0.32.0 was the last to support python 2, but the default dependency check looked for >=0.31.0dev0

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -61,7 +61,7 @@ setup(name='katdal',
       install_requires=['numpy >= 1.12.0', 'katpoint >= 0.9', 'h5py >= 2.3', 'numba',
                         'katsdptelstate[rdb] >= 0.8', 'dask[array] >= 1.2.1',
                         'requests >= 2.18.0', 'pyjwt', 'future',
-                        'cityhash >= 0.2.2', 'llvmlite==0.32.0'],
+                        'cityhash >= 0.2.2', 'llvmlite < 0.32; python_version < "3"'],
       extras_require={
           'ms': ['python-casacore >= 2.2.1'],
           's3': [],
--- a/setup.py
+++ b/setup.py
@@ -61,7 +61,7 @@ setup(name='katdal',
       install_requires=['numpy >= 1.12.0', 'katpoint >= 0.9', 'h5py >= 2.3', 'numba',
                         'katsdptelstate[rdb] >= 0.8', 'dask[array] >= 1.2.1',
                         'requests >= 2.18.0', 'pyjwt', 'future',
-                        'cityhash >= 0.2.2'],
+                        'cityhash >= 0.2.2', 'llvmlite==0.32.0'],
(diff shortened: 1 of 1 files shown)


adriaanph:
@ludwigschwardt please review - github doesn't allow me to add reviewers to this PR.

ludwigschwardt:
This could be called a Numba bug (see e.g. numba/numba#5035). Whether there is appetite to backport the `max_llvmlite_version` of  numba/numba#5046 to 0.47.0 and cut a 0.47.1 release is unclear, but that would be the cleanest solution.

It could even be a pip bug, since pip is picking a newer llvmlite source package from the local cache, even though it is incompatible with the Python version. At least that's what happens on my machine (and I expect on yours too).

setup.py
```diff
@@ -61,7 +61,7 @@
       install_requires=['numpy >= 1.12.0', 'katpoint >= 0.9', 'h5py >= 2.3', 'numba',
                         'katsdptelstate[rdb] >= 0.8', 'dask[array] >= 1.2.1',
                         'requests >= 2.18.0', 'pyjwt', 'future',
-                        'cityhash >= 0.2.2'],
+                        'cityhash >= 0.2.2', 'llvmlite==0.32.0'],
```
ludwigschwardt, line 64: A more appropriate change would be:
```
'llvmlite < 0.32; python_version < "3"'
```
llvmlite 0.32.0 is already too late, since it is Python 3 only. You want 0.31.0 (and any patch releases on top of that if it ever happens).

adriaanph: awesome, thanks for background & the proposed refinement.

ludwigschwardt:
@bmerry, any opinion on this? Should we go ahead and fix this in katdal, or get the Numba guys to cut a 0.47.1? :-D

bmerry:
It looks like the problem is that llvmlite didn't set `python_requires` in 0.32. Unfortunately that can't be fixed after the fact; even if there is a 0.32.1 release that has it, that'll just stop pip picking up 0.32.1, but won't stop it picking up 0.32.

My opinionated opinion is that we shouldn't be spending *any* effort on supporting Python 2 and that people who insist on using it have to solve their own problems, but if you are going to do something, this seems like the simplest solution.

ludwigschwardt:
Thanks, I haven't even looked at llvmlite as the culprit 🙂 

This [discussion](https://github.com/numba/llvmlite/issues/471#issuecomment-631935869) also points out this issue and references [PEP 592](https://www.python.org/dev/peps/pep-0592/). It looks like there is now a way to "yank" a released file from PyPI - good to remember.

adriaanph:
thanks, gentlemen.

this change does not have zero value - it is critical for continued support of katholog, which is currently still stuck in python2. it would be wasteful to discard any useful work that is generated within the organisation.

PR #329, opened 2021-03-07
Refactor requirement files

bmerry:
- Remove recursive dependencies
- Add -c options to pin the recursive dependencies
- Add a pyproject.toml to ensure katversion is available during setup

See SPR1-833.

### DIFF
--- a/requirements.txt
+++ b/requirements.txt
@@ -1,30 +1,13 @@
+-c https://raw.githubusercontent.com/ska-sa/katsdpdockerbase/master/docker-base-build/base-requirements.txt
+
 botocore
-certifi
-chardet
 cityhash
-dask
-docutils
-ephem
-jmespath
-future
+dask[array]
 h5py
-idna
-katversion
-llvmlite
-msgpack
-netifaces
 numba
 numpy
-pyephem
 pyjwt
-python-dateutil
-python-lzf
-rdbtools
-redis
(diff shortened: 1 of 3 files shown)


requirements.txt
```diff
@@ -1,30 +1,13 @@
...
-rdbtools
-redis
 requests
-six
-toolz
-urllib3
 
 katpoint @ git+https://github.com/ska-sa/katpoint
-katsdptelstate @ git+https://github.com/ska-sa/katsdptelstate
+katsdptelstate[rdb] @ git+https://github.com/ska-sa/katsdptelstate
```
ludwigschwardt, line 13: It's nice that you can now add the extras for URLs too.

test-requirements.txt
```diff
@@ -1,2 +1,4 @@
+-c https://raw.githubusercontent.com/ska-sa/katsdpdockerbase/master/docker-base-build/base-requirements.txt
```
ludwigschwardt, line 1: Was the base requirements implicit before for test requirements?

ludwigschwardt (approved):

test-requirements.txt
bmerry: And in fact it still is, but this allows `pip install -r test-requirements.txt` to obtain the right versions.

PR #331, opened 2021-04-14
SPR1-910 Support h5py 3.0 API in MVF3

ludwigschwardt:
The latest major version of h5py (3.0, released Oct 2020) totally
revamps string handling. All string-valued attributes in MVF3 now
return `str` in Python 3 (while h5py 2.x returns `bytes`).

Extend `telstate_decode` to handle both versions and add some tests.

### DIFF
--- a/katdal/test/test_sensordata.py
+++ b/katdal/test/test_sensordata.py
@@ -23,7 +23,7 @@ from nose.tools import assert_equal, assert_in, assert_not_in, assert_raises, as
 from unittest.mock import Mock
 
 from katdal.sensordata import (SensorCache, SensorData, SimpleSensorGetter, to_str,
-                               remove_duplicates_and_invalid_values)
+                               remove_duplicates_and_invalid_values, telstate_decode)
 
 
 def assert_equal_typed(a, b):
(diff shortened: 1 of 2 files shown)


katdal/test/test_sensordata.py
```diff
@@ -71,6 +71,20 @@ def test_numpy_object(self):
         np.testing.assert_array_equal(to_str(a), b)
 
 
+@mock.patch('katsdptelstate.encoding._allow_pickle', True)
```
bmerry, line 74: I haven't checked if it's necessary here, but in telstate unit tests that do unpickling I've done this to prevent the test emitting a warning:
```suggestion
@mock.patch('katsdptelstate.encoding._allow_pickle', True)
@mock.patch('katsdptelstate.encoding._warn_on_pickle', False)
```

bmerry (changes requested):
Looks good. Just one suggestion - if you're happy with it feel free to merge without another round.

Will this fix SPR1-783?

katdal/test/test_sensordata.py
ludwigschwardt: I haven't seen any warnings during testing but I'll add this anyway.

ludwigschwardt:
Yes, this fixes SPR1-783. Now I know that my medium-term memory is shorter than two months 😂

PR #333, opened 2021-04-21
KAT-1196: Support planets and other non-radec targets in mvftoms

ludwigschwardt:
It is time...

@mauch, I used the script you modified for @nirujmohanr as basis, and also threw in scans for good measure.

### DIFF
--- a/katdal/ms_extra.py
+++ b/katdal/ms_extra.py
@@ -863,7 +863,7 @@ def populate_pointing_dict(num_antennas, observation_duration, start_time, phase
     start_time : float
         Start time of observation, as a Modified Julian Date in seconds
     phase_center : array of float, shape (2,)
-        Direction of phase center, in ra-dec coordinates as 2-element array
+        Direction of phase center, in (az, el) coordinates as 2-element array (?)
     pointing_name : string, optional
         Name for pointing
 
--- a/scripts/mvftoms.py
+++ b/scripts/mvftoms.py
@@ -611,10 +611,6 @@ def main():
                     if options.verbose:
                         print(f'{prefix} skipped - too short')
                     continue
-                if target.body_type != 'radec':
-                    if options.verbose:
-                        print(f"{prefix} skipped - target '{target.name}' not RADEC")
-                    continue
                 print(f"{prefix} loaded. Target: '{target.name}'. Writing to disk...")
 
                 # Get the average dump time for this scan (equal to scan length
@@ -625,13 +621,13 @@ def main():
                 utc_seconds = dataset.timestamps[:]
                 # Update field lists if this is a new target
                 if target.name not in field_names:
-                    # Since this will be an 'radec' target, we don't need antenna
-                    # or timestamp to get the (astrometric) ra, dec
-                    ra, dec = target.radec()
-
                     field_names.append(target.name)
+                    # Set direction of the field center to the (ra, dec) at the start
+                    # of the first valid scan, based on the reference (catalogue) antenna.
+                    field_time = katpoint.Timestamp(utc_seconds[0])
+                    ra, dec = target.radec(field_time)
                     field_centers.append((ra, dec))
-                    field_times.append(katpoint.Timestamp(utc_seconds[0]).to_mjd() * 60 * 60 * 24)
+                    field_times.append(field_time.to_mjd() * 60 * 60 * 24)
                     if options.verbose:
                         print(f"Added new field {len(field_names) - 1}: '{target.name}' {ra} {dec}")
                 field_id = field_names.index(target.name)
--- a/scripts/mvftoms.py
+++ b/scripts/mvftoms.py
@@ -603,9 +603,9 @@ def main():
                 s = time.time()
                 scan_len = dataset.shape[0]
                 prefix = f'scan {scan_ind:3d} ({scan_len:4d} samples)'
-                if scan_state != 'track':
+                if scan_state not in {'track', 'scan'}:
                     if options.verbose:
-                        print(f"{prefix} skipped '{scan_state}' - not a track")
+                        print(f"{prefix} skipped '{scan_state}' - not a track or scan")
                     continue
                 if scan_len < 2:
                     if options.verbose:


katdal/ms_extra.py
```diff
@@ -863,7 +863,7 @@ def populate_pointing_dict(num_antennas, observation_duration, start_time, phase
     start_time : float
         Start time of observation, as a Modified Julian Date in seconds
     phase_center : array of float, shape (2,)
-        Direction of phase center, in ra-dec coordinates as 2-element array
+        Direction of phase center, in (az, el) coordinates as 2-element array (?)
```
mauch, line 866: What does the (?) signify here?

mauch (approved):
Just a minor question about the update to the POINTING table docstring, but otherwise looks good.

Great to see a KAT-7 JIRA ticket getting some attention :)

scripts/mvftoms.py
```diff
@@ -625,13 +621,13 @@ def _cp_index(a1, a2, pol):
...
                 # Update field lists if this is a new target
                 if target.name not in field_names:
-                    # Since this will be an 'radec' target, we don't need antenna
-                    # or timestamp to get the (astrometric) ra, dec
-                    ra, dec = target.radec()
-
                     field_names.append(target.name)
+                    # Set direction of the field center to the (ra, dec) at the start
+                    # of the first valid scan, based on the reference (catalogue) antenna.
+                    field_time = katpoint.Timestamp(utc_seconds[0])
```
spassmoor, line 627: This is very similar to the hack that I did in 2013 when I image Jupiter.  I used one ra,dec coordinate for the full observation. Does this give one coordinate pair  per track or  one coordinate pair per observation ?

I also noted star trails in the automated reduction that uses katdal->obit  .  How are they doing it  @mauch  ?

spassmoor (approved):

scripts/mvftoms.py
mauch: In katsdpcontim the target ra,dec is derived (for all observation time) from the the first timestamp of the observation.

katdal/ms_extra.py
ludwigschwardt: I'm still not sure... Ideally this should be an array of shape e.g. `(num_times, 2, num_antennas)` but a single phase center for all antennas and all times is definitely not it.

scripts/mvftoms.py
ludwigschwardt: Yes Sean, I was pleased to see that in KAT-1196 after I rediscovered the ticket. I'm also glad you didn't close it, which provided me with a ready-to-use ticket :-)

This gives one coordinate pair per observation. It doesn't use the first timestamp of the entire observation, but the first timestamp of a track on the actual source, mirroring CASA.

I tried the option of one coordinate pair per track, but that would imply the correspondence:
```
MeerKAT track <-> CASA field
```
The reality is the following correspondence:
```
MeerKAT track <-> CASA scan
MeerKAT target <-> CASA field / source
```
So you are stuck with a single (ra, dec) per target.

This "hack" is also apparently used by CASA in the `fixplanets` task. There you have 3 options:
  - Use the POINTING (az, el) table to get (ra, dec) for either the first timestamp of the field (the default option) or the median unflagged timestamp, using a given reference antenna (defaults to antenna 0).
  - Use an ephemeris, but this is a bit obscure. It gets added to an EPHEMERIS table while the FIELD (ra, dec) = (0, 0).
  - Override the field direction explicitly.

It seems that the three field centers don't really matter, unless you plan to run `fixvis` to change the UVW. As long as the delay tracking did the right thing for your source, CASA is happy.

o-smirnov:
Thanks @ludwigschwardt, I started the download and I can see the Jupiter scans arriving now. :+1:

scripts/mvftoms.py
ludwigschwardt: Or maybe the field centers are used to determine the absolute coordinates of the imaging grid?

o-smirnov: OK, so I'm just *slightly* confused. I think I understand, but please confirm that I do:

I see one RA/Dec entry for Jupiter in FIELD. So this is position of Jupiter at at the start of the first scan on it, correct? Thereafter, the correlator delay (and UVWs, etc.) is tracking Jupiter, correct? So it's a "kludge" in the sense that only the first timeslot's delay is really towards the given RA/Dec of its stated field.

I should therefore expect the imager to give me a proper image of Jupiter, and smudged tracks for the background sources, correct?

ludwigschwardt: Correct, correct and correct.

It seems that CASA rolls the same way. If I look at the Jupiter demo in the [CASA example scripts](https://casa.nrao.edu/casa_scripts.shtml), the MS has the following FIELD table (see the [script](https://casa.nrao.edu/Doc/Scripts/jupiter6cm_demo.py)):
```
# Fields: 13
#   ID   Name          Right Ascension  Declination   Epoch   
#   0    0137+331      01:37:41.30      +33.09.35.13  J2000   
#   1    0813+482      08:13:36.05      +48.13.02.26  J2000   
#   2    0542+498      05:42:36.14      +49.51.07.23  J2000   
#   3    0437+296      04:37:04.17      +29.40.15.14  J2000   
#   4    VENUS         04:06:54.11      +22.30.35.91  J2000   
#   5    0521+166      05:21:09.89      +16.38.22.05  J2000   
#   6    1411+522      14:11:20.65      +52.12.09.14  J2000   
#   7    1331+305      13:31:08.29      +30.30.32.96  J2000   
#   8    MARS          14:21:41.37      -12.21.49.45  J2000   
#   9    NGC7027       21:07:01.59      +42.14.10.19  J2000   
#   10   NEPTUNE       20:26:01.14      -18.54.54.21  J2000   
#   11   URANUS        21:15:42.83      -16.35.05.59  J2000   
#   12   JUPITER       00:55:34.04      +04.45.44.71  J2000   
```
I have checked that Venus's (ra, dec) in FIELD is close to what I get for the VLA core at the start of the first track on Venus. The small difference is probably because the default antenna has ID 0, which is way out in the West arm:
```
# Antennas: 27:
#   ID   Name  Station   Diam.    Long.         Lat.         
#   0    1     VLA:W9    25.0 m   -107.37.25.1  +33.53.51.0
```
The FIELD table supports fancy time-varying (ra, dec), but I've never seen an example of it in actual use.

o-smirnov: Yep, and I guess we look at moving objects rarely enough to worry about a "proper" solution. But the conceptual problem now is, there's no info in the MS of what the delay centres **actually** were. Is there a way I can get them out by hand using katdal? Or is there a cheeky copy in POINTING?

(The simple alternative is to use casacore.measures to get Jupiter's ra/dec at each point in time, of course, but then I just have to take your word for it that that's exactly where the correlator was tracking...)

ludwigschwardt: In katdal you could try:
```Python
d = katdal.open(...)
d.catalogue['Jupiter'].radec(d.timestamps)
```
This does not consider target offsets in the delay model but those would not normally be used.

o-smirnov: IMG_20210422_192506_433

ludwigschwardt: Mesmerising 🤩 

The basic parameters look good:
- Jupiter was 5.09 AU from us at the Conjunction and with 140 000 km diameter you get an extent of ~38" ✅ 
- Jupiter's (ra, dec) changed by about 210" in 6.5 hours, which is about the length of the star trails ✅ 

I think that's Jupiter :-)

o-smirnov: And now, the movie version! http://ratt.radiopadre.net/~oms/jove.gif. This is about 4 hours' worth.

mauch:
Ooh! Thats cool. Is that Saturn to the North moving slowly WRT the
background?

o-smirnov:
Yep, it can only be Saturn. This is the "conjunction" track.

scripts/mvftoms.py
ludwigschwardt: I was going to ask whether there would be any parallactic rotation of Jupiter's radio axis with respect to (ra, dec) and now I know, thanks to the movie :-D

And the spacing between Jupiter and Saturn also seems right for the conjunction (~0.1º).

I like the radio galaxy photobombing this auspicious moment :joy:

o-smirnov:
Every good movie deserves a sequel, and I gotta say, this is one case of the sequel blowing away the original: http://ratt.radiopadre.net/~oms/jove3b.png

ludwigschwardt:
Wow, two photobombers! Awesome stuff.

o-smirnov:
Look carefully, you might spot a transient going **pop** near one of them. And the steady little dot in the centre of the upper left inset is Callisto.

Is there a way to get the actual per-integration delay centres out of katdal? I'm trying to figure out if the shenanigans we engaged in with coordinates in the MS have introduced a small systematic offset into the astrometry...

o-smirnov:
> I'm trying to figure out if the shenanigans we engaged in with coordinates in the MS have introduced a small systematic offset into the astrometry...

Actually, I think we're all good. I checked against an ephemeris calculator, and Saturn is spot-on.

bennahugo:
You can query elev and az directly from katdal?

o-smirnov:
I would need to be taught. But not this time. The ephemeris checks out, the astrometry is spot-on. Onwards and upwards.

PR #335, opened 2021-05-25
SPR1-897: Expose the raw flags of MVFv4 datasets

ludwigschwardt:
Expose the raw flags of MVFv4 datasets, which are the uint8s stored in the chunk store (L0 or L1), augmented by the `data_lost` bit that also covers the corresponding vis and weights chunks, and the `postproc` bit produced by the optional applycal step. It is therefore the final flag bits before they are combined into bools, which tells you why data points are flagged.

The `raw_flags` indexer will honour stage 1 and stage 2 selections done on the time, frequency and corrprod axes, but ignores flag selections. It always returns all available flag bits.

Also refactor some of the cruft in `_set_keep` while we are at it.

### DIFF
--- a/katdal/visdatav4.py
+++ b/katdal/visdatav4.py
@@ -549,46 +549,43 @@ class VisibilityDataV4(DataSet):
 
         """
         DataSet._set_keep(self, time_keep, freq_keep, corrprod_keep, weights_keep, flags_keep)
-        update_all = time_keep is not None or freq_keep is not None or corrprod_keep is not None
-        update_flags = update_all or flags_keep is not None
         if not self.source.data:
             self._vis = self._weights = self._flags = self._raw_flags = self._excision = None
-        elif update_flags:
-            # Create first-stage index from dataset selectors. Note: use
-            # the member variables, not the parameters, because the parameters
-            # can be None to indicate no change
-            stage1 = (self._time_keep, self._freq_keep, self._corrprod_keep)
-            if update_all:
-                # Cache dask graphs for the data fields
-                self._vis = DaskLazyIndexer(self._corrected.vis, stage1)
-                self._weights = DaskLazyIndexer(self._corrected.weights, stage1)
-                self._raw_flags = DaskLazyIndexer(self._corrected.flags, stage1)
-            flag_transforms = []
-            if ~self._flags_select != 0:
-                # Copy so that the lambda isn't affected by future changes
-                select = self._flags_select.copy()
-                flag_transforms.append(lambda flags: da.bitwise_and(select, flags))
-            flag_transforms.append(lambda flags: flags.view(np.bool_))
-            self._flags = DaskLazyIndexer(self._corrected.flags, stage1, flag_transforms)
-            unscaled_weights = self._corrected.unscaled_weights
-            if unscaled_weights is None or self.accumulations_per_dump is None:
-                self._excision = None
-            else:
-                # The maximum / expected number of CBF dumps per SDP dump
-                cbf_dumps_per_sdp_dump = round(self.dump_period / self.cbf_dump_period)
-                accs_per_sdp_dump = np.float32(self.accumulations_per_dump)
-                accs_per_cbf_dump = accs_per_sdp_dump / np.float32(cbf_dumps_per_sdp_dump)
-                # Each unscaled weight represents the actual number of accumulations per SDP dump.
-                # Correct most of the weight compression artefacts by forcing each weight to be
-                # an integer multiple of CBF n_accs, and then convert it to an excision fraction.
-                def integer_cbf_dumps(w):
-                    return da.round(w / accs_per_cbf_dump) * accs_per_cbf_dump
+            return
+        # Create first-stage index from dataset selectors. Note: use
+        # the member variables, not the parameters, because the parameters
+        # can be None to indicate no change
+        stage1 = (self._time_keep, self._freq_keep, self._corrprod_keep)
+        # Cache dask graphs for the data fields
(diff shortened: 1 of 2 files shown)


ludwigschwardt:
I would review one commit at a time.

The second commit is mostly whitespace changes, so you might want to use `diff -b` there.

ludwigschwardt:
This is *slightly* confusing to developers (not users) because there is also a `flags_raw` dask key hidden in the bowels that only refers to the chunk store uint8s. I figured folks would want all the flag bits they can get.

The flip side is that this indexer is quite slow... From my laptop at home, I get the following in a spot check:
- `d.vis[1, 2, 3]`: 8-10 s
- `d.weights[2, 2, 3]`: 16 s
- `d.raw_flags[10, 2, 3]`: 74 s
- `d.flags[12, 2, 3]`: 79 s

I guess it's the `data_lost` machinery.

ludwigschwardt:
I have done basic tests where I opened RDB files and selected and displayed flags (both normal and concatenated datasets).

katdal/visdatav4.py
```diff
@@ -560,13 +560,20 @@ def _set_keep(self, time_keep=None, freq_keep=None, corrprod_keep=None,
...
         flag_transforms = []
         if ~self._flags_select != 0:
-            # Copy so that the lambda isn't affected by future changes
+            # Copy so that the closure isn't affected by future changes
             select = self._flags_select.copy()
-            flag_transforms.append(lambda flags: da.bitwise_and(select, flags))
-        flag_transforms.append(lambda flags: flags.view(np.bool_))
+            def bitwise_and(flags): return da.bitwise_and(select, flags)
+            flag_transforms.append(bitwise_and)
+        # Convert uint8 to bool by ORing all the bits
```
bmerry, line 571: It doesn't OR all the bits - they're still there. It's just that numpy treats any non-zero value as truthy.

bmerry (changes requested):
Looks good apart from my nitpick. It's a pity that flags and raw_flags can't share the stage1-indexing. It could probably be done if DaskLazyIndexer was split to separate the stage1 indexing from the transformations, but it doesn't seem worth worrying about for now for a niche feature.

ludwigschwardt:
Another way is to allow a DaskLazyIndexer to be constructed from another one, and when the `dataset` is realised, it does a normal getitem on it, before adding more transforms.

katdal/visdatav4.py
bmerry: Would that survive the flag selection transformation?

ludwigschwardt: You would then only see the selected flags, but that's OK. You could just select `flags='all'` beforehand.

ludwigschwardt:
```
In [30]: d.select(dumps=12, channels=slice(10,20), flags='static,cam')                             
In [38]: with ProgressBar(): 
    ...:     z = d.flags[0, :3, :5] 
    ...:                                                                                           
[########################################] | 100% Completed |  2min  5.8s

In [39]: z                                                                                         
Out[39]: 
array([[ True,  True,  True,  True,  True],
       [ True,  True,  True,  True,  True],
       [ True,  True,  True,  True,  True]])
In [41]: import numpy as np                                                                        

In [42]: z.view(np.uint8)                                                                          
Out[42]: 
array([[2, 2, 2, 2, 2],
       [2, 2, 2, 2, 2],
       [2, 2, 2, 2, 2]], dtype=uint8)

In [43]: d.select(dumps=12, channels=slice(10,20), flags='all')                                    

In [44]: with ProgressBar(): 
    ...:     z = d.flags[0, :3, :5] 
    ...:                                                                                           
[########################################] | 100% Completed |  1min 15.7s

In [45]: z                                                                                         
Out[45]: 
array([[ True,  True,  True,  True,  True],
       [ True,  True,  True,  True,  True],
       [ True,  True,  True,  True,  True]])

In [46]: z.view(np.uint8)                                                                          
Out[46]: 
array([[66, 66, 66, 66, 66],
       [66, 66, 66, 66, 66],
       [66, 66, 66, 66, 66]], dtype=uint8)
```

ludwigschwardt: The question is whether it is sufficient to tell end users about this trick. Or should we expose it like in `raw_flags`?

bmerry: I don't think that should be how end users access this - it's pretty fragile and would prevent us ever doing anything that collapsed the flags into a 0/1 bool.

ludwigschwardt: That's a con, agreed. A pro is that the trick is basically instant, while `raw_flags` needs another trip to the archive.

bmerry: That's fixable though - and anyway, it'll only make a difference if someone is loading both `flags` and `raw_flags` in the same get call, which seems like a fairly pointless use case since you could work directly off `raw_flags` and forgo `flags`.

bmerry:
Incidentally, https://github.com/dask/dask/pull/7655 should be an interesting one to watch. It might make a lot of the laziness in DaskLazyIndexer redundant since slicing expressions will become very cheap to construct and only slow to compute.

katdal/visdatav4.py
ludwigschwardt: My use case is:
 - Use `flags` as before.
 - Ask: "why is that patch over there flagged?"
 - Invoke `raw_flags`.
 
I'm trying to shield the user from having to do bitwise ands and view-as-bools in the first place :-D

ludwigschwardt:
You can have a look again. I've implemented nested indexers but will send that in a new PR (on the same ticket :-D).

bmerry (approved):

PR #336, opened 2021-05-26
SPR1-897: Allow nested DaskLazyIndexers

ludwigschwardt:
The main use case is to slap some more transforms onto an existing indexer (think `flags` piggy-backing on `raw_flags`) but it also allows N-stage indexing. The first indexer's `dataset` Dask array becomes the original dataset for the second indexer (lazily), and an additional selection and transform set will then be applied to it to form the `dataset` of the second indexer.

It's only used for the flags indexers for now.

### DIFF
--- a/katdal/lazy_indexer.py
+++ b/katdal/lazy_indexer.py
@@ -532,17 +532,6 @@ class DaskLazyIndexer:
                 self._orig_dataset = None
             return self._dataset
 
-    def add_transform(self, transform):
-        """Add another transform to the end of the transform chain.
-
-        The `transform` is a callable that takes a dask array and returns
-        another dask array.
-        """
-        with self._lock:
-            if self._dataset is not None:
-                self._dataset = transform(self._dataset)
-            self._transforms.append(transform)
-
     def __getitem__(self, keep):
         """Extract a selected array from the underlying dataset.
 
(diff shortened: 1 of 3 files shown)


bmerry:
This scares me somewhat, because DaskLazyIndexer is mutable (via `add_transform`). Whether a transform added to an upstream indexer will be applied in a downstream indexer depends on whether the downstream indexer has been materialised before the transform was added.

I can see a few options that would make this less of spooky action at a distance:

1. When creating a downstream indexer, set a flag on the upstream indexer to make it immutable, and raise an exception from `add_transform`.
2. Take some action to ensure that added transforms propagate. For example, each indexer could keep a WeakSet of downstream indexers and invalidate them on alteration (might need some careful thinking about lock ordering), or the downstream indexer could mirror the list of transforms and check for changes on each access.
3. Remove `add_transform`. It's not used at all inside katdal and only in one place in the ska-sa Github organisation (in katsdpscripts), so it shouldn't need too much work to eliminate it.

ludwigschwardt:
1. 🤔 
2. 🤯 
3. 🥳 

Nesting allows you to add transforms while remaining immutable, so who needs `add_transform` anymore?

mattieudv:
What about step 2.5 🚿

ludwigschwardt:
Apparently I added `add_transform` late in the game in #180 to cater for the impending `applycal`, but `applycal` doesn't even use it... It uses `da.map_blocks` and friends instead. Tsk tsk. This is why you should not "prepare to prepare" (to quote #180).

bmerry:
@ludwigschwardt I'll assume this is on hold until you've sorted out add_transform.

ludwigschwardt:
OK, `add_transform` is gone :-)

bmerry (approved):

PR #337, opened 2021-06-17
Better name and data source name handling

cchristelis:
A better experience for 
```
>>> d = katdal.open(rdb_file)
>>> print(d)
```
also added
```
d.url
```


ludwigschwardt (changes requested):
Let's discuss this first... I've only realised now that I probably want to get rid of the `AttrSensors`, `VisWeightFlags` and `DataSource` names altogether :-D

katdal/test/test_datasources.py
```diff
@@ -23,6 +23,7 @@
 
 import numpy as np
 from nose.tools import assert_raises
+
```
ludwigschwardt, line 26: Technically there should not be an empty line here. These packages are all in the category of "external dependencies", i.e. not in the standard library and not in `katdal`.

katdal/dataset.py
```diff
@@ -358,10 +358,11 @@ class DataSet:
 
     """
 
-    def __init__(self, name, ref_ant='', time_offset=0.0):
+    def __init__(self, name, ref_ant='', time_offset=0.0, url=''):
```
ludwigschwardt, line 361: The base `DataSet` has tons of attributes that are not passed as parameters to the initialiser. I'm hesitant to add `url` at this level because it's kind of arbitrary. In fact, I'm thinking of getting rid of the rest in a future release too.

katdal/test/test_datasources.py
```diff
@@ -296,3 +298,23 @@ def test_construction_from_url(self):
...
+        view, cbid, sn, _, _ = make_fake_data_source(self.telstate, self.store, (20, 16, 40))
+        rdb_dir = os.path.join(self.tempdir, cbid)
+        os.mkdir(rdb_dir)
+        rdb_filename = os.path.join(rdb_dir, f'{cbid}_{sn}.rdb')
+        self.telstate['capture_block_id'] = cbid
+        self.telstate['stream_name'] = sn
+        with RDBWriter(rdb_filename) as rdbw:
+            rdbw.save(self.telstate)
+        source_from_file = open_data_source(rdb_filename)
+        source_name = f'{cbid}_{sn}'
```
ludwigschwardt, line 318: I would just add the following three lines to `test_construction_from_url` to avoid code duplication... Or spin out everything above this to a helper function.

ludwigschwardt (changes requested):
We're moving along, thanks for the effort :-)

katdal/test/test_datasources.py
```diff
@@ -296,3 +297,15 @@ def test_construction_from_url(self):
             open_data_source('ftp://unsupported')
         with assert_raises(DataSourceNotFound):
             open_data_source(rdb_filename[:-4])
+        source_name = f'{cbid}_{sn}'
+        assert source_from_file.name == source_name
+        assert rdb_filename in source_from_file.url
+
+    def test_name_change(self):
+        # Ideally this test should do the following
+        # >>> d = katdal.open(RDBfile)
```
ludwigschwardt, line 306: This kind of test will have to reside in `test_dataset` or `test_visdatav4`, not `test_datasources`.

katdal/visdatav4.py
```diff
@@ -445,14 +446,11 @@ def _before(date):
                 corrected_weights = self._make_corrected(apply_weights_correction,
                                                          self.source.data.weights)
                 unscaled_weights = self.source.data.unscaled_weights
-                name = self.source.data.name
                 # Acknowledge that the applycal step is making the L1 product
-                if 'sdp_l0' in name:
-                    name = name.replace('sdp_l0', 'sdp_l1')
-                else:
-                    name = name + ' (corrected)'
+                if 'sdp_l0' in self.name:
```
ludwigschwardt, line 450: I would harvest the cal streams from `self.applycal_products` like `{cp.split('.')[0] for cp in self.applycal_products}`, then do two ifs:
```Python
if 'sdp_l0' in self.name and 'l1' in cal_streams:
    ...
if 'sdp_l1' in self.name and 'l2' in cal_streams:
    ...
```

ludwigschwardt (changes requested):

cchristelis:
Hi @ludwigschwardt , I don't think I quite understand how the L2 streams should best be handled. I think I'll try and find a gap somewhere where we can jump on a quick call...

ludwigschwardt (changes requested):
We should also use the opportunity to fix the V1/V2/V3 cases. I have three test files lying around (about 150 MB).

The name could be `pathlib.Path(name).stem`, which would be the closest thing to a CBID from that era, and the URL could be `urllib.parse.urlparse(name, scheme='file').geturl()`.

cchristelis:
Thanks for the update on handling historic h5 files. I think it's a good approach to have the URL.

ludwigschwardt (approved):
Your part looks good!

ludwigschwardt:
Now please have a look at the new bits :-D

katdal/h5datav3.py
```diff
@@ -169,7 +170,9 @@ class H5DataV3(DataSet):
     def __init__(self, filename, ref_ant='', time_offset=0.0, mode='r',
                  time_scale=None, time_origin=None, rotate_bls=False,
                  centre_freq=None, band=None, keepdims=False, **kwargs):
-        DataSet.__init__(self, filename, ref_ant, time_offset)
+        # The closest thing to a capture block ID is the Unix timestamp in the original filename
+        cbid = pathlib.Path(filename).stem
+        DataSet.__init__(self, cbid, ref_ant, time_offset, url=filename)
```
cchristelis, line 175: With all this repetition it almost feels like we could have a H5Data super between DataSet and the H5DataV{n} classes. But since I do think we should have these classes deprecate at the next major version bump this is probably not worth doing...

cchristelis (commented):
I'm happy with the changes you've added @ludwigschwardt , I can't do anything but comment since I created this PR.

katdal/h5datav3.py
ludwigschwardt: Agreed, there is a lot of historical overlap.

ludwigschwardt:
Yes, it's sad that the PR originator can't also review things... Although it could probably be abused :-D

PR #338, opened 2021-06-19
SPR1-1138: Improve the exposure of CBF sensors

ludwigschwardt:
There is now the following set of virtual sensors that expose the underlying CBF sensors (`wide_antenna_channelised_voltage_{inp}_delay/eq`) in a more katdal-friendly way:
```
Correlator/Inputs/{inp}/applied_delay [seconds]
Correlator/Inputs/{inp}/applied_phase [radians]
Correlator/Inputs/{inp}/applied_gain [complex64, shape (n_chans,)]
```
The idea is to use them to undo delay tracking and/or F-engine gains.

@spassmoor I recall you did similar experiments at some stage, so I'm including you in the review mix.

PS the `applied_phase` sensor is there for completeness but I'm still figuring out the best way to use it in delay tracking inversion.

### DIFF
--- a/katdal/visdatav4.py
+++ b/katdal/visdatav4.py
@@ -27,8 +27,8 @@ from .dataset import (DataSet, BrokenFile, Subarray, DEFAULT_SENSOR_PROPS,
                       _selection_to_list)
 from .vis_flags_weights import VisFlagsWeights
 from .spectral_window import SpectralWindow
-from .sensordata import SensorCache
-from .categorical import CategoricalData
+from .sensordata import SensorCache, SimpleSensorGetter
+from .categorical import CategoricalData, ComparableArrayWrapper
 from .lazy_indexer import DaskLazyIndexer
 from .applycal import (add_applycal_sensors, calc_correction,
                        apply_vis_correction, apply_weights_correction,
@@ -76,9 +76,35 @@ def _calc_azel(cache, name, ant):
     return sensor_data
 
 
+def _calc_delay(cache, name, inp):
+    """Extract virtual delay correction sensors from custom-formatted CBF sensors."""
+    # Obtain the relevant CBF attributes from the cache
+    instrument = cache.get('Correlator/instrument')[0]
+    sync_time = cache.get('Correlator/sync_time')[0]
+    scale_factor_timestamp = cache.get('Correlator/scale_factor_timestamp')[0]
+    # Don't extract the sensor since the real timestamps are hidden in the values
+    getter = cache.get(f'{instrument}_antenna_channelised_voltage_{inp}_delay',
+                       extract=False)
+    sensor_data = getter.get()
+    # Unwrap and unpack the five components of the CBF sensor
+    values = [ComparableArrayWrapper.unwrap(v) for v in sensor_data.value]
+    adc_sample_counts, delays, delay_rates, phases, phase_rates = zip(*values)
+    # Convert the ADC sample count of each delay update to proper Unix timestamp
+    times = sync_time + np.array(adc_sample_counts) / scale_factor_timestamp
+    # Use the actual delay rate used by F-engine to interpolate between updates.
+    # Insert interpolation endpoint just before next update for strict monotonicity.
+    next_times = np.r_[times[1:], 2 * times[-1] - times[-2]] - 1e-6
+    next_delays = delays + delay_rates * (next_times - times)
+    times = np.c_[times, next_times].ravel()
+    delays = np.c_[delays, next_delays].ravel()
+    cache[name] = sensor_data = SimpleSensorGetter(name, times, delays)
+    return sensor_data
+
+
 VIRTUAL_SENSORS = dict(DEFAULT_VIRTUAL_SENSORS)
 VIRTUAL_SENSORS.update({'Antennas/{ant}/az': _calc_azel,
-                        'Antennas/{ant}/el': _calc_azel})
+                        'Antennas/{ant}/el': _calc_azel,
+                        'Correlator/Inputs/{inp}/applied_delay': _calc_delay})
 
 DEFAULT_CAL_PRODUCTS = ('l1.K', 'l1.B', 'l1.G', 'l2.GPHASE')
 
@@ -189,8 +215,10 @@ class VisibilityDataV4(DataSet):
             # XXX: should use telstate.join if attrs is a telstate
             self.cbf_dump_period = attrs[correlator_stream + '_int_time']
             cbf_n_accs = attrs[correlator_stream + '_n_accs']
+            self._cbf_instrument = attrs[correlator_stream + '_instrument_dev_name']
         except (KeyError, IndexError):
             self.cbf_dump_period = self.accumulations_per_dump = None
+            self._cbf_instrument = 'UNKNOWN'
         else:
             cbf_dumps_per_sdp_dump = round(self.dump_period / self.cbf_dump_period)
             self.accumulations_per_dump = cbf_n_accs * cbf_dumps_per_sdp_dump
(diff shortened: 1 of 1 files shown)


katdal/visdatav4.py
```diff
@@ -76,6 +76,57 @@ def _calc_azel(cache, name, ant):
...
+                       extract=False)
+    sensor_data = getter.get()
+    # Unwrap and unpack the five components of the CBF sensor
+    values = [ComparableArrayWrapper.unwrap(v) for v in sensor_data.value]
+    adc_sample_counts, delays, delay_rates, phases, phase_rates = zip(*values)
+    # Convert the ADC sample count of each delay update to proper Unix timestamp
+    times = sync_time + np.array(adc_sample_counts) / scale_factor_timestamp
+    # Use actual delay/phase rate used by F-engine to interpolate between updates.
+    # Insert interpolation endpoint just before next update for strict monotonicity.
+    next_times = np.r_[times[1:], 2 * times[-1] - times[-2]] - 1e-6
```
bmerry, line 96: This will raise IndexError if there are not at least two samples.

bmerry: Rather than extrapolating forward by an estimating update interval, should the last rate not be extrapolated forward to the final dump? I'm thinking about a case where something goes wrong and delay updates stop happening and the last delay rate gets applied indefinitely.

katdal/visdatav4.py
```diff
@@ -76,6 +76,57 @@ def _calc_azel(cache, name, ant):
...
+    cache[name.replace('applied_delay', 'applied_phase')] = phase_data
+    return delay_data if name.endswith('delay') else phase_data
+
+
+def _calc_gain(cache, name, inp):
+    """Extract virtual applied F-engine gain sensors from raw CBF sensors."""
+    instrument = cache.get('Correlator/instrument')[0]
+    # The real/imag parts are cast to int16 in the F-engine but the CBF sensor
+    # seems to report back the CAM request, so round them here to be more accurate
+    sensor_data = cache.get(f'{instrument}_antenna_channelised_voltage_{inp}_eq',
```
bmerry, line 114: Again, don't hardcode assumption about stream name.

katdal/visdatav4.py
```diff
@@ -76,6 +76,57 @@ def _calc_azel(cache, name, ant):
...
+                            transform=lambda g: np.array(g, dtype=np.complex64).round())
```
bmerry, line 115: - I don't know whether the transform applies to individual samples or to the whole timeseries, but could using `np.array` cause users to get back 0d arrays in place of scalars?
- While extremely unlikely, converting to complex64 then rounding could give a different result to directly rounding.

katdal/visdatav4.py
```diff
@@ -185,12 +229,14 @@ def _before(date):
...
+            correlator_stream = attrs['src_streams'][0]
             # XXX: should use telstate.join if attrs is a telstate
-            self.cbf_dump_period = attrs[src_stream + '_int_time']
-            cbf_n_accs = attrs[src_stream + '_n_accs']
+            self.cbf_dump_period = attrs[correlator_stream + '_int_time']
+            cbf_n_accs = attrs[correlator_stream + '_n_accs']
+            self._cbf_instrument = attrs[correlator_stream + '_instrument_dev_name']
         except (KeyError, IndexError):
             self.cbf_dump_period = self.accumulations_per_dump = None
+            self._cbf_instrument = 'UNKNOWN'
```
bmerry, line 239: Maybe None would be better (as you'd done for dump information).

katdal/visdatav4.py
```diff
@@ -301,6 +347,26 @@ def _before(date):
         _add_sensor_alias(self.sensor, 'Antennas/array/activity', 'obs_activity')
         _add_sensor_alias(self.sensor, 'Antennas/array/target', 'cbf_target')
 
+        # ------ Extract correlator settings ------
+
+        self.sensor['Correlator/instrument'] = CategoricalData(
+            [self._cbf_instrument], all_dumps)
```
bmerry, line 353: We shouldn't assume that all the relevant CBF streams come from the same instrument: this is the instrument for the baseline-correlation-products stream.

katdal/visdatav4.py
```diff
@@ -301,6 +347,26 @@ def _before(date):
...
+                [sync_time], all_dumps)
+        try:
+            scale_factor_timestamp = attrs[self._cbf_instrument
+                                           + '_scale_factor_timestamp']
+        except KeyError:
+            pass
+        else:
+            self.sensor['Correlator/scale_factor_timestamp'] = CategoricalData(
+               [scale_factor_timestamp], all_dumps)
+
```
bmerry, line 369: Seems a bit inconsistent that `Correlator/instrument` always exists but might be UNKNOWN while these other sensors will no exist at all if the information is not present.

bmerry (changes requested):

katdal/visdatav4.py
```diff
@@ -76,6 +76,57 @@ def _calc_azel(cache, name, ant):
...
 
+def _calc_delay(cache, name, inp):
+    """Extract virtual applied delay/phase sensors from raw CBF sensors."""
+    # Obtain the relevant CBF attributes from the cache
+    instrument = cache.get('Correlator/instrument')[0]
+    sync_time = cache.get('Correlator/sync_time')[0]
+    scale_factor_timestamp = cache.get('Correlator/scale_factor_timestamp')[0]
+    # Don't extract the sensor since the real timestamps are hidden in the values
+    getter = cache.get(f'{instrument}_antenna_channelised_voltage_{inp}_delay',
+                       extract=False)
```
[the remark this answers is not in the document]
ludwigschwardt: Yes, I cut some corners here... I'll go back to the drawing board :-D

katdal/visdatav4.py
ludwigschwardt: - I wilfully ignored the first issue, but I'll try to patch it up.
- The second suggestion is great, since that is exactly what will happen. Will do!

katdal/visdatav4.py
ludwigschwardt: - The transform applies to the individual values of a categorical sensor (had to look that up!). Contrary to popular (my) belief, it doesn't even apply to numerical sensors... However, the values in `*_eq` are lists of `n_chans` complex values, so they are (a) categorical and (b) not scalar. So the answer is no.
- This ordering happened because I first had only the complex64 cast and then later considered the rounding too. The sensor values are definitely low-resolution (i.e. 7-8 decimal places), so I don't think the cast will change anything. The real question is what CBF does inside CMC + corr2... Do they work with complex128 or complex64 before the cast to int16? Since you are now an insider, maybe you can tell me 😄

katdal/visdatav4.py
ludwigschwardt: I'm removing the need for this.

ludwigschwardt:
The new iteration is definitely better, and not as onerous as I had thought.

The technical debt is wracking up though :-D Better aliases, better integration with telstate, a shortcut for attribute-like sensors, etc.

katdal/visdatav4.py
```diff
@@ -79,21 +79,22 @@ def _calc_azel(cache, name, ant):
 def _calc_delay(cache, name, inp):
     """Extract virtual applied delay/phase sensors from raw CBF sensors."""
     # Obtain the relevant CBF attributes from the cache
-    instrument = cache.get('Correlator/instrument')[0]
+    stream = cache.get('Correlator/f_engine_stream')[0]
```
bmerry, line 82: Officially this is called "antenna channelised voltage" - not just in documentation, but also in CAM sensors, SDP-CAM ICD (for stream types), and so on. It's probably best to be consistent.

katdal/visdatav4.py
bmerry: > Since you are now an insider, maybe you can tell me

I'm only working on the next-gen correlator, so no insights into the current one. The case I'm thinking of is:
```python
> np.array(1.49999999+0j, np.complex64).round()
(2+0j)
> np.array(1.49999999+0j).round()
(1+0j)
```
I'm not sure it's worth worrying about though. If it becomes a problem we should fix the CBF sensors to report what is used.

bmerry (changes requested):
Looking good I think (one problem with doing 1 day a week is I forget details from one week to the next); just the naming of the sensor for the antenna-channelised-voltage stream.

katdal/visdatav4.py
ludwigschwardt: The real culprit is katcorelib (i.e. CAM or me :-)) that calls `?gain` with float real/imags while the ICD actually calls for integer real/imags. This is not helped by CBF generously accepting our invalid data and rounding it for us...

The upside is that we know what we would have liked it to be, and could correct for the quantisation effects (although that would be of little use in practice).

katdal/visdatav4.py
ludwigschwardt: I see your point. I've slipped up once already in katcorelib...

ludwigschwardt:
Ready for another check.

bmerry (approved):

PR #339, opened 2021-06-24
SPR1-1210: Fix mvftoms on macOS and Python 3.8+

ludwigschwardt:
Multiprocessing changed its default process start method on macOS to "spawn" in Python 3.8. This requires the parameters to `Process.__init__` to be picklable. One of them is an instance of the `CPInfo` namedtuple. This class is defined locally in the `corrprod_index` function, where pickle cannot find it.

Fix the issue by moving the definition of `CPInfo` to the module/script level.


spassmoor (approved):
looks ok , you are just moving CPInfo to be global.

PR #340, opened 2021-07-19
SPR1-1143: Allow deselection of antennas in katdal using '~m0XX'

cchristelis:
This change allows the deselection of a set of antenna by prefacing the antenna with ~

### DIFF
--- a/katdal/dataset.py
+++ b/katdal/dataset.py
@@ -168,6 +168,17 @@ def _selection_to_list(names, **groups):
         return [names]
 
 
+def _is_deselection(selectors):
+    """If all the selectors have a tilde ~ , then this is treated as a
+     deselect and we are going to invert the selection.
+     TODO: For version 1 release the deselector interface should just have a leading ~
+     """
+    for selector in selectors:
+        if selector[0] != '~':
+            return False
+    return True
+
+
 DEFAULT_SENSOR_PROPS = {
     '*nd_coupler': {'categorical': True, 'greedy_values': (True,), 'initial_value': '0',
                     'transform': lambda x: x not in ('0', 'False', 0)},
@@ -550,14 +561,6 @@ class DataSet:
                 model.min_freq_MHz = new_min_freq
                 model.max_freq_MHz = new_max_freq
 
-    def _is_deselection(self, selectors):
-        """If all the selectors have a tilde ~ , then this is treated as a
-         deselect and we are going to invert the selection."""
-        for selector in selectors:
-            if selector[0] != '~':
-                return False
-        return True
-
     def _set_keep(self, time_keep=None, freq_keep=None, corrprod_keep=None,
                   weights_keep=None, flags_keep=None):
         """Set time, frequency and/or correlation product selection masks.
@@ -832,7 +835,7 @@ class DataSet:
             elif k == 'ants':
                 ants = _selection_to_list(v)
                 ant_names = [(ant.name if isinstance(ant, katpoint.Antenna) else ant) for ant in ants]
-                if self._is_deselection(ant_names):
+                if _is_deselection(ant_names):
                     ant_names = [ant_name[1:] for ant_name in ant_names]
                     self._corrprod_keep &= [(inpA[:-1] not in ant_names and inpB[:-1] not in ant_names)
                                         for inpA, inpB in self.subarrays[self.subarray].corr_products]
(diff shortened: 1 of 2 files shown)


katdal/dataset.py
```diff
@@ -822,7 +832,12 @@ def select(self, **kwargs):
             elif k == 'ants':
                 ants = _selection_to_list(v)
                 ant_names = [(ant.name if isinstance(ant, katpoint.Antenna) else ant) for ant in ants]
-                self._corrprod_keep &= [(inpA[:-1] in ant_names and inpB[:-1] in ant_names)
+                if self._is_deselection(ant_names):
```
ludwigschwardt, line 835: I prefer this to a separate method (especially a static one):
```Python
is_deselection = set(name[0] for name in ant_names) == {'~'}
```

katdal/dataset.py
```diff
@@ -658,7 +666,9 @@ def select(self, **kwargs):
             value via a sequence of string pairs, or select all autocorrelations
             via 'auto' or all cross-correlations via 'cross'.
         ants : string or :class:`katpoint.Antenna` object or sequence, optional
-            Select antennas by name or object
+            Select antennas by name or object. If all antennas specified are
```
ludwigschwardt, line 669: I haven't considered the issue of mixing '~' with normal selections, which is kind of pointless :-)

Maybe we should consider just having a single tilde at the start to negate the entire selection?

ludwigschwardt (commented):
Some things to discuss...

katdal/dataset.py
cchristelis: Yeah, I think that is the only case that makes sense. It would be pointless to deselect some antenna and then select others... 
In terms of interface consistency, will we be able to treat _all_ the deselects in the same way i.e. `select(kwarg=selections, kwarg2=~deselections...)`

katdal/dataset.py
cchristelis: I do want us to use the same is deselection throughout the select method, so I do want it to exist as a function. 
If you feel it is too small for a method, then how about a function or even a lambda within the scope of this method?

ludwigschwardt: A function is fine (see e.g. `_selection_to_list`). My issue is that `self` is not involved so there is no need to have it on the class.

katdal/dataset.py
ludwigschwardt: Good question... The only deselection currently occurs in `scans` and `compscans`. I see that it also uses `_selection_to_list` and checks individual terms for a leading tilde. It therefore expects e.g. `'~slew,~stop'`.

But then it dutifully goes and ORs those two selections together... 🤦🏾 So it only really makes sense for one negative term at a time. We could also fix that to be a single leading tilde too.

Bear in mind that katdal 1.0 will happen sometime this year, which will provide the opportunity to fix any warts.

katdal/dataset.py
```diff
@@ -168,6 +168,17 @@ def _selection_to_list(names, **groups):
         return [names]
 
 
+def _is_deselection(selectors):
+    """If all the selectors have a tilde ~ , then this is treated as a
+     deselect and we are going to invert the selection.
+     TODO: For version 1 release the deselector interface should just have a leading ~
```
ludwigschwardt, line 174: Part of me says that we should fix this now, because users have not used multiple `~`s before and this PR will make them used to it. Their code will then break come katdal 1.0.

On the other hand, many other things will probably break with katdal 1.0 😄

ludwigschwardt (approved):
I'm tentatively approving this, but we need a ticket to remind us to fix the negation for all cases in katdal 1.0 then.

It's also useful to check what [the other guys did](https://casacore.github.io/casacore-notes/263.html). CASA has a single `!` in front of a group of antennas to invert the selection.

katdal/dataset.py
cchristelis: Yeah, I think targeting 1.0 will be the best, IIRC the other deselects are on the individual terms.

ludwigschwardt: That might be true, but it's also broken...

ludwigschwardt: That's my point: no-one has ever used multiple tildes, or they would have hit a lovely "and vs or" bug.

cchristelis:
I'm merging, the uniform negation will be followed up in SPR1-1323

PR #341, opened 2021-07-27
SPR1-1155: Select targets by their tags (first attempt)

cchristelis:
Allow users to select target by their tags.
e.g. 
```Python
d = katdal.open(url)
d.select(target_tags='gaincal')
```

### DIFF
--- a/katdal/dataset.py
+++ b/katdal/dataset.py
@@ -702,12 +702,13 @@ class DataSet:
             If `spw` or `subarray` is out of range
 
         """
-        time_selectors = ['dumps', 'timerange', 'scans', 'compscans', 'targets']
+        time_selectors = ['dumps', 'timerange', 'scans', 'compscans',
+                          'targets', 'target_tags']
         freq_selectors = ['channels', 'freqrange']
         corrprod_selectors = ['corrprods', 'ants', 'inputs', 'pol']
         # Check if keywords are valid and raise exception only if this is explicitly enabled
         valid_kwargs = time_selectors + freq_selectors + corrprod_selectors + \
-            ['spw', 'subarray', 'weights', 'flags', 'reset', 'strict', 'target_tags']
+            ['spw', 'subarray', 'weights', 'flags', 'reset', 'strict']
         # Check for definition of strict
         strict = kwargs.get('strict', True)
         if strict and set(kwargs.keys()) - set(valid_kwargs):
@@ -803,9 +804,10 @@ class DataSet:
                 tags = v if is_iterable(v) else [v]
                 target_keep = np.zeros(len(self._time_keep), dtype=np.bool)
                 target_index_sensor = self.sensor.get('Observation/target_index')
-                for target_index in self.target_indices:
-                    target_tags = self.catalogue.targets[target_index].tags
+                for target in self.catalogue.targets:
+                    target_tags = target.tags
                     if set(target_tags) & set(tags):
+                        target_index = self.catalogue.targets.index(target)
                         target_keep |= (target_index_sensor == target_index)
                 self._time_keep &= target_keep
             # Selections that affect frequency axis
--- a/katdal/dataset.py
+++ b/katdal/dataset.py
@@ -694,7 +694,7 @@ class DataSet:
         corrprod_selectors = ['corrprods', 'ants', 'inputs', 'pol']
         # Check if keywords are valid and raise exception only if this is explicitly enabled
         valid_kwargs = time_selectors + freq_selectors + corrprod_selectors + \
-            ['spw', 'subarray', 'weights', 'flags', 'reset', 'strict']
+            ['spw', 'subarray', 'weights', 'flags', 'reset', 'strict', 'target_tags']
         # Check for definition of strict
         strict = kwargs.get('strict', True)
         if strict and set(kwargs.keys()) - set(valid_kwargs):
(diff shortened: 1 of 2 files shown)


katdal/test/test_dataset.py
```diff
@@ -164,3 +165,9 @@ def test_uvw(self):
         assert_array_equal(self.dataset.u[:, 5], u)
         assert_array_equal(self.dataset.v[:, 5], v)
         assert_array_equal(self.dataset.w[:, 5], w)
+
+    def test_tag_selection(self):
+        self.dataset.select(target_tags='radec')
+        assert_equal(len(self.dataset.target_indices), 1)
```
ludwigschwardt, line 171: You should be able to compare the actual value of `target_indices`, not just the length.

katdal/dataset.py
```diff
@@ -786,6 +786,15 @@ def select(self, **kwargs):
                         continue
                     target_keep |= (target_index_sensor == target_index)
                 self._time_keep &= target_keep
+            elif k == 'target_tags':
+                tags = v if is_iterable(v) else [v]
+                target_keep = np.zeros(len(self._time_keep), dtype=np.bool)
+                target_index_sensor = self.sensor.get('Observation/target_index')
+                for target_index in self.target_indices:
```
ludwigschwardt, line 793: I would not use `self.target_indices` inside `select`, since that attribute is selection-dependent and you will in effect be applying a selection on top of another selection. Rather work with `self.catalogue`, which is selection-independent.

ludwigschwardt (changes requested):
Consider `katpoint.Catalogue.filter`, which selects on tags, and even allows negation 😄 

For the record, that negation is done per item in the list, so maybe worth revisiting my previous stance on #340...

katdal/dataset.py
```diff
@@ -694,7 +694,7 @@ def select(self, **kwargs):
         corrprod_selectors = ['corrprods', 'ants', 'inputs', 'pol']
         # Check if keywords are valid and raise exception only if this is explicitly enabled
         valid_kwargs = time_selectors + freq_selectors + corrprod_selectors + \
-            ['spw', 'subarray', 'weights', 'flags', 'reset', 'strict']
```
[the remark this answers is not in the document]
cchristelis: Learning by doing, thanks @ludwigschwardt !

katdal/dataset.py
cchristelis: Good point, this would have caused a bug on second select...

katdal/test/test_dataset.py
```diff
@@ -114,7 +114,8 @@ def test_selection_to_list():
 
 class TestVirtualSensors:
     def setup(self):
-        self.target = Target('PKS1934-638, radec, 19:39, -63:42')
+        self.target = Target('PKS1934-638, radec, 19:39, -63:42',
+                             tags='radec')
```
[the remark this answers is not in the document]
cchristelis: :+1:

katdal/dataset.py
```diff
@@ -799,6 +800,16 @@ def select(self, **kwargs):
                         continue
                     target_keep |= (target_index_sensor == target_index)
                 self._time_keep &= target_keep
+            elif k == 'target_tags':
+                tags = v if is_iterable(v) else [v]
+                target_keep = np.zeros(len(self._time_keep), dtype=np.bool)
+                target_index_sensor = self.sensor.get('Observation/target_index')
+                for target in self.catalogue.targets:
+                    target_tags = target.tags
```
ludwigschwardt, line 808: This line is superfluous. Rather incorporate it in the next line.

katdal/dataset.py
```diff
@@ -799,6 +800,16 @@ def select(self, **kwargs):
...
+                    if set(target_tags) & set(tags):
+                        target_index = self.catalogue.targets.index(target)
```
ludwigschwardt, line 810: Rather use `enumerate` on the for loop instead of looking up the target again.

katdal/dataset.py
```diff
@@ -799,6 +800,16 @@ def select(self, **kwargs):
...
+                        target_keep |= (target_index_sensor == target_index)
+                self._time_keep &= target_keep
```
ludwigschwardt, line 812: It's worth collecting all the tags in the catalogue in a set and highlighting any unknown strings in `tags` at the end to help the user debug the selection, using `logger.warning` like in the `targets` selection.

katdal/test/test_dataset.py
```diff
@@ -165,6 +165,12 @@ def test_uvw(self):
         assert_array_equal(self.dataset.v[:, 5], v)
         assert_array_equal(self.dataset.w[:, 5], w)
 
+    def test_tag_selection(self):
+        self.dataset.select(target_tags='radec')
+        assert_equal(self.dataset.target_indices, [0])
+        self.dataset.select(target_tags='incorrect_tag')
+        assert_equal(self.dataset.target_indices, [])
```
ludwigschwardt, line 172: You could also check that a warning was logged here.

ludwigschwardt (changes requested):
There are many more things to try in combination with the code in `Catalogue`, but let's do that later :-)

ludwigschwardt:
Also, a general tip: https://chris.beams.io/posts/git-commit/ 🙂

ludwigschwardt:
Any progress on this? :-D I'm making a release, but I guess this could go into the next one.

katdal/dataset.py
ludwigschwardt: Done.

katdal/dataset.py
ludwigschwardt: Done.

katdal/dataset.py
ludwigschwardt: Done.

katdal/test/test_dataset.py
ludwigschwardt: Done.

ludwigschwardt:
Taken up in #377 again.

Issue #315, opened 2020-12-09
issue error message when dataset is not on Ceph


o-smirnov:
Current behaviour is to return zero data, which is a handy way to mislead users into using up their disk quota!

ludwigschwardt:
Closed by #344.

PR #344, opened 2021-08-09
SPR1-766: Verify that S3 bucket is not empty

ludwigschwardt:
Whenever a chunk is not found, go and check the associated bucket. If the bucket is missing or empty, raise `StoreUnavailable` instead since this is a fatal error. In the past katdal would happily return a bucketload of zeros with no complaints.

The bucket exists but is empty if the dataset has been archived to tape. A completely missing bucket is unlikely, but check for it anyway.

~~We perform the check with every missing chunk because that is simpler than tracking which buckets we verified. The overhead should be low, since we typically either have all chunks missing (the aim of this fix) or only a handful of chunks missing. Besides that, 404s are quite quick compared to the downloading of actual chunks.~~

We cache all verified buckets on the store, assuming that a bucket won't empty while we are accessing it with katdal.

### DIFF
--- a/katdal/chunkstore_s3.py
+++ b/katdal/chunkstore_s3.py
@@ -611,7 +611,9 @@ class S3ChunkStore(ChunkStore):
         """Check that bucket associated with `url` exists and is not empty."""
         bucket = _bucket_url(url)
         try:
-            response = self.complete_request('GET', bucket, process=lambda r: r)
+            # Speed up the request by only checking that the bucket has at least one key
+            response = self.complete_request('GET', bucket, process=lambda r: r,
+                                             params={'max-keys': 1})
         except S3ObjectNotFound as err:
             # There is no point continuing if the bucket is completely missing
             raise StoreUnavailable(err) from chunk_error
--- a/katdal/chunkstore_s3.py
+++ b/katdal/chunkstore_s3.py
@@ -607,6 +607,19 @@ class S3ChunkStore(ChunkStore):
             else:
                 return result
 
+    def _verify_bucket(self, url, chunk_error=None):
+        """Check that bucket associated with `url` exists and is not empty."""
+        bucket = _bucket_url(url)
+        try:
+            response = self.complete_request('GET', bucket, process=lambda r: r)
+        except S3ObjectNotFound as err:
+            # There is no point continuing if the bucket is completely missing
+            raise StoreUnavailable(err) from chunk_error
+        # An empty bucket response has no Contents elements (no need for full XML parsing)
+        if response.ok and b'<Contents>' not in response.content:
+            msg = f'S3 bucket {bucket} is empty - your data is not currently accessible'
+            raise StoreUnavailable(msg) from chunk_error
+
     def get_chunk(self, array_name, slices, dtype):
         """See the docstring of :meth:`ChunkStore.get_chunk`."""
         dtype = np.dtype(dtype)
@@ -615,8 +628,14 @@ class S3ChunkStore(ChunkStore):
         # Our hacky optimisation to speed up response reading doesn't
         # work with non-identity encodings.
         headers = {'Accept-Encoding': 'identity'}
-        chunk = self.complete_request('GET', url, chunk_name, _read_chunk,
-                                      headers=headers, stream=True)
+        try:
+            chunk = self.complete_request('GET', url, chunk_name, _read_chunk,
+                                          headers=headers, stream=True)
+        except S3ObjectNotFound as err:
+            # If the entire bucket is gone, this becomes StoreUnavailable instead
+            self._verify_bucket(url, err)
+            # If the bucket checks out, treat this as a random missing chunk and reraise
+            raise
         if chunk.shape != shape or chunk.dtype != dtype:
             raise BadChunk(f'Chunk {chunk_name!r}: dtype {chunk.dtype} and/or shape {chunk.shape} '
                            f'in store differs from expected dtype {dtype} and shape {shape}')
--- a/katdal/chunkstore_s3.py
+++ b/katdal/chunkstore_s3.py
@@ -476,6 +476,7 @@ class S3ChunkStore(ChunkStore):
         self._session_pool = _Pool(session_factory)
         self._url = to_str(url)
         self._retries = retries
+        self._verified_buckets = set()
         self.timeout = timeout
         self.public_read = public_read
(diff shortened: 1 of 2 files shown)


katdal/chunkstore_s3.py
```diff
@@ -598,6 +607,19 @@ def complete_request(self, method, url, chunk_name='',
             else:
                 return result
 
+    def _verify_bucket(self, url, chunk_error=None):
+        """Check that bucket associated with `url` exists and is not empty."""
+        bucket = _bucket_url(url)
+        try:
+            response = self.complete_request('GET', bucket, process=lambda r: r)
```
bmerry, line 614: If you pass `max-keys=1` as a query parameter it'll limit the size of the response when the bucket is non-empty.

katdal/chunkstore_s3.py
```diff
@@ -598,6 +607,19 @@ def complete_request(self, method, url, chunk_name='',
...
+        except S3ObjectNotFound as err:
+            # There is no point continuing if the bucket is completely missing
+            raise StoreUnavailable(err) from chunk_error
+        # An empty bucket response has no Contents elements (no need for full XML parsing)
+        if response.ok and b'<Contents>' not in response.content:
+            msg = f'S3 bucket {bucket} is empty - your data is not currently accessible'
```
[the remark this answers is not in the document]
ludwigschwardt: That's what I had originally:
```
S3 bucket <x> is empty. Your data is probably archived to tape - please ask support to restage it.
```
The archive folks did not like it though, since terms like "restage" and "tape" are considered internal implementation details to be abstracted away under ideal circumstances. I added the word "currently" to reduce user panic at the very least :-)

katdal/chunkstore_s3.py
ludwigschwardt: Good point, done.

katdal/chunkstore_s3.py
```diff
@@ -598,6 +607,19 @@ def complete_request(self, method, url, chunk_name='',
...
+    def _verify_bucket(self, url, chunk_error=None):
+        """Check that bucket associated with `url` exists and is not empty."""
+        bucket = _bucket_url(url)
+        try:
+            response = self.complete_request('GET', bucket, process=lambda r: r)
+        except S3ObjectNotFound as err:
+            # There is no point continuing if the bucket is completely missing
+            raise StoreUnavailable(err) from chunk_error
+        # An empty bucket response has no Contents elements (no need for full XML parsing)
+        if response.ok and b'<Contents>' not in response.content:
```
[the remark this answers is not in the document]
bmerry: In that case, should the test of `response.ok` be removed?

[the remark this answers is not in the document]
ludwigschwardt: I think so too. Done.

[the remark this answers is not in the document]
bmerry: I don't see the change yet...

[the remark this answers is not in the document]
ludwigschwardt: There now... I just saw your comment too late for the previous round.

bmerry (approved):

ludwigschwardt:
@bruce, I've considered your concerns about the expense of listing the bucket repeatedly, and decided to add a cache instead of figuring out the performance hit :-)

This cache is not too bad, since once a bucket is non-empty, it generally remains that way for the immediate future. Unless someone is archiving it to tape at that very moment :joy:, which can be managed with the tokens.

katdal/test/test_chunkstore_s3.py
```diff
@@ -342,6 +344,12 @@ def test_missing_or_empty_buckets(self):
         # Without put_chunk the bucket is empty
         with assert_raises(StoreUnavailable):
             self.store.get_chunk(f'{BUCKET}-empty/x', slices, dtype)
+        # Check that the standard bucket has not been verified yet
+        bucket_url = urllib.parse.urljoin(self.store._url, BUCKET)
+        assert_true(bucket_url not in self.store._verified_buckets)
```
bmerry, line 349:
```suggestion
        assert_not_in(bucket_url, self.store._verified_buckets)
```

katdal/test/test_chunkstore_s3.py
```diff
@@ -342,6 +344,12 @@ def test_missing_or_empty_buckets(self):
...
+        # Check that the standard bucket remains verified after initial check
+        self.test_chunk_non_existent()
+        assert_true(bucket_url in self.store._verified_buckets)
```
bmerry, line 352:
```suggestion
        assert_in(bucket_url, self.store._verified_buckets)
```

bmerry (changes requested):
Looks good. A minor suggestion on the test, but I'm happy for you to merge without waiting another week for me.

katdal/chunkstore_s3.py
```diff
@@ -168,6 +168,15 @@ def _read_chunk(response):
     return chunk
 
 
+def _bucket_url(url):
```
cchristelis, line 171: between the various repos we are repeating a lot of these kind of functions. It might be nice to pull these into a central repo at some point.

cchristelis (approved):
LGTM

katdal/chunkstore_s3.py
ludwigschwardt: As long as there is no danger of falling into the trap of https://github.com/ska-sa/katmisc, which has turned me off common utility libraries for a while. :-D

katdal/test/test_chunkstore_s3.py
ludwigschwardt: Done, thanks! I did not know about `assert_in`.

katdal/test/test_chunkstore_s3.py
ludwigschwardt: Done.

katdal/chunkstore_s3.py
cchristelis: haha, yeah, anything with the name misc is bound to cause confusion. Something along the lines of S3 utilities is probably more specific and is the kind of thing that you don't want to re-develop in each library...

ludwigschwardt: I count about 100-140 lines in katdal that qualifies as "S3 utility" (tests excluded). It will have to get more substantial to justify an entire package but it's good to keep an eye on it.

ludwigschwardt:
This also closed #315.

PR #348, opened 2021-09-29
SPR1-1312: Continue MS conversion on interrupt

cchristelis:
When a MVF to MS conversion happens the connection might drop which leads to a half written MS file. This change
adds functionality to detect how many rows have been written to the MS file and continues after that number of frames.

### DIFF
--- a/katdal/ms_extra.py
+++ b/katdal/ms_extra.py
@@ -42,6 +42,12 @@ def open_table(name, readonly=False, verbose=False, **kwargs):
     return tables.table(name, readonly=readonly, ack=verbose, **kwargs)
 
 
+def table_rows(name, readonly=True, verbose=False, **kwargs):
+    """Open casacore table and count how many dumps have been completed"""
+    with tables.table(name, readonly=readonly, ack=verbose, **kwargs) as t:
+        return t.nrows()
+
+
 def create_ms(filename, table_desc=None, dm_info=None):
     """Create an empty MS with the default expected sub-tables and columns."""
     with tables.default_ms(filename, table_desc, dm_info) as main_table:
--- a/katdal/ms_extra.py
+++ b/katdal/ms_extra.py
@@ -42,12 +42,6 @@ def open_table(name, readonly=False, verbose=False, **kwargs):
     return tables.table(name, readonly=readonly, ack=verbose, **kwargs)
 
 
-def table_rows(name, readonly=True, verbose=False, **kwargs):
-    """Open casacore table and count how many dumps have been completed"""
-    with tables.table(name, readonly=readonly, ack=verbose, **kwargs) as t:
-        return t.nrows()
-
-
 def create_ms(filename, table_desc=None, dm_info=None):
     """Create an empty MS with the default expected sub-tables and columns."""
     with tables.default_ms(filename, table_desc, dm_info) as main_table:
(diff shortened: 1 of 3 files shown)


katdal/ms_extra.py
```diff
@@ -42,6 +42,12 @@ def open_table(name, readonly=False, verbose=False, **kwargs):
     return tables.table(name, readonly=readonly, ack=verbose, **kwargs)
 
 
+def table_rows(name, readonly=True, verbose=False, **kwargs):
```
ludwigschwardt, line 45: I'm not a fan of a two-line function that is only ever used once (adding three more lines and an indirection to a different module for little gain). I would be happy to reinstate this once the second user surfaces.

There is probably a need to do further checks on the table, so then we will need it open in the calling code.

scripts/mvftoms.py
```diff
@@ -428,11 +428,6 @@ def _cp_index(a1, a2, pol):
                        dumps=slice(options.quack, None))
 
         # The first step is to copy the blank template MS to our desired output
```
ludwigschwardt, line 430: The blank template MS was last used in 2017 (and last seen in 2018), so this is an ancient wart indeed :joy: Please chop this too.

scripts/mvftoms.py
```diff
@@ -545,31 +540,38 @@ def _cp_index(a1, a2, pol):
         field_names, field_centers, field_times = [], [], []
         obs_modes = ['UNKNOWN']
         total_size = 0
-
-        # Create the MeasurementSet
-        table_desc, dminfo = ms_extra.kat_ms_desc_and_dminfo(
+        dump_start = 0
+        if os.path.exists(ms_name):
+            print(f"MS '{ms_name}' already exists - continuing...")
+            nbaselines = dataset.shape[2]/4
```
ludwigschwardt, line 546: There is already a variable called `nbl` just a few lines back that will also be more accurate (since you might have a single polarisation).

scripts/mvftoms.py
```diff
@@ -545,31 +540,38 @@ def _cp_index(a1, a2, pol):
...
+            nrows = ms_extra.table_rows(ms_name, verbose=options.verbose)
+            dump_start = int(nrows // nbaselines)
```
ludwigschwardt, line 548: `nbl` had better divide exactly into `nrows`... So no need for the `int` cast (I realise it's there because `nbaselines` was `float`).

I would also do some more spot checks of the existing MS, like the actual channels and baselines involved. It is possible that the second `mvftoms` run uses different script options that will result in an incompatible MS even though the MVF part (and CBID) is identical.

Or we can store these script options inside the MS somehow - that would be cool 🙂 The HISTORY sub-table might be the right place for this.

scripts/mvftoms.py
```diff
@@ -645,11 +647,16 @@ def _cp_index(a1, a2, pol):
                 # Iterate over time in some multiple of dump average
                 ntime = utc_seconds.size
                 ntime_av = 0
+                skipped = 0
 
                 for ltime in range(0, ntime - tsize + 1, tsize):
```
ludwigschwardt, line 652: Why not just start at `dump_start * tsize` instead of 0? The `skipped` variable also seems superfluous.

ludwigschwardt:
This is ready to be checked again. I've tested this on a local and remote dataset and the MS conversion seems to recover from an interruption. Two outstanding issues (for another PR, hopefully):
- Testing (MS routines have exactly 0% coverage...)
- Verification that we are continuing with the same dataset as before

ludwigschwardt (approved):
These changes were addressed last year. The new feature seems to work (with some warts) so we forge ahead for now.

PR #351, opened 2022-03-17
SPR1-1944: dask 2022.01.1 breaks katdal due to missing da.core.getem

ludwigschwardt:
The latest dask has renamed `da.core.getem` to `da.core.graph_from_arraylike` with a different interface (see dask/dask#7417). Revisit the `ChunkStore.{get|put}_dask_array` machinery to replace `getem`.

~~For the "get" case we stick with `getem` in older dasks, and switch to `graph_from_arraylike` in newer ones.~~
For the "get" case we revert to `da.from_array` with a custom arraylike object and getter function.

For the "put" case we switch to `da.map_blocks` all around (but we would ideally like to use `blockwise` in the future).

More rationale follows 😄

### DIFF
--- a/katdal/chunkstore.py
+++ b/katdal/chunkstore.py
@@ -19,12 +19,8 @@
 import contextlib
 import functools
 import io
-import uuid
 
 import dask.array as da
-from dask.highlevelgraph import HighLevelGraph
-from dask.blockwise import blockwise as core_blockwise
-from dask.layers import ArraySliceDep
 import numpy as np
 
 
@@ -166,21 +162,17 @@ def _add_offset_to_slices(func, offset):
     return func_with_offset
 
 
-def _scalar_to_chunk(func):
-    """Modify chunk get/put/has to turn a scalar return value into a chunk.
-
-    This modifies the given function so that it returns its result as an
-    ndarray with the same number of (singleton) dimensions as the corresponding
-    chunk to enable assembly into a dask array.
-    """
-
-    def func_returning_chunk(array_name, slices, *args, **kwargs):
-        """Turn scalar return value into chunk of appropriate dimension."""
-        value = func(array_name, slices, *args, **kwargs)
-        singleton_shape = len(slices) * (1,)
-        return np.full(singleton_shape, value)
-
-    return func_returning_chunk
+def _put_map_blocks(chunk, block_info=None, store=None, array_name=None, offset=()):
+    """Adapt `put` to `map_blocks` interface."""
+    put = store.put_chunk_noraise
+    if offset:
+        put = _add_offset_to_slices(put, offset)
+    # Turn block info into slices specification
+    slices = tuple(slice(*loc) for loc in block_info[0]["array-location"])
+    success = put(array_name, slices, chunk)
+    # Turn scalar output into ndarray with same ndims as chunk
+    singleton_shape = chunk.ndim * (1,)
+    return np.full(singleton_shape, success)
 
 
 def npy_header_and_body(chunk):
@@ -570,27 +562,13 @@ class ChunkStore:
             Dask array of objects indicating success of transfer of each chunk
             (None indicates success, otherwise there is an exception object)
         """
-        # Make out_name unique to avoid clashes and caches
-        out_name = f'store-{array_name}-{offset}-{uuid.uuid4().hex}'
-        put = _scalar_to_chunk(self.put_chunk_noraise)
-        if offset:
-            put = _add_offset_to_slices(put, offset)
-        out_ind = tuple(range(array.ndim))
-        layer = core_blockwise(
-            # Function signature: put(array_name, slices, chunk)
-            put,
-            # Output: we want all output chunks
-            out_name, out_ind,
-            # Arg 1: the same array name is passed to all tasks, so no out_ind
-            array_name, None,
-            # Arg 2: produce slice spec per chunk on the fly
-            ArraySliceDep(array.chunks), out_ind,
-            # Arg 3: this represents a chunk of `array`
-            array.name, out_ind,
-            # We need to tell blockwise about the external array's shape
-            numblocks={array.name: array.numblocks},
+        return da.map_blocks(
+            _put_map_blocks,
+            array,
+            name=f'store-{array_name}-{offset}',
+            dtype=object,
+            chunks=array.ndim * (1,),
+            store=self,
+            array_name=array_name,
+            offset=offset,
         )
-        dask_graph = HighLevelGraph.from_collections(out_name, layer, [array])
-        # The success array has one element per chunk in the input array
-        out_chunks = tuple(len(c) * (1,) for c in array.chunks)
-        return da.Array(dask_graph, out_name, out_chunks, object)
--- a/katdal/chunkstore.py
+++ b/katdal/chunkstore.py
@@ -149,6 +149,19 @@ def _add_offset_to_slices(func, offset):
     return func_with_offset
 
 
+def _put_map_blocks(chunk, block_info=None, store=None, array_name=None, offset=()):
+    """Adapt `put` to `map_blocks` interface."""
+    put = store.put_chunk_noraise
+    if offset:
+        put = _add_offset_to_slices(put, offset)
+    # Turn block info into slices specification
+    slices = tuple(slice(*loc) for loc in block_info[0]["array-location"])
+    success = put(array_name, slices, chunk)
+    # Turn scalar output into ndarray with same ndims as chunk
+    singleton_shape = chunk.ndim * (1,)
+    return np.full(singleton_shape, success)
+
+
 def npy_header_and_body(chunk):
     """Prepare a chunk for low-level writing.
 
@@ -533,23 +546,13 @@ class ChunkStore:
             Dask array of objects indicating success of transfer of each chunk
             (None indicates success, otherwise there is an exception object)
         """
-        put = self.put_chunk_noraise
-        if offset:
-            put = _add_offset_to_slices(put, offset)
-
-        def put_map_blocks(chunk, block_info=None):
-            """Adapt `put` to `map_blocks` interface."""
-            # Turn block info into slices specification
-            slices = tuple(slice(*loc) for loc in block_info[0]["array-location"])
-            success = put(array_name, slices, chunk)
-            # Turn scalar output into ndarray with same ndims as chunk
-            singleton_shape = chunk.ndim * (1,)
-            return np.full(singleton_shape, success)
-
         return da.map_blocks(
-            put_map_blocks,
+            _put_map_blocks,
             array,
             name=f'store-{array_name}-{offset}',
             dtype=object,
             chunks=array.ndim * (1,),
+            store=self,
+            array_name=array_name,
+            offset=offset,
         )
--- a/katdal/chunkstore.py
+++ b/katdal/chunkstore.py
@@ -132,19 +132,28 @@ def generate_chunks(shape, dtype, max_chunk_size, dims_to_split=None,
     return da.core.blockdims_from_blockshape(shape, dim_elements)
 
 
-def _graph_from_arraylike(array_name, chunks, getter):
-    """Create dask graph from chunk spec like the older :func:`da.core.getem`."""
+def _graph_from_chunks(array_name, chunks, getter, out_name):
+    """Create dask graph from chunk spec like the older :func:`da.core.getem`.
+
+    The basic graph structure (for a 1-D array) is:
+
+      {
+        (out_name, 0): (getter, array_name, (slice(start_0, end_0),)),
+        (out_name, 1): (getter, array_name, (slice(start_1, end_1),)),
+        ...
+      }
+    """
     try:
         return da.core.graph_from_arraylike(
             array_name,
             chunks,
             shape=tuple(sum(c) for c in chunks),
-            name=array_name,
+            name=out_name,
             getitem=getter,
             inline_array=True,
         )
     except AttributeError:
-        return da.core.getem(array_name, chunks, getter)
+        return da.core.getem(array_name, chunks, getter, out_name=out_name)
 
 
 def _add_offset_to_slices(func, offset):
@@ -535,9 +544,12 @@ class ChunkStore:
 
         if any(offset):
             getter = _add_offset_to_slices(getter, offset)
+        # The final name must differ from array_name, so pick deterministic hash
+        token = da.core.tokenize(self, chunks, dtype, index)
+        out_name = f'{array_name}-{offset}-{token}'
         # Use dask utility function that forms the core of da.from_array
-        dask_graph = _graph_from_arraylike(array_name, chunks, getter)
-        array = da.Array(dask_graph, array_name, chunks, dtype)
+        dask_graph = _graph_from_chunks(array_name, chunks, getter, out_name)
+        array = da.Array(dask_graph, out_name, chunks, dtype)
         return array[index]
 
     def put_dask_array(self, array_name, array, offset=()):
--- a/katdal/chunkstore.py
+++ b/katdal/chunkstore.py
@@ -17,7 +17,6 @@
 """Base class for accessing a store of chunks (i.e. N-dimensional arrays)."""
 
 import contextlib
-import functools
 import io
 
 import dask.array as da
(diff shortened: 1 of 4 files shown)


ludwigschwardt:
Why did we use `da.core.getem`?

It provides a convenient way to build a graph around a getter/putter and a chunk specification, in order to simulate an actual array on top of our zarr-like chunk store. We designed our getters and putters to accept dask-style array names and slice specifications as inputs, in order to have synergy with the underlying dask graphs.

There are two use cases: getting a dask array from the chunk store, and putting a dask array into the chunk store.

### 1. The “get” use case

`def ChunkStore.get_dask_array(array_name, chunks, dtype, ...):`

The chunk store interface relies on a `get_chunk(array_name, slices, dtype)` function to retrieve individual chunks. The dtype is partial’ed in, leaving `array_name` and `slices` as main parameters. For example, `getter('x', (slice(2, 4), slice(3, 6)))` will retrieve a 2-by-3 chunk of array `'x'` The name and slice specification are combined into a URL to be retrieved from the underlying object store behind the scenes.

This is very similar to `da.core.getter`, with the array `x` represented by its name instead of the actual array.

The relevant code in `get_dask_array` is:
```Python
graph = da.core.getem(array_name, chunks, getter)
array = da.Array(graph, array_name, chunks, dtype)
return array
```
It produces a basic graph like
```
# Called as: dx = store.get_dask_array('x', ((2, 2),), int)
# This is dx.dask.layers['x'].mapping
{
  ('x', 0): (getter, 'x', (slice(0, 2),)),
  ('x', 1): (getter, 'x', (slice(2, 4),)),
}
```
Note that `'x'` is overloaded to be the name of the dask array, high-level graph, materialised layer and part of the URL name for the custom getter...

This could be replaced by something like
```Python
class DummyArray:
    def __init__(self, chunks, dtype):
        self.shape = tuple(sum(c) for c in chunks)
        self.ndim = len(shape)
        self.dtype = np.dtype(dtype)
dummy_array = DummyArray(chunks, dtype)
array = da.from_array(dummy_array, chunks, array_name, getitem=getter)
```
The `dummy_array` will provide `.shape`, `.ndim` and more importantly `.dtype`. Since it won’t be of type `np.ndarray`, it will use the custom getter instead. (Hint: `namedtuple` is not good either since it is a subclass of `tuple`, which is converted to `np.ndarray` internally by `from_array`.)

Unfortunately its graph also contains the dummy array, which throws a spanner in the works:
```
{
  ('x', 0): (getter, 'x', (slice(0, 2),)),
  ('x', 1): (getter, 'x', (slice(2, 4),)),
  'x': dummy_array,
}
```
The custom getter will now be called on the dummy array, instead of the expected array name.

Can you just pop this key from the underlying mapping? Probably not.

What about `HighLevelGraph.from_collection`? Give it a shot, following the example in the docstring:
```Python
layer = {
    (array_name, i): (getter, array_name, slices)
    for i, slices in enumerate(da.core.slices_from_chunks(chunks))
}
graph = HighLevelGraph.from_collections(array_name, layer)
array = da.Array(graph, array_name, chunks, dtype)
```
Unfortunately this only works for 1-D arrays. The index `i` needs to be multidimensional, and constructing the keys like in `getem`:
```Python
keys = product([out_name], *(range(len(bds)) for bds in chunks))
```
really begs for the restoration of that function or the newer `graph_from_arraylike`. Oh well, let’s create a shim to choose between the two APIs then…

The nice thing is that `graph_from_arraylike` works pretty much like `getem`. The `arr` parameter may be a placeholder string instead of an actual arraylike object, which is critical for us. It should really be called `graph_from_chunks`… 😂 We can use `inline_array=True` to avoid adding the array itself to the graph.

Here is our shim (we are keeping the simpler getem-based API):
```Python
def _graph_from_chunks(array_name, chunks, getter):
    """Create dask graph from chunk spec like the older :func:`da.core.getem`."""
    try:
        return da.core.graph_from_arraylike(
            array_name,
            chunks,
            shape=tuple(sum(c) for c in chunks),
            name=array_name,
            getitem=getter,
            inline_array=True,
        )
    except AttributeError:
        return da.core.getem(array_name, chunks, getter)
```

ian-r-rose:
Thanks for the ping @ludwigschwardt, this was a good read!

I have a question about your attempt to use `da.from_array()`: it seems like your primary objection is that, by default, `from_array` inserts a copy of your dummy array into the graph. But later down you use `inline_array=True` with your invocation of `graph_from_arraylike()`. Would it also work to thread `inline_array=True` into `da.from_array()`? I don't know enough about your custom getter to know whether that is a good idea, but this is exactly the type of scenario for which `inline_array` is intended: where the array-ish class is some lazy representation of a larger store somewhere else.

gjoseph92:
> The custom getter will now be called on the dummy array, instead of the expected array name.

> As before, we don’t want dask to substitute 'x' with its value in this case

It seems like in both get and put, you want the name of the array (like `'x'`) to be passed as a literal string argument to your function, as opposed to? in addition to? the object at key `'x'` being substituted in. I'm assuming you need that string-name in order to come up with a URL in your `getter`/`putter` function.

Doing those sorts of substitutions is kind of dask's whole job. Instead, I'd think about other ways to get that information into the function. There are two strategies that come to mind:
1. When you control the object at key `'x'` (it's a `DummyArray`), just give it a `name` field
2. When you don't control the object (it's an arbitrary NumPy array), use `functools.partial(putter, name=...)` to pre-pass-in the name value.

```python
class DummyArray:
    def __init__(self, chunks, dtype, name):
        self.name = name
        self.shape = tuple(sum(c) for c in chunks)
        self.ndim = len(shape)
        self.dtype = np.dtype(dtype)

    def __getitem__(self, *idx):
        return getter(self.name, idx)

    def __setitem__(self, *idx):
        return putter(self.name, idx)

dummy_array = DummyArray(chunks, dtype, array_name)
array = da.from_array(dummy_array, chunks, array_name, inline_array=False)
... # do stuff to array
output = DummyArray(result.chunks, result.dtype, "result-name")
da.store(result, output)  # or continue using blockwise if you need it to remain an array
```


> Even after the array is created, it is about 18% slower to transfer data with `map_blocks` compared to `getem`.

That's interesting, and surprising. I'd expect the `map_blocks` and `blockwise` graphs to be nearly identical, the only difference being that `map_blocks` inefficiently generates a materialized graph because https://github.com/dask/dask/pull/7686 still hasn't happened.

> it results in massive serialisations and pickles on the way to a tiny hash string, causing `put_dask_array` to take more than a minute to create the array

Also surprised by this. Yes, the `map_putter` function would be pickled once to tokenize it, but I don't see why that would add a full minute. Could be something else going on (again, the inefficient graph generation due to using `block_info` could be an issue).

ludwigschwardt:
Thanks for taking the time to read this and for the helpful comments! I hope it gave you a glimpse of how folks are (ab)using dask... 😁

>  It seems like your primary objection is that, by default, from_array inserts a copy of your dummy array into the graph.

My primary objection is that I cannot return to this idyllic two-liner 🙂 
```Python
graph = da.core.getem(array_name, chunks, getter)
array = da.Array(graph, array_name, chunks, dtype)
```
Ideally I want to avoid yet another dummy class layer, unless it has a long-term benefit of greater maintainability. That said, I completely missed `inline_array` on `da.from_array` in all my meandering... 🤦🏼‍♂️ It's even got a great docstring! 

> It seems like in both get and put, you want the name of the array (like 'x') to be passed as a literal string argument to your function, as opposed to? in addition to? the object at key 'x' being substituted in.

As opposed to. It just happened to work in the old regime to have the same string for the object name and the getter argument, but now I need to have a different `out_name` anyway.

I basically made the getter function signature compatible with a dask slicing task and then leveraged `getem` to set up the calls.

> I'm assuming you need that string-name in order to come up with a URL in your getter/putter function.

Yes.

> There are two strategies that come to mind...

Thanks a lot, I'll give `DummyArray` another shot (both with or without `inline_array`).

> That's interesting, and surprising. I'd expect the map_blocks and blockwise graphs to be nearly identical.

I first thought it was due to the extra contortions I added to support `map_blocks` but it's not that. Maybe the `block_info` machinery? Here's a stripped-down example where the put call does nothing:
```Python
import dask.array as da
import numpy as np

dx = da.zeros(1000000, chunks=1000)

def _dummy_put(slices, chunk):
    return np.atleast_1d(None)

# OPTION 1: map_blocks

def _put_map_blocks(chunk, block_info=None):
    slices = tuple(slice(*loc) for loc in block_info[0]["array-location"])
    return _dummy_put(slices, chunk)

px = da.map_blocks(_put_map_blocks, dx, dtype=object)
%timeit px.compute()   # 148 ms

# OPTION 2: blockwise

from dask.layers import ArraySliceDep
from dask.highlevelgraph import HighLevelGraph
from dask.blockwise import blockwise as core_blockwise
out_name = "y"
out_ind = (0,)

layer = core_blockwise(
    _dummy_put,
    # Function signature: success = _dummy_put(slices, chunk)
    out_name, out_ind,
    # Parameter 1: slices
    ArraySliceDep(dx.chunks), out_ind,
    # Parameter 2: chunk
    dx.name, out_ind,
    numblocks={dx.name: dx.numblocks},
)
dask_graph = HighLevelGraph.from_collections(out_name, layer, [dx])
out_chunks = tuple(len(c) * (1,) for c in dx.chunks)
py = da.Array(dask_graph, out_name, out_chunks, object)
%timeit py.compute()   # 112 ms
```
I get a consistent 25% improvement with blockwise, until the chunks become few and large enough.

> Also surprised by this. Yes, the map_putter function would be pickled once to tokenize it, but I don't see why that would add a full minute.

Have a look at this:
```Python
In [14]: import numpy as np

In [15]: import dask.array as da

In [16]: class Store:
    ...:     def __init__(self, x):
    ...:         self.x = x
    ...: 
    ...:     def put(self, slices, chunk):
    ...:         self.x[slices] = chunk

In [17]: s = Store(np.zeros((100, 1000, 1000)))  # store 1 GB of data

In [18]: s.put
Out[18]: <bound method Store.put of <__main__.Store object at 0x121dea310>>

In [19]: %time da.core.tokenize(s.put)
CPU times: user 44.7 s, sys: 38.3 s, total: 1min 23s
Wall time: 1min 28s
Out[19]: 'affa99f676ec67eb789e6dbd2048a531'
```
[As an aside, I absent-mindedly copied and pasted the above code block as is into IPython... and it ran 🤯 😁 ]

Interrupting the process reveals that it calls `cloudpickle` and/or normal `pickle` on `s.put`, which seems to serialise the entire 1 GB store in the process. Not a big deal in practice, because we only use these kinds of in-memory stores in the unit tests. But the unit tests do suffer...

ludwigschwardt:
Thanks @gjoseph92 and @ian-r-rose for clearing up my mind!

I've gone full circle back to `da.from_array` for the "get" use case:
```Python
class _ArrayLikeGetter:
    """Present an array-like interface to chunk getter function."""
    def __init__(self, getter, array_name, chunks, dtype, **kwargs):
        self.getter = getter
        self.array_name = array_name
        self.kwargs = kwargs
        # Basic array-like attributes needed by :func:`dask.array.from_array`
        self.shape = tuple(sum(c) for c in chunks)
        self.ndim = len(self.shape)
        self.dtype = np.dtype(dtype)

    def __getitem__(self, slices):
        """Call chunk getter on slices set up by :func:`dask.array.from_array`."""
        return self.getter(self.array_name, slices, self.dtype, **self.kwargs)

getter_shim = _ArrayLikeGetter(getter, array_name, chunks, dtype, **kwargs)
return da.from_array(getter_shim, chunks, out_name, asarray=False)
```
The performance is similar, probably because we swapped a `functools.partial` (inserting the `dtype`) for a `__setitem__` wrapper function. Since the `array_name` does not pass through dask's paws, the need for `inline_array` has also disappeared. 🙂 

The “put” use case is still using `map_blocks` instead of using this dummy array class with `da.store` because `__setitem__` does not lend itself to returning errors.

Unless we patch `da.core.load_store_chunk` like this...
```Python
def load_store_chunk(x, out, index, lock, return_stored, load_stored, return_errors_of_type=UnusedException):
    result = None
    if return_stored and not load_stored:
        result = out

    if lock:
        lock.acquire()
    try:
        if x is not None:
            if is_arraylike(x):
                out[index] = x
            else:
                out[index] = np.asanyarray(x)
        if return_stored and load_stored:
            result = out[index]
    except return_errors_of_type as err:
        result = err
    finally:
        if lock:
            lock.release()

    return result
```
You pass in the base class of errors you want to return in `return_errors_of_type` (`ChunkStoreError` in our case). The above function then returns either `None` (i.e. success) or the exception object, without crashing the whole thing.

Just thinking out aloud 🙂

ludwigschwardt:
Hopefully the final issues:

The `_ArrayLikeGetter` scheme only works properly on dask 2022.03.0, and then only because of a regression (see dask/dask#8753).

When using the `index` parameter of `get_dask_array` to slice the array into a smaller one, the dask `fuse` graph optimisation step merges slices of slices into single slices, but this makes it incompatible with the chunk store, which expects very specific slices. That is, the chunk store dask array has a fixed set of chunks and is not arbitrarily sliceable as expected by `fuse`.
    
The solution is to specify a custom getter to `da.from_array`, even though it is the expected `_ArrayLikeGetter.__getitem__`. This prevents further optimisation of those getter tasks.
    
However, the plot thickens... The custom getter is not called by the latest dask 2022.04.0 unless you also set `inline_array=True`. This parameter first appeared in dask 2021.01.0, which also happens to be one of the last releases that still supports Python 3.6, so it's all good.

I now call `from_array` like this:
```Python
        getter_shim = _ArrayLikeGetter(getter, array_name, chunks, dtype, **kwargs)
        array = da.from_array(
            getter_shim,
            chunks,
            out_name,
            # Don't cast chunks to ndarray because some of them may be `PlaceholderChunk`s
            asarray=False,
            # Set custom getter anyway to prevent slice fusion during graph optimisation,
            # which ends up calling chunk store with wrong chunks.
            getitem=_ArrayLikeGetter.__getitem__,
            # Dask 2022.04.0 ignores the custom getter unless this is True
            inline_array=True
        )
        return array[index]
```

ludwigschwardt:
Ready for review again, @KimMcAlpine and @richarms 🙂

katdal/test/test_chunkstore.py
```diff
@@ -91,6 +91,20 @@ def test_max_dim_elements_ignore(self):
...
+    # The chunk start-stop boundaries on each axis are:
+    # ((0, 10, 20, 30, 40), (0, 2, 4, 6), (0, 40))
+    index = np.s_[13:34, :4, 10:]
+    new_chunks, new_index, new_offset = _prune_chunks(chunks, index)
+    # The new chunk start-stop boundaries on each axis are:
+    # ((10, 20, 30, 40), (0, 2, 4), (0, 40))
+    assert_equal(new_chunks, ((10, 10, 10), (2, 2), (40,)))
+    assert_equal(new_index, np.s_[3:24, 0:4, 10:40])
+    assert_equal(new_offset, (10, 0, 0))
+
```
KimMcAlpine, line 106: You could also add a test here for the case of a non-unit step for the slice.

KimMcAlpine (approved):
Looks good. Thanks for all the helpful write ups along the way :)

katdal/test/test_chunkstore.py
ludwigschwardt: Good idea, done!

katdal/chunkstore.py
```diff
@@ -131,6 +128,71 @@ def generate_chunks(shape, dtype, max_chunk_size, dims_to_split=None,
...
+            shape[axis] -= c
+        chunks[axis] = chunks[axis][start_chunk:stop_chunk]
+        if not chunks[axis]:
+            chunks[axis] = (0,)   # Dask doesn't allow empty chunk lists
+        index[axis] = slice(start, stop)
+    # Go back to tuples
+    chunks = tuple(tuple(c) for c in chunks)
+    index = tuple(index)
+    offset = tuple(offset)
+    return chunks, index, offset
```
richarms, line 193: Much more legible in a separate function, thank you.

richarms (approved):
Wow, that was quite a rollercoaster. Thanks @ludwigschwardt for all the helpful explanation along the way. Hopefully `from_array` and `map_blocks` will be with us in consistent form for many months to come!

ludwigschwardt:
At least the code now works on all dasks back to 1.2.1, which is a good sign for compatibility :-)

PR #352, opened 2022-04-13
SPR1-1401: Fix the handling of 403 errors

ludwigschwardt:
The original MeerKAT Ceph RADOS Gateway returned a 403 (Forbidden) for missing objects due to a bug in the interaction with the bucket policy. We therefore treated 403 as 404 and raised `ChunkNotFound`. See #273, 09e100d and 5a7790c for more details.

The current gateway has fixed this, and missing objects now return 404 as expected. Treat 403 the same as 401 and raise `AuthorisationFailed` instead.

I could not find any external user code that caught the old exception, so the update seems safe to roll out.


cchristelis (approved):
LGTM

PR #355, opened 2022-04-28
SPR1-1944: Fix bad Bad Chunk bug in DaskLazyIndexer optimisation

ludwigschwardt:
Previously, if the `slices` parameter found on many ChunkStore methods (aka the "chunk ID") was not a sequence of unit-stride `slice` objects with explicit `start` and `stop` attributes, a `BadChunk` was raised. The base class of `BadChunk` is `ValueError`, however, while an incorrect type for `slices` should be a `TypeError` instead.

This came to a head with the recent rework of the dask machinery in the chunk store, in quite a convoluted way... When a single dump is selected in `DaskLazyIndexer`, an extra step of `dask.optimization.cull` is performed. This tries to find some shortcuts during `__getitem__` which ends up calling `__getitem__(0)` on the new `_ArrayLikeGetter` class. Dask `cull` would have tried something else if this raised a `KeyError`, `IndexError` or `TypeError`, but `BadChunk` sails right through.

**Solution:** Turn an invalid `slices` parameter into a plain `TypeError` instead of a `BadChunk` to make `dask` happy and to clarify the purpose of `BadChunk`.

Add unit tests that perform `dataset.vis[0]` as faithfully as possible without invoking `VisibilityDataV4` itself (no unit tests for that yet...). It does pretty much everything at the L0 level (i.e. just not applycal). This should prevent being blindsided by similar dasky bugs in the future.

### DIFF
--- a/katdal/test/test_chunkstore.py
+++ b/katdal/test/test_chunkstore.py
@@ -117,11 +117,11 @@ class TestChunkStore:
     def test_metadata_validation(self):
         store = ChunkStore()
         # Bad slice specifications
-        assert_raises(BadChunk, store.chunk_metadata, "x", 3)
-        assert_raises(BadChunk, store.chunk_metadata, "x", [3, 2])
-        assert_raises(BadChunk, store.chunk_metadata, "x", slice(10))
-        assert_raises(BadChunk, store.chunk_metadata, "x", [slice(10)])
-        assert_raises(BadChunk, store.chunk_metadata, "x", [slice(0, 10, 2)])
+        assert_raises(TypeError, store.chunk_metadata, "x", 3)
+        assert_raises(TypeError, store.chunk_metadata, "x", [3, 2])
+        assert_raises(TypeError, store.chunk_metadata, "x", slice(10))
+        assert_raises(TypeError, store.chunk_metadata, "x", [slice(10)])
+        assert_raises(TypeError, store.chunk_metadata, "x", [slice(0, 10, 2)])
         # Chunk mismatch
         assert_raises(BadChunk, store.chunk_metadata, "x", [slice(0, 10, 1)],
                       chunk=np.ones(11))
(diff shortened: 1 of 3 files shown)


katdal/test/test_chunkstore.py
```diff
@@ -242,7 +242,7 @@ def test_chunk_zero_size(self):
     def test_put_chunk_noraise(self):
         name = self.array_name('x')
         self.store.create_array(name)
-        result = self.store.put_chunk_noraise(name, (1, 2), [])
+        result = self.store.put_chunk_noraise(name, (slice(1, 2),), np.ones(4))
```
richarms, line 245: If I have this correctly, previously `BadChunk` was being returned by `put_chunk_noraise` because the slice had no explicit start and stop attributes with which to verify. However, now `BadChunk` is being returned because of the `dtype` mismatch. And the previous call should result in a `TypeError` being returned. Correct?

ludwigschwardt:
> Looks good to me :)
> 
> So now just to check I understand. Now BadChunk (ValueError) is only raised when there is a chunk mismatch, which falls outside of dask's error handling mechanism and should percolate up to the calling code?

Yes! 🙂 

Dask was only probing the chunk store capabilities in an attempt to speed up access, so it expects failure but not of `ValueError` type (rather `KeyError` / `IndexError` / `TypeError`).

katdal/test/test_chunkstore.py
ludwigschwardt: Yes, except it's not a `dtype` mismatch, but a shape mismatch: `slice(1, 2)` => (1,) vs `np.ones(4)` => (4,). Put_chunk accepts the dtype of the given array (as long as it's not `object`).

This is an example of `BadChunk` abuse 😂  An invalid slice should never really be condoned, since it is a programmer error that must be fixed. It is even debatable if `BadChunk` should be ignored in the same way as `ChunkNotFound`. I just wanted a quick way to test `put_chunk_noraise`, which needs to raise a `ChunkStoreError` on demand.

I haven't really made a good case for `BadChunk`... It's basically something that should never happen :-)

PR #75, opened 2017-01-31
Use python-casacore's pending MeasurementSet creation functionality for creating Fixed Shape columns

sjperkins:
Previously, h5toms created variable length data columns when creating a MeasurementSet. Catering for this variability means the MS's are not performant. This PR 

1. Modifies h5toms to use python-casacore's upcoming functionality for creating canonical Measurement Set's. This removes the need to maintain table definitions for the MS within katdal.
2. Modfies h5toms to define fixed shape columns for `DATA`, `MODEL_DATA`, `CORRECTED_DATA`, `FLAG`, `FLAG_CATEGORY`, `IMAGING_WEIGHTS`, `WEIGHTS` and `SIGMA`, as well as the original hypercolumn definitions that were bases on these columns.

Depends on https://github.com/casacore/python-casacore/pull/72.

PR #356, opened 2022-07-12
SPR1-2367: Remove IMAGING_WEIGHT column from MS

ludwigschwardt:
This column is initialised full of zeroes and just takes up space (4 bytes per main table entry, i.e. time-frequency-baseline cell, so about 6-12% overhead). It can be recreated in CASA at a later stage.

The column is a leftover of good old `blank.ms`, which included it along with `MODEL_DATA` and `CORRECTED_DATA`. The latter two columns were made optional in the sneaky 41af4b8 as part of #75 in version 0.9, but this one was missed.

Testing an MS produced by the new code would be much appreciated.

### DIFF
--- a/katdal/ms_extra.py
+++ b/katdal/ms_extra.py
@@ -228,16 +228,6 @@ def kat_ms_desc_and_dminfo(nbl, nchan, ncorr, model_data=False):
     dmgroup_spec[dm_group] = dmspec(desc["desc"])
     additional_columns.append(desc)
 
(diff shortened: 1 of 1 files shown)


Jordatious (approved):

Jordatious:
Looks good! Thanks!

ludwigschwardt:
`check_data.py` is also happy (just checked), so merging.

PR #360, opened 2023-01-20
SPR1-2674: Fix unit tests on Jenkins

ludwigschwardt:
Since around May 2020 (v0.17 or so) the failure rate of unit tests on Jenkins has gone up a lot. Interestingly, these issues do not occur on my MacBook Pro or on an equivalent Docker container used outside Jenkins (with identical MinIO version etc).

The failures are mostly attributed to:

  - broken pipes in token proxy server
  - read timeouts on MinIO (especially large dask array activity)
  - 503 errors on MinIO startup (typically manifests during `test_chunk_bool_1dim_and_too_small` test)
  - duration check failures (not addressed yet)

While we are at it, make the tests pass with the latest PyJWT.

### DIFF
--- a/katdal/test/s3_utils.py
+++ b/katdal/test/s3_utils.py
@@ -122,7 +122,7 @@ class S3Server:
 
         with contextlib.ExitStack() as exit_stack:
             exit_stack.callback(self._process.terminate)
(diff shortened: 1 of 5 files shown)


richarms (approved):
All changes look good.

PR #361, opened 2023-01-23
SPR1-2674: Move unit tests from nose to pytest

ludwigschwardt:
Nose hasn’t been updated in 7 years and finally does not run on Python 3.10 anymore. This makes it irritating to test katdal, which now needs a Python 3.9 environment just for testing.

Convert all the tests to pytest. Try the nose2pytest script to help with this.

In the process, make the duration checks optional on Jenkins (where it frequently fails with large discrepancies).

Also resolve a bunch (269 928!) of NumPy deprecation warnings.

### DIFF
--- a/katdal/test/conftest.py
+++ b/katdal/test/conftest.py
@@ -29,10 +29,8 @@ def pytest_runtest_makereport(item, call):
         and report.passed
         and not minimum <= report.duration <= maximum
     ):
-        time_range = [minimum, maximum]
         # Mark test as failed and report the timing discrepancy
         report.outcome = 'failed'
-        report.longrepr = f"""
-        Test took {report.duration:g} seconds, which is outside the range {time_range}
-        """
+        report.longrepr = (f"\nTest took {report.duration:g} seconds, "
+                           f"which is outside the range [{minimum:g}, {maximum:g}]\n")
     return report
--- a/katdal/test/conftest.py
+++ b/katdal/test/conftest.py
@@ -1,18 +1,38 @@
 import pytest
 
-check_durations = False
+TEST_DURATION_TOLERANCE = 0.1
 
 
 def pytest_addoption(parser):
     parser.addoption(
         "--check-durations",
         action="store_true",
-        help="Verify how long some tests run (the ones with @duration decorator)",
+        help="Verify how long some tests run (the ones with an `expected_duration` mark)",
     )
 
 
-def pytest_configure(config):
-    # Save option at module level to be imported into standard function decorator,
-    # which seems much simpler than the fixture route.
-    global check_durations
-    check_durations = config.getoption("--check-durations", default=False)
+@pytest.hookimpl(hookwrapper=True)
+def pytest_runtest_makereport(item, call):
+    """Optionally override pytest report creation to verify test duration."""
+    report = (yield).get_result()
+    # Only continue if the user requests this and the test has an expected_duration mark
+    check_durations = item.config.getoption("--check-durations", default=False)
+    mark = item.get_closest_marker("expected_duration")
+    if not check_durations or mark is None:
+        return report
+    # The test will take at least as long as the expected duration and probably a bit longer
+    minimum = mark.args[0]
+    maximum = minimum + TEST_DURATION_TOLERANCE
+    # Only verify duration if the test itself passes (and we are in the 'call' phase of test)
+    if (
+        report.when == 'call'
+        and report.passed
+        and not minimum <= report.duration <= maximum
+    ):
+        time_range = [minimum, maximum]
+        # Mark test as failed and report the timing discrepancy
+        report.outcome = 'failed'
+        report.longrepr = f"""
+        Test took {report.duration:g} seconds, which is outside the range {time_range}
+        """
+    return report
--- a/katdal/test/conftest.py
+++ b/katdal/test/conftest.py
@@ -1,0 +1,18 @@
+import pytest
+
+check_durations = False
+
+
+def pytest_addoption(parser):
+    parser.addoption(
+        "--check-durations",
+        action="store_true",
+        help="Verify how long some tests run (the ones with @duration decorator)",
+    )
+
+
+def pytest_configure(config):
+    # Save option at module level to be imported into standard function decorator,
+    # which seems much simpler than the fixture route.
+    global check_durations
+    check_durations = config.getoption("--check-durations", default=False)
--- a/katdal/test/test_chunkstore.py
+++ b/katdal/test/test_chunkstore.py
@@ -103,7 +103,8 @@ def test_prune_chunks():
     assert new_chunks == ((10, 10, 10), (2, 2), (40,))
     assert new_index == np.s_[3:24, 0:4, 10:40]
     assert new_offset == (10, 0, 0)
-    pytest.raises(IndexError, _prune_chunks, chunks, np.s_[13:34:2, ::-1, :])
+    with pytest.raises(IndexError):
+        _prune_chunks(chunks, np.s_[13:34:2, ::-1, :])
 
 
 class TestChunkStore:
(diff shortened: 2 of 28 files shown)


richarms (approved):
Looks good Ludwig.

katdal/test/test_chunkstore_s3.py
```diff
@@ -175,34 +175,75 @@ def test_jwt_broken_token(self):
...
+    The check only happens if the module-level variable `check_durations` is True.
+    """
+    if maximum is None:
+        maximum = minimum + TEST_DURATION_TOLERANCE
+
+    def decorate(func):
+        def newfunc(*arg, **kw):
+            start = time.time()
+            result = func(*arg, **kw)
+            duration = time.time() - start
```
bmerry, line 240: `time.monotonic` is the recommended way to measure short elapsed time.

katdal/test/test_chunkstore_s3.py
```diff
@@ -175,34 +175,75 @@ def test_jwt_broken_token(self):
...
+            if check_durations and not minimum <= duration <= maximum:
+                raise UnexpectedDuration(f"Test took {duration:g} seconds, which is "
+                                         f"outside the range [{minimum:g}, {maximum:g}]")
+            return result
+        return newfunc
+    return decorate
```
bmerry, line 246: This is probably fine (modulo Richard's comment about functools.wraps), but if you wanted a pytest-onic way to do this it would be to use a `pytest.mark` to specify the range for each test, and one of the pytest hook functions to examine the mark and override how the test is run. As far as I recall the hook functions have access to the config object, so can examine the configuration without needing a nasty global variable.

katdal/test/test_chunkstore.py
```diff
@@ -316,7 +332,7 @@ def test_get_dask_array_index(self):
             np.s_[5:5, 31:31]
         ]   # TODO: use pytest.mark.parametrize when converted to pytest
```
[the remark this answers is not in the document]
ludwigschwardt: Implemented, thanks for the reminder!

katdal/test/test_chunkstore_s3.py
ludwigschwardt: Thanks for chipping in, Bruce, much appreciated! Ah, the old leap seconds :-)

katdal/test/test_chunkstore_s3.py
ludwigschwardt: So would this work with a method too (looks like it)? That was why I ended up with old-school decorators instead of fixtures.

bmerry: I would imagine so - marks can also be set on a class to affect all the methods in the class. I've not actually looked into the hooks that affect running a test, so I'm a little hazy on what the implementation would actually look like.

There is a package called pytest-timeout that might be interesting to look at as an example. It applies just an upper limit, although has a lot more complexity so that it can kill the test if it exceeds the limit.

ludwigschwardt: Thanks Bruce, you've now got me hooked on hooks 😛  and I have a nifty solution in the wings.

I did look at pytest-timeout but it's a different usecase. It basically wants to terminate tests that take too long, with all its concurrency challenges. I want to know that the test takes "exactly" as long as expected, to test my assumptions of the retry mechanisms.

katdal/test/test_chunkstore_s3.py
```diff
@@ -175,34 +175,75 @@ def test_jwt_broken_token(self):
...
         claims = decode_jwt(encode_jwt(payload))
-        assert_equal(payload, claims)
+        assert payload == claims
+
+
+class UnexpectedDuration(AssertionError):
+    """Test took either too short or too long to complete."""
+
+
+def duration(minimum, maximum=None):
```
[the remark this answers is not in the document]
ludwigschwardt: (now obsolete)

katdal/test/test_chunkstore_s3.py
ludwigschwardt: (now obsolete)

katdal/test/conftest.py
```diff
@@ -0,0 +1,38 @@
...
+        report.when == 'call'
+        and report.passed
+        and not minimum <= report.duration <= maximum
+    ):
+        time_range = [minimum, maximum]
+        # Mark test as failed and report the timing discrepancy
+        report.outcome = 'failed'
+        report.longrepr = f"""
+        Test took {report.duration:g} seconds, which is outside the range {time_range}
+        """
```
bmerry, line 37: Does pytest do sensible things with the whitespace in the triple-quoted string?

katdal/test/conftest.py
```diff
@@ -0,0 +1,38 @@
...
+
+TEST_DURATION_TOLERANCE = 0.1
+
+
+def pytest_addoption(parser):
+    parser.addoption(
+        "--check-durations",
+        action="store_true",
+        help="Verify how long some tests run (the ones with an `expected_duration` mark)",
+    )
```
bmerry, line 11: At the risk of disappearing down the rabbit-hole of improvements, is it worth making TEST_DURATION_TOLERANCE settable from the command line too?

katdal/test/conftest.py
```diff
@@ -0,0 +1,38 @@
...
+    """Optionally override pytest report creation to verify test duration."""
+    report = (yield).get_result()
+    # Only continue if the user requests this and the test has an expected_duration mark
+    check_durations = item.config.getoption("--check-durations", default=False)
+    mark = item.get_closest_marker("expected_duration")
+    if not check_durations or mark is None:
+        return report
+    # The test will take at least as long as the expected duration and probably a bit longer
+    minimum = mark.args[0]
+    maximum = minimum + TEST_DURATION_TOLERANCE
```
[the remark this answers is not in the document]
ludwigschwardt: Yes. I originally found that some tests took much longer in Jenkins, so I added arbitrary maximums to make it pass. 

I also had the interesting case of `test_store_unavailable_unresponsive_server`, which has different behaviour on Linux and macOS. I worked around this in the past by picking a custom maximum.

Now I've ditched duration checking on Jenkins and dealt with the platform issues up front, so I can go back to a single expected duration, plus a fixed tolerance that works on my laptop.

katdal/test/conftest.py
ludwigschwardt: It looks like this:
```
_______________________________________ TestS3ChunkStoreToken.test_handle_read_paused_within_npy_array ________________________________________

        Test took 0.122755 seconds, which is outside the range [0.1, 0.11]
        
------------------------------------------------------------ Captured stdout setup ------------------------------------------------------------
```
So I guess it leaves it as is. I don't mind though.

Or do you think it's more standard to move the string to the start of the line?

katdal/test/conftest.py
ludwigschwardt: :joy:

The typically overshoot on my laptop is 0.02 to 0.45 seconds. I even started with 0.05, then 0.06 and finally settled on 0.1. It runs 10x without failing, which makes it OK, I guess :-)

My idea with this is that I'd run `--check-durations` every now and then on my laptop, and just keep running it until all tests pass, instead of relaxing the tolerance.

katdal/test/conftest.py
bmerry: That's certainly not awful. What if you just made it a one-liner string with single quotes? Does it look okay without the embedded newlines?

ludwigschwardt: I remember it looked like this originally:
```
_______________________________________ TestS3ChunkStoreToken.test_handle_read_paused_within_npy_header _______________________________________
Test took 0.121048 seconds, which is outside the range [0.1, 0.11]
------------------------------------------------------------ Captured stdout setup ------------------------------------------------------------
```
Compared to the usual pytest blowout of a trace for a single error, this seemed too compact and easy to miss to me :-)

Plus I was trying to keep the line length down...

ludwigschwardt: OK, tweaked it (and lost two more lines of code :-))

bmerry: I'm happy with that.

PR #363, opened 2023-03-09
SPR1-813: Handle truncated reads, socket timeouts and reset connections (aka glitches) in S3ChunkStore

ludwigschwardt:
This improves the robustness of katdal's archive access.

While glitches where typically picked up while reading the response header, the response body had no such checks, as is the wont of the Requests library (only discovered that now 😅). Since the bulk of the data resides in the body, this meant that chunk transfers were rarely retried.

Fix this by integrating retries into the response reading too. Maintain a single `Retry` object per session, which is passed to the request, retrieved from the response and updated in case the response content has issues. Completely overhaul `S3ChunkStore.request`. Replace `TruncatedRead` with the more standard `IncompleteRead` exception. Retries can now be overridden per request.

This gets rid of the clunky fake 555 status code that treats truncated reads as status errors (they should be read errors instead).

Ensure that glitches are also handled when downloading data in non-streaming fashion via `response.content` (e.g. when downloading RDB files). Add a whole bunch of unit tests that verify glitch behaviour.

This addresses #296 and #362.

Review tip: There was quite an evolution as I tried and abandoned various approaches. I would start with the full diff and work backwards to specific commits to make sense of it.

### DIFF
--- a/katdal/chunkstore_s3.py
+++ b/katdal/chunkstore_s3.py
@@ -538,7 +538,7 @@ class S3ChunkStore(ChunkStore):
         """Assemble URL corresponding to chunk (or array) name."""
         return urllib.parse.urljoin(self._url, to_str(urllib.parse.quote(chunk_name + extension)))
 
-    def complete_request(
+    def request(
         self,
         method,
         url,
@@ -613,7 +613,7 @@ class S3ChunkStore(ChunkStore):
             return
         try:
             # Speed up the request by only checking that the bucket has at least one key
-            response = self.complete_request('GET', bucket, params={'max-keys': 1})
+            response = self.request('GET', bucket, params={'max-keys': 1})
         except S3ObjectNotFound as err:
             # There is no point continuing if the bucket is completely missing
             raise StoreUnavailable(err) from chunk_error
@@ -633,8 +633,8 @@ class S3ChunkStore(ChunkStore):
         # work with non-identity encodings.
         headers = {'Accept-Encoding': 'identity'}
         try:
-            chunk = self.complete_request('GET', url, _read_chunk, chunk_name=chunk_name,
-                                          headers=headers, stream=True)
+            chunk = self.request('GET', url, _read_chunk, chunk_name=chunk_name,
+                                 headers=headers, stream=True)
         except S3ObjectNotFound as err:
             # If the entire bucket is gone, this becomes StoreUnavailable instead
             self._verify_bucket(url, err)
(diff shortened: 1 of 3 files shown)


richarms:
In testing with both an isolated docker container, and from within a virtual environment, I get the following error consistently using the `spr1-813-handle-read-errors` branch, but not with the `master` branch:

```Traceback (most recent call last):
  File "/home/kat/ve3/bin/mvftoms.py", line 916, in <module>
    main()
  File "/home/kat/ve3/bin/mvftoms.py", line 679, in main
    averager.average_visibilities(vis_data, weight_data, flag_data,
  File "/home/kat/ve3/lib/python3.8/site-packages/katdal/averager.py", line 149, in average_visibilities
    _average_visibilities(vis, weight, flag, timeav, chanav, flagav)
  File "/home/kat/ve3/lib/python3.8/site-packages/numba/core/dispatcher.py", line 415, in _compile_for_args
    error_rewrite(e, 'typing')
  File "/home/kat/ve3/lib/python3.8/site-packages/numba/core/dispatcher.py", line 358, in error_rewrite
    reraise(type(e), e, None)
  File "/home/kat/ve3/lib/python3.8/site-packages/numba/core/utils.py", line 80, in reraise
    raise value.with_traceback(tb)
numba.core.errors.TypingError: Failed in nopython mode pipeline (step: nopython frontend)
No implementation of function Function(<built-in function empty>) found for signature:

 >>> empty(Literal[int](128), dtype=Function(<class 'bool'>))

There are 2 candidate implementations:
   - Of which 2 did not match due to:
   Overload of function 'empty': File: numba/core/typing/npydecl.py: Line 504.
     With argument(s): '(int64, dtype=Function(<class 'bool'>))':
    No match.

During: resolving callee type: Function(<built-in function empty>)
During: typing of call at /home/kat/ve3/lib/python3.8/site-packages/katdal/averager.py (49)


File "../ve3/lib/python3.8/site-packages/katdal/averager.py", line 49:
def _average_visibilities(vis, weight, flag, timeav, chanav, flagav):
    <source elided>
        weight_sum = np.empty(bl_step, weight.dtype)
        flag_any = np.empty(bl_step, dtype=bool)
        ^
```

Any ideas what might be going on?

ludwigschwardt:
That is a bit strange... I would have thought that this is a Numba + NumPy version issue, but if those are the same between the two branches, it's hard to say.

If I read the error message correctly, there is a signature mismatch on np.empty.

Try replacing the line
```Python
bl_step = 128
```
with
```Python
bl_step = np.int64(128)
```

richarms (approved):

ludwigschwardt:
Thanks Richard! Could you please add some notes about the testing you did to the SPR1-813 ticket? And what happened to the Numba issue discussed above?

PR #366, opened 2023-07-31
Update mvftoms.py to support OTF scans

Jordatious:


ludwigschwardt (approved):
Looks good! I'm still not sure how the field and scan info interact in the MS, but this seems to work for folks.

ludwigschwardt:
This addresses MKAIV-1630.

rstofi:
Hi, I quickly looked at an OTF scanning MS created with this and it seems good. Thanks a lot!

ludwigschwardt:
Thanks Kistóf! That is mostly what I needed: someone to tell me what I'm supposed to do, and to check that it works 😁

PR #269, opened 2019-10-02
Issue 267

bennahugo:
- Exposes field, antenna and scan subset selections as switches
- Switch to argparse because optparse is deprecated and does not have support for custom types

The following is an example:
```
(venvkatdal) > $ mvftoms.py --antenna m000,m001 --scans 0~5 --field J1939-6342,J1331+3030 --flags cam,data_lost -f 1559937657.rdb
No calibration products will be applied
Per user request the following antennae will be dumped: 'm000', 'm001'
Per user request the following fields will be dumped: 'J1939-6342', 'J1331+3030'
Per user request the following scans will be dumped: 1, 3
Extract MS for spw 0: centre frequency 1284000000 Hz
Will create MS output in 1559937657.ms

#### Producing a full polarisation MS (HH,HV,VH,VV) ####


Using array as the reference antenna. All targets and scans will be based on this antenna.


Iterating through scans in dataset(s)...

Writing static meta data...
scan   1 ( 151 samples) loaded. Target: 'J1939-6342'. Writing to disk...
Wrote scan data (99.154664 MiB) in 111.232912 s (0.891415 MiBps)

scan   3 ( 601 samples) loaded. Target: 'J1331+3030'. Writing to disk...
Wrote scan data (394.648696 MiB) in 442.600555 s (0.891659 MiBps)


Writing dynamic fields to disk....

```  

and a listobs as follows:
```
2019-10-02 20:11:34     INFO    listobs::::+    ##########################################
2019-10-02 20:11:34     INFO    listobs::::+    ##### Begin Task: listobs            #####
2019-10-02 20:11:34     INFO    listobs::::     listobs(vis="1559937657.ms/",selectdata=True,spw="",field="",antenna="",
2019-10-02 20:11:34     INFO    listobs::::+            uvrange="",timerange="",correlation="",scan="",intent="",
2019-10-02 20:11:34     INFO    listobs::::+            feed="",array="",observation="",verbose=True,listfile="",
2019-10-02 20:11:34     INFO    listobs::::+            listunfl=False,cachesize=50,overwrite=False)
2019-10-02 20:11:34     INFO    listobs::ms::summary    ================================================================================
2019-10-02 20:11:34     INFO    listobs::ms::summary+              MeasurementSet Name:  /scratch/bhugo/CLUSTER_SURVEY_TAKE2/1559937657.ms      MS Version 2
2019-10-02 20:11:34     INFO    listobs::ms::summary+   ================================================================================
[...]

PR #369, opened 2023-10-12
SPR1-2912: Select multiple targets with the same name

ludwigschwardt:
A `katpoint.Catalogue` can have multiple different targets with the same name, usually differing in tags only. The DataSet selection mechanism [`d.select(targets="name")`] still used a dict-style lookup for target names though, which is constrained to return a single target: the last target with the given name.

This would normally not be a big issue as there are other ways to select targets. It became a problem after #269 (released in katdal 0.18 in April 2021) which added field selection to `mvftoms.py`. This now always selects targets by name, even if the user wants all the targets and the selection is therefore unnecessary. This caused folks to lose e.g. a flux or bpcal calibrator in the MS conversion if they used the same target as a gain calibrator later on in the observation.

Fix this by ensuring all targets with a given name are picked up in the selection.

### DIFF
--- a/katdal/dataset.py
+++ b/katdal/dataset.py
@@ -781,20 +781,26 @@ class DataSet:
                 self._time_keep &= scan_keep
             elif k == 'targets':
                 targets = v if is_iterable(v) else [v]
-                target_keep = np.zeros(len(self._time_keep), dtype=bool)
-                target_index_sensor = self.sensor.get('Observation/target_index')
+                target_indices = []
                 for t in targets:
                     try:
                         if isinstance(t, numbers.Integral):
-                            target_index = t
+                            target_indices.append(t)
                         elif isinstance(t, katpoint.Target) or isinstance(t, str) and ',' in t:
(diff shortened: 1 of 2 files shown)


spassmoor:
How is this going @ludwigschwardt ?

ludwigschwardt:
Yes, I'm waiting for someone to tell me that it works :-)

spassmoor:
I have got a test block https://archive.sarao.ac.za/scheduleblock/20231025-0009/

ludwigschwardt:
Wow, excellent!

spassmoor (approved):
This works on the dataset .  still trying to get casacore installed to test mvftoms.py which was the original  place where the bug was seen.

ludwigschwardt:
The bug is all in mvftoms, so it would be good to test that...

spassmoor:
I have successfully run a mvftoms.py on a file and it works .

spassmoor:
I think we can merge this @ludwigschwardt

PR #370, opened 2023-10-16
SPR1-2990: Change underscores in S3 bucket names to dashes

ludwigschwardt:
Before 19 December 2018 the MeerKAT Ceph archive had underscores in its S3 bucket names, even though it was in violation of the S3 spec. Since October 2023 the archive runs a stricter version of Ceph and those older buckets were renamed to use dashes / hyphens instead. The metadata in the corresponding RDB files still refer to bucket names with underscores though.

Use the `S3ChunkStore.make_url` method as a centralised tool to build URLs from paths, fixing the underscores in the bucket names in the process. All store URLs should emanate from this method.

Some alternatives I considered:
  - Fix the vis prefix in the RDB file or as it comes out in telstate (downside: the prefix applies to NPY stores as well and there are also flag bucket prefixes to consider - maybe still an option...)
  - Fix the URL inside the already centralised `S3ChunkStore.request` (downside: some URLs with underscores are floating around, which could hamper debugging as they are invalid)

### DIFF
--- a/katdal/averager.py
+++ b/katdal/averager.py
@@ -46,8 +46,8 @@ def _average_visibilities(vis, weight, flag, timeav, chanav, flagav):
         vis_sum = np.empty(bl_step, vis.dtype)
         vis_weight_sum = np.empty(bl_step, vis.dtype)
         weight_sum = np.empty(bl_step, weight.dtype)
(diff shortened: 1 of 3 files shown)


ctgschollar (approved):
Tested this out on the transfer nodes and it worked.

PR #377, opened 2024-03-18
SPR1-1155: Select targets by their tags

ludwigschwardt:
This is an attempt to resurrect PR #341. I've addressed all my own comments in the old PR. :-)

This does not yet do negation, but that needs a whole new PR (see SPR1-1323).

### DIFF
--- a/katdal/dataset.py
+++ b/katdal/dataset.py
@@ -700,7 +700,8 @@ class DataSet:
             If `spw` or `subarray` is out of range
 
         """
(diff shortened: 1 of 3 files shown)


bennahugo (approved):
Looks fine to my eye. Haven't run this, but I trust it is well tested :)

bennahugo:
Might put in set difference with a "!" operator ala CASA, but that doesn't need to be in this PR. It would be a bit non-trivial to implement on channels, so some thought needs to be put into that.

cchristelis (approved):

PR #380, opened 2024-05-07
SPR1-2655: New `mvf_download` script

ludwigschwardt:
The new `mvf_download` script downloads an MVFv4 dataset (or a subset of chunks) from the MeerKAT archive to disk using rclone. You need a new-ish [rclone](https://rclone.org/downloads/) if it is not on your system. Run the script like this:
```
mvf_download.py https://archive/1698676533/1698676533_sdp_l0.full.rdb?token=<> dest
```
Data will appear in three subdirectories in the specified output directory as
```
  dest/1698676533/...
  dest/1698676533-sdp-l0/...
  dest/1698676533-sdp-l1-flags/...
```
If the script crashes or you terminate it, you can just run it again and it will carry on, fixing any half-downloaded chunks along the way. If it completes, you can be sure that all your data is safely downloaded.

BONUS: you can even copy just parts of the data (e.g. the tracks and not the slews). This works as long as your selection picks out a subset of the chunks but leaves the chunks themselves intact. It is well suited for time-based selections, e.g.
```
mvf_download.py url directory --select='{"scans": "track"}'
mvf_download.py url directory --select='{"scans": 1}'
mvf_download.py url directory --select='{"scans": [0, 1, 2]}'
mvf_download.py url directory --select='{"targets": "J1939-6342"}'
```
The chunks that are not copied will appear as "lost" data in the downloaded dataset, but that is fine. If you apply the same selection, you won't see it.

In the process I also
- improved the chunk store to expose lists of chunk names via PlaceholderChunks,
- improved dataset opening times by avoiding unnecessary initial downloads from the archive (using dask "meta"),
- cleared out crappy old scripts (FringeFinder?? 🪦) and installed `mvf_copy` properly,
- exposed useful dask array info on `VisFlagsWeights`, and
- improved documentation all around.


scripts/mvf_copy.py
```diff
@@ -39,13 +39,53 @@
...
+        formatter_class=argparse.RawDescriptionHelpFormatter,
+    )
     parser.add_argument('source', help='Dataset URL (or input RDB file path)')
     parser.add_argument('dest', type=Path, help='Output directory')
     parser.add_argument('--corrprods',
-                        help='Select correlation products (kwarg to katdal.DataSet.select). '
-                        'Keeps all corrprods by default.')
+                        help='Select correlation products (kwarg to '
+                             'katdal.DataSet.select). Keeps all corrprods by default.')
     parser.add_argument('--workers', type=int, default=8 * dask.system.CPU_COUNT,
```
spassmoor, line 89: `8 * dask.system.CPU_COUNT` Although this number of workers is ok for client side of things , the server side of things could get overwhelmed fairly quickly if  multiple users or downloads are done in parallel as would be expected for this script.

ludwigschwardt: I challenge you to benchmark this and come up with a better number once and for all 😁

ludwigschwardt:
Any more comments, @spassmoor and @Jordatious?

scripts/mvf_copy.py
```diff
@@ -39,13 +39,53 @@
...
+
+  dest/1698676533/...
+  dest/1698676533-sdp-l0/...
+  dest/1698676533-sdp-l1-flags/...
+
+Open the local dataset like this:
+
+  d = katdal.open("dest/1698676533/1698676533_sdp_l0.full.rdb")
+
+While dask allows multiple retries while downloading chunks, it currently has
```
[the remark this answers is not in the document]
ludwigschwardt: Good idea. I tried to distinguish them some more.

PR #384, opened 2025-06-05
SPR1-583: Set L2 gains to 1.0 on calibrators

ludwigschwardt:
The original scheme had a default value of `INVALID_GAIN` for all gains. When no gain solutions were found for a target, its corrections were therefore set to NaNs, thereby disabling it. This becomes an issue for L2 gains, since selfcal gains are only obtained on the imaging target and not on the calibrators. Including L2 gains in applycal (via e.g. `applycal="all"`) therefore invalidates all calibrator data by flaggingit with the `POSTPROC` flag.

A better idea is to set the L2 gains to 1.0 on calibrators. This ends up preserving the calibrator data as L1-corrected only instead of flagging it. It only applies in the case of a target without any gain solutions of the requested cal stream. A target with a fully flagged channel or "IF" in the gain solutions will still end up with NaNs in the data and `POSTPROC` flags in that channel, to avoid letting through RFI contamination.

This addresses SPR1-583 (5 years later!) as well as TESD-648 (the impetus for this fix).

### DIFF
--- a/katdal/applycal.py
+++ b/katdal/applycal.py
@@ -1,5 +1,5 @@
 ###############################################################################
-# Copyright (c) 2018-2021, National Research Foundation (SARAO)
+# Copyright (c) 2018-2021,2025, National Research Foundation (SARAO)
 #
 # Licensed under the BSD 3-Clause License (the "License"); you may not use
 # this file except in compliance with the License. You may obtain a copy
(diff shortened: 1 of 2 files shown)


KimMcAlpine (approved):
Looks good

PR #385, opened 2025-06-18
SPR1-3236: Refresh for Python 3.13 and NumPy 2.0

ludwigschwardt:
- [broken tests] Don't put zero-dimensional arrays (i.e. `np.array(1.0)`) into `S3ChunkStore` if on Python 3.12. This is only for testing corner cases anyway and no functionality is lost.
- [broken tests] Replace puzzling `float32` timestamps with proper `float64` ones (uncovered thanks to NumPy 2.0 data type promotion changes).
- [broken tests] Work around Dask slicing bug dask/dask#11993 for now (another corner case scenario).
- [warnings] Don't divide by NaNs when normalising gains.
- [warnings] Suppress `"invalid value"` warnings when `np.reciprocal` is applied to zeros and NaNs.

### DIFF
--- a/katdal/applycal.py
+++ b/katdal/applycal.py
@@ -17,6 +17,7 @@
 """Utilities for applying calibration solutions to visibilities and weights."""
 
 import logging
+import warnings
 
 import dask.array as da
 import numba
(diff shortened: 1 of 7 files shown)


richarms (approved):
This all looks great Lud. Your work on `quiet_reciprocal` will be appreciated by future loggers everywhere ;-)

PR #391, opened 2026-03-02
SPP1-189: Make TestS3ChunkStore more flexible

ludwigschwardt:
It is convenient to start a local S3 server using `TestS3ChunkStore`. I use this to benchmark Zarr, for example. It would be nice to open buckets with anonymous read support as they are on the real archive. This enables the Zarr HTTP filesystem backend (instead of s3fs), a lighter interface with simple support for our bearer tokens.

The `public_read` parameter to `S3ChunkStore` does this, and is passed into `TestS3ChunkStore.prepare_store_args` within the `setup_class` call. Make it possible to add more parameters this way, by adding `**kwargs` to `setup_class`.

### DIFF
--- a/katdal/test/test_chunkstore_s3.py
+++ b/katdal/test/test_chunkstore_s3.py
@@ -264,15 +264,21 @@ class TestS3ChunkStore(ChunkStoreTestBase):
         return url, kwargs
 
     @classmethod
-    def setup_class(cls):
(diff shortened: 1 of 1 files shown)


SihlanguI (approved):
This looks good—clean change that safely adds flexibility by forwarding `**kwargs` through `setup_class` to `prepare_store_args`, enabling useful configuration like `public_read` without impacting existing tests. Approving.

PR #393, opened 2026-06-09
SPP1-214: Modernise package structure

ludwigschwardt:
Take the plunge! (While still supporting Python 3.6 😁)

- Move `katdal` to `src/katdal`
- Move `scripts` to `src/katdal/scripts` (at least the ones we install)  **[the main impetus for this PR!]**
- Move tests out of `katdal/test` to `tests`
- Move all package metadata to `pyproject.toml` and ditch `setup.py` and `setup.cfg`
- Accidentally enable `flake8` in Jenkins and fix some errors 😂
- Dabble with `setuptools_scm` but I don't like that the local version scheme won't do branch names without adding code
- Switch from `katversion` to `versioningit` to solve this in the end (R.I.P. katversion 🪦💐)


tamerakassie (commented):
Just a few comments, largely for my own understanding of how you are using `versioningit`

pyproject.toml
```diff
@@ -1,2 +1,99 @@
...
+
+[project.urls]
+Homepage = "https://github.com/ska-sa/katdal"
+Documentation = "https://katdal.readthedocs.io"
+Repository = "https://github.com/ska-sa/katdal.git"
+Issues = "https://github.com/ska-sa/katdal/issues"
+Changelog = "https://github.com/ska-sa/katdal/blob/master/NEWS.rst"
+
+[project.scripts]
+"mvf_copy.py" = "katdal.scripts.mvf_copy:main"
```
tamerakassie, line 74: I was wondering if we should drop the `.py` in these `script.py` executables; should they be extensionless or this not required as part of the package modernisation since it might confuse users?

pyproject.toml
```diff
@@ -1,2 +1,99 @@
...
+"mvf_download.py" = "katdal.scripts.mvf_download:main"
+"mvftoms.py" = "katdal.scripts.mvftoms:main"
+
+
+
+[tool.setuptools.dynamic]
+readme = {file = ["README.rst", "NEWS.rst"]}
+
+[tool.versioningit]
```
tamerakassie, line 83: Just to also confirm my understanding here, Versioningit uses git commands to derive a version of the package at runtime; it builds the format string using specific keys set in the `[tool.versioningit.format]` subtable (below), with a format as follows `tag+dev{distance}+{branch}+{commithash}`.

pyproject.toml
```diff
@@ -1,2 +1,99 @@
...
+write = {file = "src/katdal/_version.py"}
```
tamerakassie, line 84: Just to confirm, here you tell versioningit were to write the determined version dynamically at build time? And this is imported in `/src/katdal/__init__.py` at runtimes, and if katdal was never installed the fallback is the `tag+unknown+timestamp` version.
Question: previously, with katversion, the version was computed every time at import? And now, with `versioningit` or even `setuptools_scm`, its version is computed only once at build and saved in a `__version.py` and imported thereafter at runtime?

Also here, why did you choose to use the inline for the file instead of  `[tool.versioningit.write]`?

pyproject.toml
```diff
@@ -1,2 +1,99 @@
...
+
+[tool.versioningit.vcs]
+method = "git-archive"
```
tamerakassie, line 87: So, here just to also confirm my understanding the `git-archive`  method covers the building of the package on Jenkins (no tags, no git describe) so the baked in verison set  with the describe-subst version that you export to gitattributes and can handle git commands locally? Whilst the `git` method fails in a case of no `.git`

pyproject.toml
```diff
@@ -1,2 +1,99 @@
...
+describe-subst = "$Format:%(describe)$"
+# Set a default tag since the Jenkins pipeline fetches the package git repo without tags
+default-tag = "v0.0"
+
+[tool.versioningit.format]
```
tamerakassie, line 92: This is very neat, looking a bit into this subtable i read that the branch name field here is computed under the hood with `git symbolic-ref --short HEAD` while maybe some of the others like `{rev}, {base_version}` use `git describe`,  I was wondering what happens in a case where there is a detached HEAD what would happen, would it just exit with versioningit setting the branch name to None/empty (if there was a case where I checked out a commit hash, then did a local install to test something). Which means it would still install, just maybe with some other default branch name here. 

But more importantly, in the Jenkins CI case, it fetched no tags, so we use this baked in version you set (also with the default tag). So if jenkins made git checks out a commit hash, it does not really matter since it would use the baked version from the `export-subst`.

KimMcAlpine (approved):
Exciting stuff, interested to see your new tests for the mvftoms

pyproject.toml
ludwigschwardt: I've kept them for backwards compatibility, otherwise 10 people will knock on my door tomorrow complaining that their scripts don't work anymore 😋

That is a good suggestion for version 2.0, though. I also dislike `mvftoms` that would look better as `mvf_to_ms`, given the other names (although those came later, to be fair...). Maybe even better: `mvf-copy`, etc.

pyproject.toml
ludwigschwardt: Two corrections:
- `versioningit` derives the version with git at package build time (`pip install` or wheel building), not at run / import time. `katversion` did the latter as a fallback if the version wasn't baked in.
- The version depends on whether you are at a specific tag or not. If on the tag, that is just the version (normalised and without the `"v"`). This is called `{base_version}` below. If between tags, it becomes `{base_version}.post1.dev{distance}+{branch}.{vcs}{rev}`, so e.g. `1.0.post1.dev23+spr1.2373.my.shiny.new.branch.gHASH`. The `post1` ensures that the new version comes *after* the `base_version` and avoids having to guess the next version we are working towards (needed if you just have the `.dev` part). This follows `setuptools_scm`'s `no-guess-dev` approach.

Another important thing is that `versioningit` ensures that the local part (after the `"+"` sign) only contains letters, digits and periods to satisfy PEP 440, so you can't have two pluses like above (which is why you need `.dirty` too, not `setuptools_scm`'s `+dirty` 🤦🏼‍♂️ ).

pyproject.toml
ludwigschwardt: > Just to confirm, here you tell versioningit were to write the determined version dynamically at build time?

Yes.

> And this is imported in /src/katdal/__init__.py at runtimes, and if katdal was never installed the fallback is the tag+unknown+timestamp version.

Yes.

> Question: previously, with katversion, the version was computed every time at import? And now, with versioningit or even setuptools_scm, its version is computed only once at build and saved in a __version.py and imported thereafter at runtime?

No, `katversion` appended the line `__version__ = "..."` directly to the end of `__init__.py`, while deleting that block:
```Python
# BEGIN VERSION CHECK
# Get package version when locally imported from repo or via -e develop install
try:
    import katversion as _katversion
except ImportError:
    import time as _time
    __version__ = "0.0+unknown.{}".format(_time.strftime('%Y%m%d%H%M'))
else:
    __version__ = _katversion.get_version(__path__[0])
# END VERSION CHECK
```
I contend that that is even securer, since the user can forget to import `__version__` or modify it still before the end of `__init__.py`.

So all three versioning packages are similar in that regard.

> Also here, why did you choose to use the inline for the file instead of [tool.versioningit.write]?

To save lines... I don't see why one has to start a whole subtable for just one entry, just because the docs have that example 😁

pyproject.toml
ludwigschwardt: No.

Jenkins still uses a proper git repo with a `.git` directory. It just truncates the commit history to a single commit on a single branch and drops the tags (i.e. it's a **shallow** repo). `git describe` still works, but returns an empty string because there are no tags. The versioning tools then fall back to a default mechanism.

You get a "git archive" in two ways that I know of:
- Run the `git archive` command.
- Through GitHub, by clicking on tag / branch source code `zip` or `tar.gz` links like these:
618752468-ddb44e25-e48f-4082-9910-23a30e16fc94

An example git archive link is: https://github.com/ska-sa/katdal/archive/refs/tags/v0.23.zip

That has no `.git` directory, similar to a Python sdist package.

pyproject.toml
ludwigschwardt: Interesting question!

I switched to the `spp1-214-modernise-package-structure` branch and then detached and ran the `versioningit` command-line tool to inspect the version:
```
git switch HEAD~1 --detach
versioningit
```
This gave `0.23.post1.dev21+None.g1b63552`. So you are right! The branch becomes `None` but all the rest remains in place, so you should have a unique version that is traceable to the exact commit where you detached.

PR #396, opened 2026-08-06
SPP1-90: Log mvftoms conversion  to MS HISTORY subtable

thatoeugine:
This PR logs the conversion of MVF to MS commands into the HISTORY subtable. The logged entries record:

- The task name and `katdal` version

- The input dataset path(s)

- All non-default CLI options passed to `mvftoms`

- The full CLI invocation (`sys.argv`) in the `CLI_COMMAND `column

Each row carries an MJD timestamp with `APPLICATION='mvftoms'`, `ORIGIN='mvftoms'`, and `PRIORITY='INFO'`, matching the `CASA importasdm `HISTORY convention.

### DIFF
--- a/src/katdal/ms_extra.py
+++ b/src/katdal/ms_extra.py
@@ -903,9 +903,7 @@ def populate_pointing_dict(num_antennas, observation_duration, start_time, phase
     return pointing_dict
 
 
-def populate_history_dict(times, applications, origins, messages,
-                          cli_command=None, app_params=None,
-                          priorities=None, object_ids=None, observation_ids=None):
+def populate_history_dict(times, applications, origins, messages, cli_command=None, app_params=None):
     """Construct a dictionary containing all standard columns for the HISTORY subtable.
 
     Parameters
@@ -922,12 +920,6 @@ def populate_history_dict(times, applications, origins, messages,
         CLI arguments vector per row
     app_params : sequence of list of str, optional
         Application parameters vector per row
-    priorities : sequence of str, optional
-        Priority strings (e.g., 'INFO')
-    object_ids : sequence of int, optional
-        Object IDs (default 0)
-    observation_ids : sequence of int, optional
-        Observation IDs (default -1)
 
     Returns
     -------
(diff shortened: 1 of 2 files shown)


KimMcAlpine (changes requested):
Looks good, just a one small change and we can debate the format slightly :)

src/katdal/scripts/mvftoms.py
```diff
@@ -565,6 +566,52 @@ def _cp_index(a1, a2, pol):
...
+            # Append user-selected command line parameters
+            for opt, val in vars(options).items():
+                if val is not None:
+                    messages.append(f"{opt:15s} = {repr(val)}")
+
+            n_rows = len(messages)
+            times = [current_mjd_sec] * n_rows
+            applications = ["mvftoms"] * n_rows
+            origins = ["mvftoms"] * n_rows
+            cli_commands = [cli_cmd] * n_rows
```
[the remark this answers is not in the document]
ludwigschwardt: No strong opinions, since HISTORY is a bit of a dumping ground... Just consider one of the nice use cases of HISTORY:
- You write an MS and it crashes midway.
- You restart mvftoms (even weeks later) and you want to be sure that you are running it with the same invocation as before.
- You open the HISTORY and need to compare it against whatever the new run produces.
- So maybe you need to be able to load a dict of actual params from HISTORY.

src/katdal/ms_extra.py
```diff
@@ -903,6 +903,67 @@ def populate_pointing_dict(num_antennas, observation_duration, start_time, phase
     return pointing_dict
 
 
+def populate_history_dict(times, applications, origins, messages,
+                          cli_command=None, app_params=None,
+                          priorities=None, object_ids=None, observation_ids=None):
```
ludwigschwardt: For these `populate_*_dict` functions I tend to pick parameters that I'm interested to set. The default ones I simply keep local to this function. So you don't need to pass all these values if you are never going to pick another value.

src/katdal/ms_extra.py
```diff
@@ -903,6 +903,67 @@ def populate_pointing_dict(num_antennas, observation_duration, start_time, phase
...
+    if not (len(applications) == len(origins) == len(messages) == n):
+        raise ValueError("All main input sequences must have the same length")
+
+    # Supply default column data if optional sequences are not provided
+    if cli_command is None:
+        cli_command = [[""]] * n
+    if app_params is None:
+        app_params = [[""]] * n
+    if priorities is None:
+        priorities = ["INFO"] * n
```
ludwigschwardt: Rather use `np.tile("INFO", n)`, which also solves the dtype issue below.

src/katdal/ms_extra.py
```diff
@@ -903,6 +903,67 @@ def populate_pointing_dict(num_antennas, observation_duration, start_time, phase
...
+    if object_ids is None:
+        object_ids = [0] * n
+    if observation_ids is None:
+        observation_ids = [-1] * n
+
+    history_dict = {}
+    history_dict['APP_PARAMS'] = np.asarray(app_params, dtype=object)
```
ludwigschwardt: Avoid `object` dtypes as casacore won't like them. You need plain old data in the arrays.

src/katdal/ms_extra.py
```diff
@@ -903,6 +903,67 @@ def populate_pointing_dict(num_antennas, observation_duration, start_time, phase
...
+
+    # Supply default column data if optional sequences are not provided
+    if cli_command is None:
+        cli_command = [[""]] * n
+    if app_params is None:
+        app_params = [[""]] * n
+    if priorities is None:
+        priorities = ["INFO"] * n
+    if object_ids is None:
+        object_ids = [0] * n
```
ludwigschwardt: Use `np.zeros(n, dtype=np.int32)`.

src/katdal/ms_extra.py
```diff
@@ -903,6 +903,67 @@ def populate_pointing_dict(num_antennas, observation_duration, start_time, phase
...
+    if observation_ids is None:
+        observation_ids = [-1] * n
```
ludwigschwardt: Try: `np.full(n, -1, dtype=np.int32)` to have the correct dtype.

ludwigschwardt (changes requested):
You still have to write the dict out... I would also consider writing to that table twice:
- the initial big one with the params and versions and "mvftoms conversion started".
- an extra row ("mvftoms conversion complete") when the script completes.

Another approach you might consider: a custom log message handler that means you can use `log.info("dddd")` to log arbitrary stuff to the MS whenever necessary.

src/katdal/scripts/mvftoms.py
```diff
@@ -565,6 +566,52 @@ def _cp_index(a1, a2, pol):
...
                 start_time, end_time, telescope_name, dataset.observer, dataset.experiment_id)
 
+            # Log conversion into HISTORY subtable
+
+            current_mjd_sec = katpoint.Timestamp().to_mjd() * 86400.0
+            # Capture the exact command line invocation
+            cli_cmd = list(sys.argv)
+
+            # Construct message lines matching CASA task history format
+            messages = [
```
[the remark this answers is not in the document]
thatoeugine: Added kapoint version! Checked the pipeline and requirements file as well, doesn't look like astropy is used anywhere.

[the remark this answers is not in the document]
ludwigschwardt: The irony is that astropy *should* be used... This is technically a bug, uncovered by Mario's on-the-fly issues, but it first needs katdal to be converted to the New Katpoint. For now you could replace it with the version of the `ephem` package (which will be there if katpoint < 1).

src/katdal/scripts/mvftoms.py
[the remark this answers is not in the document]
thatoeugine: Thanks guys, makes sense. 
I'll keep `cli_command` as is for the reproducibility aspect.

